PR #1, opened 2020-02-07
[POC] Add openedx tracking logs parsing

jmaupetit:
## Purpose

This first work is the first step of a POC of what will be openfun/ralph.

## Proposal

- [x] configure python package
- [x] add docker environment
- [x] add GELF formatted tracking logs parser
- [x] add tests
- [x] configure the CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,357 @@
+# Configuration file anchors
+generate-version-file: &generate-version-file
+  run:
+    name: Create a version.json
+    command: |
+      # Create a version.json à-la-mozilla
+      # https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md
+      printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' \
+        "$CIRCLE_SHA1" \
+        "$CIRCLE_TAG" \
+        "$CIRCLE_PROJECT_USERNAME" \
+        "$CIRCLE_PROJECT_REPONAME" \
+        "$CIRCLE_BUILD_URL" > version.json
+
+version: 2
+jobs:
+  # Git jobs
+  # Check that the git history is clean and complies with our expectations
+  lint-git:
+    docker:
+      - image: circleci/python:3.8-buster
+    working_directory: ~/fun
+    steps:
+      - checkout
+      # Make sure the changes don't add a "print" statement to the code base.
+      # We should exclude the ".circleci" folder from the search as the very command that checks
+      # the absence of "print" is including a "print(" itself.
+      - run:
+          name: enforce absence of print statements in code
+          command: |
+            ! git diff origin/master..HEAD -- . ':(exclude).circleci' | grep "print("
+      - run:
+          name: Check absence of fixup commits
+          command: |
+            ! git log | grep 'fixup!'
+      - run:
+          name: Install gitlint
+          command: |
+            pip install --user gitlint
+      - run:
+          name: lint commit messages added to master
+          command: |
+            ~/.local/bin/gitlint --commits origin/master..HEAD
+
+  # Check that the CHANGELOG has been updated in the current branch
+  check-changelog:
+    docker:
+      - image: circleci/buildpack-deps:stretch-scm
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - run:
+          name: Check that the CHANGELOG has been modified in the current branch
+          command: |
+            git whatchanged --name-only --pretty="" origin..HEAD | grep CHANGELOG
+
+  # Check that the CHANGELOG max line length does not exceed 80 characters
+  lint-changelog:
+    docker:
+      - image: debian:stretch
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - run:
+          name: Check CHANGELOG max line length
+          command: |
+            # Get the longuest line width (ignoring release links)
+            test $(cat CHANGELOG.md | grep -Ev "^\[.*\]: https://github.com/openfun" | wc -L) -le 80
+
+  # ---- Docker jobs ----
+  # Build the Docker image ready for production
+  build-docker:
+    docker:
+      - image: circleci/buildpack-deps:stretch
+    working_directory: ~/fun
+    steps:
+      # Checkout repository sources
+      - checkout
+      # Generate a version.json file describing app release
+      - <<: *generate-version-file
+      # Activate docker-in-docker (with layers caching enabled)
+      - setup_remote_docker:
+          docker_layer_caching: true
+      # Each image is tagged with the current git commit sha1 to avoid
+      # collisions in parallel builds.
+      - run:
+          name: Build production image
+          command: docker build -t ralph:${CIRCLE_SHA1} --target production .
+      - run:
+          name: Check built image availability
+          command: docker images "ralph:${CIRCLE_SHA1}*"
+
+  # ---- Backend jobs ----
+  # Build backend development environment
+  build:
+    docker:
+      - image: circleci/python:3.8-buster
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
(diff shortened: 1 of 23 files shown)


.dockerignore
```diff
@@ -0,0 +1,30 @@
+# Python
+__pycache__
+*.pyc
+**/__pycache__
+**/*.pyc
+build/
+dist/
+**/*.egg-info
+venv
+.venv
```
lunika: you should also add

```
**/venv
**/.venv
```

.circleci/config.yml
```diff
@@ -0,0 +1,357 @@
...
+        "$CIRCLE_BUILD_URL" > version.json
+
+version: 2
+jobs:
+  # Git jobs
+  # Check that the git history is clean and complies with our expectations
+  lint-git:
+    docker:
+      - image: circleci/python:3.8-buster
+    working_directory: ~/fun
```
lunika: Why did you chosse `fun` and not `ralph` which is the project name ?

.circleci/config.yml
```diff
@@ -0,0 +1,357 @@
...
+  # Build backend development environment
+  build:
+    docker:
+      - image: circleci/python:3.8-buster
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ .Revision }}
```
lunika: In other proejcts (marsha, richie) we are also using a more generic key, here it should be `v1-dependencies`, why not doing this in this project ?

src/ralph/filters.py
```diff
@@ -1,6 +1,23 @@
 """
 Ralph tracking logs filters.
 """
+from .exceptions import EmptyEventsCollection, EventKeyError
+
+
+def empty(events):
```
lunika: should be `not_empty` ? The description is the opposite of the function name.

tests/test_parsers.py
```diff
@@ -0,0 +1,136 @@
...
+    assert len(events) == 2
+    assert events.equals(pd.DataFrame({"username": ["foo", ""]}))
+
+
+# pylint: disable=invalid-name,unused-argument
+def test_gelfparser_parse_gzipped_file(fs, gelf_logger):
+    """Test the GELFParser parsing using a gzipped GELF file."""
+
+    gelf_logger.info('{"username": "foo"}')
+    gelf_logger.info('{"username": "boo"}')
```
lunika:
```suggestion
    gelf_logger.info('{"username": "bar"}')
```
please respect the `foo` `bar` `baz` convention 😁

src/ralph/parsers.py
```diff
@@ -0,0 +1,86 @@
...
+                logger.info("Current filter: %s", _filter.__name__)
+                if records.empty:
+                    logger.warning(
+                        "Current chunk contains no records, will stop filtering."
+                    )
+                    break
+                records = _filter(records)
+                logger.debug("Filter: %s (records: %d)", _filter.__name__, len(records))
+
+            events = events.append(records, ignore_index=True, sort=False)
```
madmatah: Suggestion : Remove the `empty` filter and add a test here :

```suggestion
            if not records.empty:
                events = events.append(records, ignore_index=True, sort=False)
```

(Or maybe i missed a use case for this empty filter !)

src/ralph/filters.py
```diff
@@ -0,0 +1,38 @@
...
+    Args:
+        events (DataFrame): events to filter
+
+    Returns:
+        Unfiltered input events (pandas DataFrame).
+
+    """
+
+    if len(events) < 1:
+        raise EmptyEventsCollection("Input events collection is empty.")
```
madmatah: Why raising an exception if the DataFrame is empty ?
Don't we only want to ignore them ?

src/ralph/parsers.py
```diff
@@ -0,0 +1,86 @@
...
+        events = pd.DataFrame()
+
+        chunks = pd.read_json(input_file, lines=True, chunksize=chunksize)
+        for chunk in chunks:
+            records = pd.read_json("\n".join(chunk["short_message"]), lines=True)
+            logger.debug("Records before filtering: %d", len(records))
+
+            for _filter in filters:
+                logger.info("Current filter: %s", _filter.__name__)
+                if records.empty:
```
madmatah: The `empty` filter is never called with this test, right ?

setup.cfg
```diff
@@ -0,0 +1,80 @@
...
+    Natural Language :: English
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.8
+
+[options]
+include_package_data = True
+install_requires =
+    pandas==1.0.0
+package_dir =
+    =src
```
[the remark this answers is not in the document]
jmaupetit: That's not obvious, but those two lines are perfectly fine. Reference: https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages

.dockerignore
jmaupetit: Do you think people are willing to create virtualenvs everywhere in the project?

.circleci/config.yml
jmaupetit: This is the standard used on most projects except... Marsha :wink:

.circleci/config.yml
jmaupetit: From what I remember, we already had package collisions by falling back to a more generic key in Richie's CI (which I basically copied here)

By doing so, we ensure that each CI workflow as it's own new consistent cache. I think it's worth it as installing python dependencies is not that time/resources consuming.

src/ralph/filters.py
jmaupetit: This is a silly filter that is no longer required, I'll remove it.

src/ralph/parsers.py
jmaupetit: Appending an empty DataFrame to a DataFrame has no effect so I think it's not worth adding such test.

src/ralph/filters.py
jmaupetit: This filter is silly. I'll remove it.

src/ralph/parsers.py
jmaupetit: Yup, I've changed this implementation and left the unused `empty` filter. I'll remove it.

src/ralph/parsers.py
SergioSim: Hello, I have a question, you are reading chunk by chunk and then appending all to the events dataframe - will it not result in loading almost the full file in memory (filtered lines removed)?

jmaupetit: Good catch! Yes it will.. even if it's highly optimized in pandas, for this first implementation, we will need to: i. inspect memory usage, and ii. split input files to mitigate the (potential) issue as a first naive approach. We discussed internally of what could be the second implementation using stream processing approach and reactive programing.

jmaupetit: :thinking: In the meantime, returning a generator would be more efficient.

jmaupetit:
As mentioned in [this comment](https://github.com/openfun/ralph/pull/1#discussion_r379444124), I've re-implemented the GELF parser using a generator to lower memory footprint, and it works pretty well! I've performed a quick profiling of this function by comparing the old and new implementation with various chunk sizes in the following figure:

naive2iterators

As you can see, by using a chunksize of 3000, the parser is able to process a 350Mo gzipped file within 2 or 3 minutes using less than 300 Mo of RAM.

Thank you all for your feedback! :pray:

src/ralph/parsers.py
```diff
@@ -0,0 +1,82 @@
...
+        """Parse GELF formatted logs (one json string event per row).
+
+        Args:
+            input_file (string): Path to the log file to parse.
+            filters (list): A list of functions to apply on parsed results to
+                filter them.
+            chunksize (int): The amount of log records to process at a time.
+
+        Returns:
+            Parser implementation should return a pandas DataFrame object with
```
madmatah: The return value of the BaseParser should be a generator, to be consistent with GELFParser ?

PR #2, opened 2020-02-27
Add "event" test fixture factory

SergioSim:
## Purpose

- Document the different events that are logged as "tracking logs" in the LMS to understand how, were and when things get logged
- A event test fixture factory could help to test the function of the GELFParser

## Proposal

- [ ] Implementation of the event test fixture factory
- [ ] Events documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ venv.bak/
 
 # Test fixtures
 data/
+
+# vscode
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -47,5 +47,3 @@ venv.bak/
 # Test fixtures
 data/
 
-# vscode
-.vscode
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": true
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.linting.pylintEnabled": true
-}
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,6 +36,7 @@ dev =
     bandit==1.6.2
     black==19.10b0
     djehouty==0.1.5
+    factory-boy==2.12.0
     Faker==4.0.1
     flake8==3.7.9
     ipdb==0.12.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,6 +36,7 @@ dev =
     bandit==1.6.2
     black==19.10b0
     djehouty==0.1.5
+    Faker==4.0.1
     flake8==3.7.9
     ipdb==0.12.2
     ipython==7.9.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,6 @@ dev =
     bandit==1.6.2
     black==19.10b0
     djehouty==0.1.5
-    factory-boy==2.12.0
     Faker==4.0.1
     flake8==3.7.9
     ipdb==0.12.2
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,4 +3,4 @@ Module py.test fixtures
 """
 # pylint: disable=unused-import
 
-from .fixtures.logs import gelf_logger  # noqa: F401
+from .fixtures.logs import event, gelf_logger  # noqa: F401
--- a/tests/fixtures/event/base.py
+++ b/tests/fixtures/event/base.py
@@ -1,0 +1,392 @@
+"""
+Base class definitions for the event fixture
+"""
+from faker import Faker
+
+# Faker.seed(0)
+FAKE = Faker()
+
+
+class BaseEventField:
+    """BaseEventField Interface.
+    The Event Field is just a wrapper around it's value which is obtained
+    by calling it's generator function.
+    """
+
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        generator,
+        removed=False,
+        optional=False,
+        dependency=None,
+        nullable=False,
+        emptiable_str=False,
+        emptiable_dict=False,
+        **generator_kwargs
+    ):
+        """BaseEventField Constuctor
+
+        Args:
+            generator (function): A function returning the value of the
+                EventField. The function will be called with the
+                generator_kwargs keyword arguments
+            removed (boolean): Defines if the Field will be removed
+                from the Event
+            optional (boolean): Defines if the Field might be removed
+                from the Event at random
+            dependency (list or string): A list of lists containing strings or
+                a single list of strings or just a string. Each
+                list defines a path to a Field from the root of the Event,
+                which value is required in order to compute this Event
+                value. For example, if the Event is:
+                    Event = {
+                        "key1": {
+                            "key2": "value2",
+                            "key3": "value3",
+                            "key4": {
+                                "key5": "value5",
+                                "key6": "value6"
+                            }
+                        },
+                        "key7": "value7"
+                    }
+                and the dependency argument is:
+                    dependency = [["key1", "key3"], ["key1", "key4"]]
+                this would mean that the Event["key1"]["key3"] Field and
+                the Event["key1"]["key4"] Object should be computed prior to
+                this event.
+                In case the EventField depends on only one other EventFiled
+                we can use just a list of strings. For example:
+                     dependency = ["key1", "key3"]
+                (this is equivalent to: dependency = [["key1", "key3"]])
+                In case the EventField depends on only one other EventField
+                and the EventField is at the root of the Event
+                we can use just a string. For example:
+                    dependency = "key1"
+                (this is equivalent to: dependency = [["key1"]])
+            nullable (boolean):  Defines if the Field might be None
+            emptiable_str (boolean): Defines if the Field might be an
+                empty string
+            emptiable_dict (boolean): Defines if the Field might be an
+                empty dictionary
+            generator_kwargs (kwargs): The generator_kwargs will be
+                passed to the `generator` function as kwargs
+        """
+        self.generator = generator
+        self.removed = removed
+        self.optional = optional
+        self.dependency = dependency
+        self.nullable = nullable
+        self.emptiable_str = emptiable_str
+        self.emptiable_dict = emptiable_dict
+        self.generator_kwargs = generator_kwargs
+        if dependency is not None:
+            if isinstance(self.dependency, str):
+                self.dependency = [[dependency]]
+            if isinstance(self.dependency[0], str):
+                self.dependency = [dependency]
+
+    def __add__(self, other):
+        """Delegates the addition opreration to the EventField value
+        Args:
+            other (any object): any object supporting
+                `this_field_value` + other
+        """
+
+        def wrap(*args, **kwargs):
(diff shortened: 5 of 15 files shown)


tests/fixtures/logs.py
```diff
@@ -24,3 +29,101 @@ def gelf_logger():
...
+        "ip": [data['ip']],
+        "agent": [data['agent']],
+        "accept_language": [data['accept_language']],
+        "event_source": data['event_source'],
+        "time": [data['time']],
+        "host": [data['host']],
+        "referer": [data['referer']],
+        "event_type": [data['event_type']],
+    })
+    return raw_json_event, expected_parsed_event
```
jmaupetit: IMO we don't need to JSON intermediary, I would suggest that you only return the factory, _e.g._:

```suggestion
    def _event(n):
      return pd.DataFrame(data=[{
          "username": fake.profile().get("username"),
          # [...]
      } for i in range(n)])
    return _event
```

For reference, see: https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures

tests/test_parsers.py
```diff
@@ -154,3 +154,12 @@ def test_gelfparser_parse_with_filters(fs, gelf_logger):
     events = next(parser.parse(log_file_name, filters=[filters.anonymous]))
     assert len(events) == 1
     assert events.equals(pd.DataFrame({"username": ["foo"]}))
+
+
+def test_gelfparser_parse_server_event(fs, gelf_logger, log_generator):
+    """Test the GELFParser parsing an common 'event_sourse = server' event """
```
jmaupetit: This test is not required IMO as you are testing pandas JSON conversion and our GELFParser more than any logic of your fixture.

tests/fixtures/logs.py
```diff
@@ -24,3 +29,101 @@ def gelf_logger():
...
+        "Mozilla/5.0 (PlayStation Vita 3.61) AppleWebKit/537.73 (KHTML, like Gecko) Silk/3.2",
+        "Mozilla/5.0 (Nintendo 3DS; U; ; en) Version/1.7412.EU",
+        "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)",
+        "Mozilla/5.0 (compatible; bingbot/2.0; +http://www.bing.com/bingbot.htm)",
+        "Mozilla/5.0 (compatible; Yahoo! Slurp; http://help.yahoo.com/help/us/ysearch/slurp)",
+        "Mozilla/5.0 (X11; U; Linux armv7l like Android; en-us) AppleWebKit/531.2+ (KHTML, like Gecko) Version/5.0 Safari/533.2+ Kindle/3.0+",
+        "Mozilla/5.0 (Linux; U; en-US) AppleWebKit/528.5+ (KHTML, like Gecko, Safari/528.5+) Version/4.0 Kindle/3.0 (screen 600x800; rotate)"
+    ]
+
+    return random.choice(agents)
```
jmaupetit: There is a Faker provider for this: https://faker.readthedocs.io/en/master/providers/faker.providers.user_agent.html

tests/fixtures/logs.py
```diff
@@ -24,3 +29,101 @@ def gelf_logger():
...
+
+
+def random_string_generator(size, chars=string.printable):
+    """returns a random string of size 'size' including characters in 'chars' """
+    return ''.join(random.choice(chars) for _ in range(size))
+
+@pytest.fixture
+def log_generator():
```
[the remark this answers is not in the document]
SergioSim: Thanks, good point, I have updated the name

tests/fixtures/logs.py
SergioSim: Thanks, I removed this part. Initially I thought the factory could provide the expected output too, but you're right - this tests more the pandas JSON conversion :)

tests/fixtures/logs.py
SergioSim: Thank you, this was helpful

setup.cfg
```diff
@@ -36,6 +36,7 @@ dev =
     bandit==1.6.2
     black==19.10b0
     djehouty==0.1.5
+    faker==4.0.1
```
jmaupetit: It should probably work, but, as referred in PyPI, this package's name starts with an uppercase F:

```suggestion
    Faker==4.0.1
```

tests/conftest.py
```diff
@@ -4,3 +4,4 @@
 # pylint: disable=unused-import
 
 from .fixtures.logs import gelf_logger  # noqa: F401
+from .fixtures.logs import event  # noqa: F401
```
jmaupetit:
```suggestion
from .fixtures.logs import event, gelf_logger  # noqa: F401
```

setup.cfg
SergioSim: Thanks, good catch

tests/fixtures/logs.py
```diff
@@ -24,3 +28,82 @@ def gelf_logger():
...
+                    ),
+                    "path": "/event",
+                },
+                "ip": FAKE.ipv4_public(),
+                "event": "{}",
+                "event_type": "page_close",  # could be an other sting too
+            }
+            for i in range(event_number)
+        ]
+    )
```
jmaupetit, line 93: That's a good first approach, but I think we should go deeper and handle different edx's server or browser event types, _e.g._ problem check, etc.

tests/fixtures/logs.py
```diff
@@ -24,3 +28,82 @@ def gelf_logger():
...
+                },
+                "time": FAKE.iso8601(),
+                "page": "null",
+            }
+            for i in range(event_number)
+        ]
+    )
+
+
+def browser_event(event_number):
```
jmaupetit, line 64: I would prefer factory_boy semantic and use `size` instead. WDYT?

```suggestion
def browser_event(size):
```

docs/browserLogs.md
```diff
@@ -0,0 +1,321 @@
+# browser event logs
```
jmaupetit, line 1:
```suggestion
# Browser event logs 
```

docs/browserLogs.md
```diff
@@ -0,0 +1,321 @@
...
+      "context": {
+          "user_id": 2,
+          "org_id": "universityX",
+          "course_id": "course-v1:universityX+CS111+2020_T1",
+          "path": "/event"
+      },
+      "ip": "172.18.0.1",
+      "event": "{}",
+      "event_type": "page_close"
+    }
```
jmaupetit, line 30: I prefer use the triple backquotes syntax with the target language to get syntax-highlighting when available:

```suggestion
```json
{
      "username": "toto",
      "event_source": "browser",
      "name": "page_close",
      "accept_language": "en-US,en;q=0.5",
      "time": "2020-03-02T10:12:08.992343+00:00",
      "agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:73.0) Gecko/20100101 Firefox/73.0",
      "page": "http://localhost:8072/courses/course-v1:universityX+CS111+2020_T1/courseware/5edb208a13004490909da020a9bd115d/cd2bb35541e74e8a8be5d2235d122fd9/",
      "host": "50528a92a868",
      "session": "7c26f91b2debb8fa9df150a823c9b43c",
      "referer": "http://localhost:8072/courses/course-v1:universityX+CS111+2020_T1/courseware/5edb208a13004490909da020a9bd115d/cd2bb35541e74e8a8be5d2235d122fd9/ ",
      "context": {
          "user_id": 2,
          "org_id": "universityX",
          "course_id": "course-v1:universityX+CS111+2020_T1",
          "path": "/event"
      },
      "ip": "172.18.0.1",
      "event": "{}",
      "event_type": "page_close"
}
```
```

File: tests/fixtures/logs.py (hunk 7 above)
Comment (SergioSim) [reply tests/fixtures/logs.py:93]: Thanks for the review Julien! Yes, there are more events to add, working on it

File: docs/browserLogs.md (hunk 9 above)
Comment (SergioSim) [reply docs/browserLogs.md:1]: Good catch, thanks

File: tests/fixtures/logs.py (hunk 8 above)
Comment (SergioSim) [reply tests/fixtures/logs.py:64]: Good Idea, I was not aware of factory_boy semantic

**Comment by SergioSim** [comment]
> I think you should split your notes from the log factory implementation. Maybe keep your notes in a new repository and only keep the log factory in this PR?

Will be done,  I can keep them on a separate branch on my fork

**Comment by jmaupetit** [comment]
> Will be done, I can keep them on a separate branch on my fork

File: .vscode/settings.json [hunk 11]
```diff
@@ -0,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": true
+}
```
jmaupetit, line 3: It is not recommended to version IDE settings in a repository. Instead, you should git-ignore `.vscode` in your global Git settings.

tests/fixtures/event/base.py
```diff
@@ -0,0 +1,392 @@
...
+
+class BaseEventField:
+    """BaseEventField Interface.
+    The Event Field is just a wrapper around it's value which is obtained
+    by calling it's generator function.
+    """
+
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
+    def __init__(
```
SergioSim, line 18: pylint complains about too many attributes / arguments.
Do you prefer if i wrap some arguments in a Data class instead?
On one side this will make this part cleaner but on the other side for each EventField creation more code will have to be written. Example:
current:
`self.session = FreeEventField(FAKE.md5, emptiable_str=True, raw_output=False)`
with dataclass:
`self.session = FreeEventField(FAKE.md5, properties=EventFieldProperties(emptiable_str=True), raw_output=False)`

tests/fixtures/event/base.py
```diff
@@ -0,0 +1,392 @@
...
+        self,
+        generator,
+        removed=False,
+        optional=False,
+        dependency=None,
+        nullable=False,
+        emptiable_str=False,
+        emptiable_dict=False,
```
SergioSim, line 26: Question:
The fields `nullable, emptiable_str, emptiable_dict` are only used in the `FreeEventField` class.
I can move those to the `FreeEventField` class but will have to redefine the 
 `__add__, __radd__ and __getitem__` functions which results in more code to maintain.
Do you prefer if i keep those fields here or move them?

jmaupetit: Move them where they are useful! The base class should only contain the smallest common part.

tests/fixtures/event/base.py
jmaupetit: Using a dataclass is cleaner IMO.

SergioSim: Thanks, will be done

tests/fixtures/event/base.py
SergioSim: Thanks, will be done

.gitignore
```diff
@@ -46,3 +46,6 @@ venv.bak/
 
 # Test fixtures
 data/
+
+# vscode
+.vscode
```
jmaupetit, line 51: This should be in your global Git preferences, _e.g._ `$HOME/.gitignore`:

```gitignore
# Editors
.vscode
*.sw[p,o]
```

SergioSim: Thanks

SergioSim:
Created new pull request from a feature branch, see https://github.com/openfun/ralph/pull/3

PR #3, opened 2020-05-19
Add test fixtures

SergioSim:
## Purpose

- A event test fixture factory could help to test the function of the GELFParser

## Proposal

- [ ] Implementation of the event test fixture factory

### DIFF
--- a/src/ralph/schemas/edx/base.py
+++ b/src/ralph/schemas/edx/base.py
@@ -20,7 +20,6 @@ class BaseContextSchema(Schema):
     @validates("user_id")
     def validate_user_id(self, value):
         """"Check user_id field is None, an empty string or an integer"""
-
         if value is None or value == "" or isinstance(value, int):
             return
         raise ValidationError("user_id should be None, an empty string or an integer")
@@ -31,7 +30,6 @@ class BaseContextSchema(Schema):
         "course-v1:{org_id}+{any_string}+{any_string}"
         or be an empty string if org_id is an empty string
         """
-
         org_id = data["org_id"]
         course_id = data["course_id"]
         if not org_id and course_id:
--- a/src/ralph/schemas/edx/base.py
+++ b/src/ralph/schemas/edx/base.py
@@ -1,0 +1,160 @@
+"""Base event schema definitions"""
+from ipaddress import IPv4Address
+
+from marshmallow import Schema, ValidationError, validates, validates_schema
+from marshmallow.fields import DateTime, Dict, Field, Nested, Str, Url
+from marshmallow.validate import URL, Equal
+
+# pylint: disable=no-self-use, unused-argument
+
+
+class BaseContextSchema(Schema):
+    """Represents the Base Context inherited by all event contexts"""
+
+    course_user_tags = Dict(keys=Str(), values=Str())
+    user_id = Field(required=True, allow_none=True)
+    org_id = Str(required=True)
+    course_id = Str(required=True)
+    path = Url(required=True, relative=True)
+
+    @validates("user_id")
+    def validate_user_id(self, value):
+        """"Check user_id field is None, an empty string or an integer"""
+
+        if value is None or value == "" or isinstance(value, int):
+            return
+        raise ValidationError("user_id should be None, an empty string or an integer")
+
+    @validates_schema
+    def validate_course_id(self, data, **kwargs):
+        """The course_id should be equal to
+        "course-v1:{org_id}+{any_string}+{any_string}"
+        or be an empty string if org_id is an empty string
+        """
+
+        org_id = data["org_id"]
+        course_id = data["course_id"]
+        if not org_id and course_id:
+            raise ValidationError("course_id should be empty if org_id is empty")
+        if org_id and not course_id:
+            raise ValidationError("org_id should be empty if course_id is empty")
+        if not org_id:
+            return
+        if not course_id.startswith("course-v1:"):
+            raise ValidationError("course_id should starts with 'course-v1'")
+        try:
+            organization, course, session = course_id.replace("course-v1:", "").split(
+                "+"
+            )
+        except ValueError as value_error:
+            raise ValidationError(
+                "course_id should contain an organization ID, "
+                "a course name and session separated by a +"
+            ) from value_error
+        if organization != data["org_id"]:
+            raise ValidationError(
+                "organization ID in the course ID does not match the event organization ID"
+            )
+        if len(course) == 0 or len(session) == 0:
+            raise ValidationError("course and session should not be empty")
+
+
+class ContextModuleSchema(Schema):
+    """Represents the context module field"""
+
+    usage_key = Str(required=True)
+    display_name = Str(required=True)
+
+
+class ContextSchema(BaseContextSchema):
+    """Context with module field. Present in CAPA problems related
+    events.
+    """
+
+    module = Nested(ContextModuleSchema(), required=True)
+
+    @validates_schema
+    def validate_path(self, data, **kwargs):
+        """Path should start with:
+        "/courses/{course_id}/xblock/block-v1:{course_id[10:]}
+        +type@problem+block@{usage_key}/handler/"
+        """
+
+        path = data["path"]
+        valid_path = (
+            f"/courses/"
+            f"{data['course_id']}"
+            f"/xblock/"
+            f"{data['module']['usage_key']}/"
+            f"handler/"
+        )
+        if not path.startswith(valid_path):
+            raise ValidationError(
+                f"path should start with: "
+                f"{valid_path} "
+                f"but {path[:len(valid_path)]} does not match!"
+            )
+
+
+class BaseEventSchema(Schema):
+    """Represents the Base Event Schema all events inherit from.
+    Does not define event and event_type fields
+    """
+
+    username = Str(required=True)
+    ip = Field(required=True)
+    agent = Str(required=True)
+    host = Str(required=True)
+    referer = Str(required=True)
+    accept_language = Str(required=True)
+    event_source = Str(
+        required=True,
+        validate=Equal(
+            comparable="server", error="The event event_source field is not `server`"
+        ),
+    )
+    context = Nested(BaseContextSchema(), required=True)
+    time = DateTime(format="iso", required=True)
+    page = Str(
+        required=True,
+        allow_none=True,
+        validate=Equal(comparable=None, error="The event page field is not None"),
+    )
+
+    @validates("username")
+    def validate_username(self, value):
+        """"Check username field empty or 2-30 chars long"""
+
+        if len(value) == 1 or len(value) > 30:
+            raise ValidationError(
+                "username should be empty or between 2 and 30 characters long"
+            )
+
+    @validates("referer")
+    def validate_referer(self, value):
+        """Check referer field empty or a valid URL"""
+
+        if value != "":
+            URL(relative=True)(value)
+
+    @validates("ip")
+    def validate_ip(self, value):
+        """Check the IP address is empty or a valid IPv4 address."""
+
+        if value != "":
+            try:
+                IPv4Address(value)
+            except ValueError as err:
+                raise ValidationError("Invalid IPv4 Address") from err
+
+    @staticmethod
+    def get_course_key(data):
+        """Return the course key: organization+course+session"""
+
+        return data["context"]["course_id"][10:]
+
+    @staticmethod
+    def get_block_id(data, prefix="block-v1", block_type="problem", suffix="@"):
+        """Return the block id: {prefix}:{course_key}+type@{block_type}+block{suffix}"""
+
+        return f"{prefix}:{BaseEventSchema.get_course_key(data)}+type@{block_type}+block{suffix}"
--- a/src/ralph/schemas/edx/browser.py
+++ b/src/ralph/schemas/edx/browser.py
@@ -1,0 +1,484 @@
+"""
+Browser event schema definitions
+"""
+import json
+import urllib.parse
+
+from marshmallow import ValidationError, validates, validates_schema
+from marshmallow.fields import Field, Str, Url
+from marshmallow.validate import URL, Equal, OneOf
+
+from .base import BaseEventSchema
+
+# pylint: disable=no-self-use, unused-argument
+
+MD5_HASH_LEN = 32
+
+BROWSER_EVENT_TYPE_FIELD = [
+    "book",
+    "page_close",
+    "problem_check",
+    "problem_graded",
+    "problem_reset",
+    "problem_save",
+    "problem_show",
+    "seq_goto",
+    "seq_next",
+    "seq_prev",
+    "textbook.pdf.display.scaled",
+    "textbook.pdf.outline.toggled",
+    "textbook.pdf.page.navigated",
+    "textbook.pdf.page.scrolled",
+    "textbook.pdf.thumbnail.navigated",
+    "textbook.pdf.thumbnails.toggled",
+    "textbook.pdf.zoom.buttons.changed",
+    "textbook.pdf.zoom.menu.changed",
+]
+
+BROWSER_NAME_FIELD = [
+    "textbook.pdf.page.loaded",
+    "textbook.pdf.page.navigatednext",
+    "textbook.pdf.search.executed",
+    "textbook.pdf.search.highlight.toggled",
+    "textbook.pdf.search.navigatednext",
+    "textbook.pdf.searchcasesensitivity.toggled",
+]
+
+BROWSER_EVENT_VALID_AMOUNT = [
+    "0.5",
+    "0.75",
+    "1",
+    "1.25",
+    "1.5",
+    "2",
+    "3",
+    "4",
+    "auto",
+    "page-actual",
+    "page-fit",
+    "page-width",
+]
+
+
+class BaseBrowserEventSchema(BaseEventSchema):
+    """Represents common fields and functions all BrowserEvents inherit
+    from. This type of event is triggered on (XHR) POST/GET request to
+    the `/event` URL
+    """
+
+    event_source = Str(
+        required=True,
+        validate=Equal(
+            comparable="browser", error="The event event_source field is not `browser`"
+        ),
+    )
+    event_type = Str(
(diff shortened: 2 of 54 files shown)


tests/fixtures/event/proposal.py
```diff
@@ -0,0 +1,190 @@
+"""
+Proposal for an
+edx.problem.hint.feedback_displayed event fixture definition
+"""
```
SergioSim, line 4: This is an example for the proposition of event fixture definition.
One major drawback is that it's quite similar to the current solution - just without inheritance, so I'm not sure if it's more maintainable.

tests/fixtures/event/proposal_marshmallow.py
```diff
@@ -0,0 +1,179 @@
...
+        """the course_id should be equal to
+        "course-v1:{org_id}+{any_string}+{any_string}"
+        """
+        course_id_split = data["course_id"].split("+")
+        if len(course_id_split) != 3:
+            raise ValidationError('course_id should contain 2 "+" characters')
+        if course_id_split[0] != "course-v1:{}".format(data["org_id"]):
+            raise ValidationError("course_id should start with course-v1:{org_id}")
+        if len(course_id_split[1]) == 0 or len(course_id_split[2]) == 0:
+            raise ValidationError("course name and session should not be empty")
```
jmaupetit, line 42:
```suggestion
        if not data["course_id"].startswith("course-v1:"):
            raise ValidationError("course_id should starts with 'course-v1'")
        try:
            organization, course, session = data["course_id"].replace("course-v1:").split("+")
        except ValueError:
            raise ValidationError("course_id should contain an organization ID, a course name and session separated by a +")
        if organization != data["org_id"]:
            raise ValidationError("organization ID in the course ID does not match the event organization ID")
        if not len(course) or not len(session):
            raise ValidationError("course and session should not be empty")
```

tests/fixtures/event/proposal_marshmallow.py
```diff
@@ -0,0 +1,179 @@
...
+        "/handler/xmodule_handler/problem_check"
+        """
+        if data["path"] != (
+            "/courses/{}/xblock/block-v1:{}+type@problem+block@{}"
+            "/handler/xmodule_handler/problem_check"
+        ).format(
+            data["course_id"],
+            data["course_id"][10:],
+            data["module"]["usage_key"][-32:],
+        ):
```
jmaupetit, line 59: Use f-strings!

```suggestion
        if data["path"] != (
            f"/courses/"
            f"{data['course_id']}"
            f"/xblock/"
            f"{data['module']['usage_key']}/"
            f"handler/xmodule_handler/problem_check"
        ):
```

tests/fixtures/event/proposal_marshmallow.py
```diff
@@ -0,0 +1,179 @@
...
+            raise ValidationError(
+                "path should be equal to"
+                "/courses/{course_id}/xblock/block-v1:{course_id[10:]}"
+                "+type@problem+block@{usage_key[-32:]}"
+                "/handler/xmodule_handler/problem_check"
```
jmaupetit, line 64: You may fix this message according to my suggestion.

tests/test_event_feedback_displayed.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def test_invalid_event_with_choice_all_present(valid_event):
+    """ValidationError should be raised if the event choice_all field
+    is present and the event question_type value is not "choiceresponse"
+    """
+    event = valid_event
+    event["event"]["choice_all"] = ["choice_1"]
+    with pytest.raises(ValidationError):
+        SCHEMA.load(event)
```
jmaupetit, line 165: You will need to rewrite those tests when you will have a corresponding factory.

tests/fixtures/event/proposal_marshmallow.py
```diff
@@ -0,0 +1,179 @@
...
+            )
+        # the converse is true
+        if "choice_all" not in data and data["question_type"] == "choiceresponse":
+            raise ValidationError(
+                'when event.question_type == "choiceresponse" - '
+                "event.choice_all should be present"
+            )
+
+
+class FeedbackDisplayedSchema(Schema):
```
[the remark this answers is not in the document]
SergioSim: Thanks, sorry I forgot you told me before, will be done

tests/fixtures/event/proposal_marshmallow.py
```diff
@@ -0,0 +1,179 @@
+"""
+Proposal for an
+edx.problem.hint.feedback_displayed event schema + fixture definition
+"""
+
+from faker import Faker
+from marshmallow import Schema, ValidationError, fields, validates_schema
+from marshmallow.validate import Equal, Length, OneOf
+
+FAKE = Faker()
```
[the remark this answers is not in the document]
SergioSim: Indeed, thanks

tests/fixtures/event/proposal_marshmallow.py
```diff
@@ -0,0 +1,179 @@
...
+
+    username = fields.Str(required=True, validate=Length(min=2))
+    event_type = fields.Str(
+        required=True,
+        validate=Equal(
+            comparable="edx.problem.hint.feedback_displayed",
+            error='The event event_type field is not "edx.problem.hint.feedback_displayed"',
+        ),
+    )
+    ip = fields.Str(required=True)
```
jmaupetit, line 160: Should we define a custom `IPAddressField`?

tests/fixtures/event/proposal_marshmallow.py
SergioSim: Much cleaner, thanks

tests/fixtures/event/proposal_marshmallow.py
SergioSim: Good Idea, thanks, will be done

tests/fixtures/event/mixins.py
```diff
@@ -0,0 +1,24 @@
+"""Mixins for event factories"""
+
+import factory
+
+
+class JSONFactoryMixin(factory.Factory):
```
SergioSim, line 6: It works without factory.Factory inheritance. Thanks a lot!

src/ralph/schemas/edx/base.py
```diff
@@ -0,0 +1,144 @@
...
+        try:
+            return IPv4Address(value)
+        except ValueError as error:
+            raise ValidationError("Invalid IPv4 Address") from error
+
+
+def validate_username(value):
+    """"check username field empty or 2-30 chars long"""
+    if len(value) == 1 or len(value) > 30:
+        raise ValidationError("username should be empty or between 2 and 30 chars long")
```
jmaupetit, line 110: Are you sure about this? Is this configurable in openedx?

jmaupetit: I prefer the following (extract from `FeedbackDisplayedSchema`):

```python
username = fields.Str(required=True, validate=Length(min=2, max=30))
```

tests/fixtures/event/mixins.py
```diff
@@ -0,0 +1,21 @@
+"""Mixins for event factories"""
+
+class JSONFactoryMixin():
```
jmaupetit, line 3:
```suggestion
class JSONFactoryMixin:
```

tests/fixtures/event/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        """Override the default ``_create`` with our custom call."""
+        schema = model_class()
+        print(kwargs)
+        return schema.dumps(kwargs)
+
+
+class ObjFactoryMixin():
```
jmaupetit, line 14:
```suggestion
class ObjFactoryMixin:
```

src/ralph/schemas/edx/base.py
SergioSim: The username can be empty if the user is not connected (for server events)

SergioSim: feedback_displayed events are triggered from pages containing capa problems - for course pages the user have to be connected - so the username can't be empty - if I understood is right

jmaupetit: So what about the following? Will this field be validated only if provided?

```python
username = fields.Str(required=False, validate=Length(min=2, max=30))
```

src/ralph/schemas/edx/base.py
```diff
@@ -0,0 +1,144 @@
...
+    def _deserialize(self, value, attr, data, **kwargs):
+        if value == "":
+            return ""
+        try:
+            return IPv4Address(value)
+        except ValueError as error:
+            raise ValidationError("Invalid IPv4 Address") from error
+
+
+def validate_username(value):
```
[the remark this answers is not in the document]
SergioSim: I have tried to make pylint happy in this case.
It seems validate_username can't be a static method for this usecase
but -
`
class BaseEventSchema(Schema):
    def validate_username(self, value):  # pylint: disable=no-self-use
`
Works.
I will change that, Thanks

src/ralph/schemas/edx/base.py
SergioSim: I think yes, but in our case the field is always provided.
I will give it a check, thanks

SergioSim: Unfortunately this seems to not be working when:
"username": ""

tests/test_event_base.py
```diff
@@ -33,8 +33,6 @@ def test_invalid_ip_should_raise_exception(base_events):
         base_events(1, ip="invalid_ip")
     with pytest.raises(ValidationError):
         base_events(1, ip="  ")
-    with pytest.raises(ValidationError):
-        base_events(1, ip="051.255.9.27")
```
SergioSim: As the regex ip validation was removed this test does not raise an exception and silently corrects the ip from "051.255.9.27" to "51.255.9.27".
This causes the "event" to change after deserialisation/serialisation.
If this is not a problem it's good for me - removing this test

src/ralph/schemas/edx/browser.py
```diff
@@ -227,18 +219,94 @@ def validate_event_textbook_pdf(self, data, **kwargs):
...
         try:
             event = json.loads(data["event"])
         except (json.JSONDecodeError, TypeError):
-            raise ValidationError("Event should contain a parsable json string")
-
-        required_keys = {"name", "page", "chapter"}
+            raise ValidationError("Event should contain a parsable JSON string")
         keys = set(event.keys())
-        if len(required_keys.intersection(keys)) != 3:
+        if keys != required_keys:
```
jmaupetit, line 298: I think this is not what we want, but rather:

```suggestion
        if len(required_keys.intersection(keys)) != len(required_keys):
```

WDYT?

src/ralph/schemas/edx/browser.py
```diff
@@ -227,18 +219,94 @@ def validate_event_textbook_pdf(self, data, **kwargs):
...
+    @validates("session")
+    def validate_session(self, value):  # pylint: no-self-use
+        """"Check session field empty or MD5_HASH_LEN chars long"""
+        if value != "" and len(value) != MD5_HASH_LEN:
+            raise ValidationError(
+                f"Session should be empty or {MD5_HASH_LEN} chars long (MD5)"
+            )
+
+    @staticmethod
+    def parse_json_check_keys(data, required_keys):
```
jmaupetit, line 289:
```suggestion
    def validate_event_keys(data, required_keys):
```

src/ralph/schemas/edx/browser.py
```diff
@@ -227,18 +219,94 @@ def validate_event_textbook_pdf(self, data, **kwargs):
...
+        """Parse JSON string and check that it contains the keys in the
+        `keys_array` and no other keys
```
jmaupetit, line 291: sure?

src/ralph/schemas/edx/browser.py
SergioSim: My concern here was that we might silently accept events having the required keys but also other keys that will be ignored in this case.
If this is not a problem, it's good for me, i will change that.

src/ralph/schemas/edx/browser.py
SergioSim: Good catch, thanks - required_keys is an set

tests/test_event_base.py
jmaupetit: It's not a problem to me.

src/ralph/schemas/edx/browser.py
jmaupetit: From your experience, does this occurs?

src/ralph/schemas/edx/browser.py
```diff
@@ -156,68 +162,59 @@ def validate_event_problem_graded_reset_save(self, data, **kwargs):
...
-        course_key = data["context"]["course_id"][10:]
-        if event["id"][:-32] != f"block-v1:{course_key}+type@sequential+block@":
+        block_id = self.get_block_id(data, block_type="sequential")
+        block_id_len = len(block_id)
+        if event["id"][:block_id_len] != block_id:
+            raise ValidationError(
+                f"The id should start with {block_id} , not "
+                f"{event['id'][:block_id_len]}"
+            )
+        if len(event["id"][block_id_len:]) != 32:
```
jmaupetit, line 185:
```suggestion
        if len(event["id"][block_id_len:]) != MD5_HASH_LEN:
```

src/ralph/schemas/edx/browser.py
SergioSim: If I understood it right, the event field is created by the client then send to the django back-end and in the back-end there is no validation (it just tries to parse the json if it's a string)
I suggest the event field can't be trusted as it might contain any values.

src/ralph/schemas/edx/browser.py
SergioSim: Good catch, thanks, can I commit your suggestion?

jmaupetit: Yes!

src/ralph/schemas/edx/browser.py
jmaupetit: Ok, so feel free to be more stringent then.

src/ralph/schemas/edx/browser.py
```diff
@@ -0,0 +1,330 @@
...
+        """
+        if data["name"] != "textbook.pdf.zoom.menu.changed":
+            return
+
+        event = self.validate_event_keys(data, {"chapter", "page", "name", "amaunt"})
+        self.check_chapter_page_name(data, event)
+        if event["amaunt"] not in BROWSER_EVENT_VALID_AMAUNT:
+            raise ValidationError(
+                f"amaunt should be one of {BROWSER_EVENT_VALID_AMAUNT}"
+            )
```
jmaupetit, line 279: IMO this is not a good idea to have a single `BrowserEventSchema` with x schema validation methods. This is not efficient at all! I would prefer having 1 schema per event and a `BaseBrowserEventSchema` to inherit from.

src/ralph/schemas/edx/base.py
```diff
@@ -0,0 +1,167 @@
...
+            raise ValidationError(
+                "user_id should be None or empty string or an Integer"
+            )
+
+    @validates_schema
+    def validate_course_id(self, data, **kwargs):
+        """the course_id should be equal to
+        "course-v1:{org_id}+{any_string}+{any_string}"
+        """
+        if data["org_id"] == "" and data["course_id"] != "":
```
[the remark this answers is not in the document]
jmaupetit: You should apply the same pattern everywhere.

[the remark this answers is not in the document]
SergioSim: Thanks for the review, I will change that

src/ralph/schemas/edx/browser.py
```diff
@@ -0,0 +1,330 @@
...
+BROWSER_NAME_FIELD = [
+    "textbook.pdf.page.loaded",
+    "textbook.pdf.page.navigatednext",
+    "textbook.pdf.search.executed",
+    "textbook.pdf.search.highlight.toggled",
+    "textbook.pdf.search.navigatednext",
+    "textbook.pdf.searchcasesensitivity.toggled",
+]
+
+BROWSER_EVENT_VALID_AMAUNT = [
```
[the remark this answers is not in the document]
SergioSim: Good catch Julien, I made the mistake in my notes too, will change that

src/ralph/schemas/edx/browser.py
SergioSim: Good Idea, I will give it a try, thanks

slandrault:
I didn't see any circle-ci config file. Maybe we could add one to run the test suite (and use it as a base in the future for more checks / build)? What do you think @jmaupetit ? Or maybe in a separate PR to also include some linters like in others openfun projects?

SergioSim:
> I didn't see any circle-ci config file. Maybe we could add one to run the test suite (and use it as a base in the future for more checks / build)? What do you think @jmaupetit ? Or maybe in a separate PR to also include some linters like in others openfun projects?

Hello @slandrault ,
do you search this config file https://github.com/openfun/ralph/blob/master/.circleci/config.yml ?

slandrault:
> > I didn't see any circle-ci config file. Maybe we could add one to run the test suite (and use it as a base in the future for more checks / build)? What do you think @jmaupetit ? Or maybe in a separate PR to also include some linters like in others openfun projects?
> 
> Hello @slandrault ,
> do you search this config file https://github.com/openfun/ralph/blob/master/.circleci/config.yml ?

:sweat_smile: Awesome, I don't know why I didn't see it, I was probably on the wrong branch. All good then, sorry about that!

tests/fixtures/edx/browser.py
```diff
@@ -0,0 +1,307 @@
+"""
+Broser event factory definitions
```
slandrault, line 2: s/Broser/Browser

SergioSim: Good catch, thanks, will update this

SergioSim:
Closing this pull request as it will be split into smaller pieces.
The first part is #59

PR #5, opened 2020-10-05
Add LDP backend and base CLI

jmaupetit:
### Purpose

We need to automate fetching new LDP-stored events and push them to our ES instance. To achieve this, `ralph` should be used as a stream-based CLI that can pull / convert / push learning events from a storage backend to a database backend.

### Proposal

Current work mostly focuses on defining a shared architecture for parsers, storage backends and CLI commands. An example usage follows:

```
$ ralph list -b ldp
ARCHIVE_ID_01
ARCHIVE_ID_02
ARCHIVE_ID_03
ARCHIVE_ID_04
ARCHIVE_ID_05

$ ralph fetch -b ldp ARCHIVE_ID_02 | gunzip | ralph extract -p gelf
{
  ... event ...
}
```

- [x] implement LDP storage backend
- [x] add base CLI structure
- [x] add the `list` command
- [x] add the `fetch` command
- [x] add the `extract` command
- [x] add history management to be able to list only new archives (`list --new`)
- [x] add tests
- [x] document project bootstrapping
- [x] update the changelog :sweat_smile:

### DIFF
--- a/bin/ralph
+++ b/bin/ralph
@@ -3,4 +3,4 @@
 declare DOCKER_USER
 DOCKER_USER="$(id -u):$(id -g)"
 
-DOCKER_USER=${DOCKER_USER} docker-compose run ralph
+DOCKER_USER=${DOCKER_USER} docker-compose run --rm -T app ralph ${@}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.4'
 
 services:
-  ralph:
+  app:
     build:
       context: .
       target: development
@@ -9,6 +9,8 @@ services:
         DOCKER_USER: ${DOCKER_USER:-1000}
     user: ${DOCKER_USER:-1000}
     image: ralph:development
+    env_file:
+      - .env
     environment:
       PYLINTHOME: /app/.pylint.d
     volumes:
--- a/src/ralph/backends/mixins.py
+++ b/src/ralph/backends/mixins.py
@@ -1,8 +1,11 @@
 """Backend mixins for Ralph"""
 
 import json
+import logging
 
 from ralph.defaults import HISTORY_FILE
+
+logger = logging.getLogger(__name__)
 
 
 class HistoryMixin:
@@ -12,6 +15,9 @@ class HistoryMixin:
     @property
     def history(self):
         """Get backend history"""
+
+        logging.debug("Loading history file: %s", str(HISTORY_FILE))
+
         if not hasattr(self, "_history"):
             try:
                 with HISTORY_FILE.open() as history_file:
@@ -23,6 +29,8 @@ class HistoryMixin:
     # pylint: disable=no-self-use
     def write_history(self, history):
         """Write given history as a JSON file"""
+
+        logging.debug("Writting history file: %s", str(HISTORY_FILE))
 
         if not HISTORY_FILE.parent.exists():
             HISTORY_FILE.parent.mkdir(parents=True)
--- a/src/ralph/backends/mixins.py
+++ b/src/ralph/backends/mixins.py
@@ -38,6 +38,9 @@ class HistoryMixin:
         with HISTORY_FILE.open("w") as history_file:
             json.dump(history, history_file)
 
+        # Update history
+        self._history = history
+
     def clean_history(self, selector):
         """Clean selected events from the history.
 
--- a/src/ralph/backends/mixins.py
+++ b/src/ralph/backends/mixins.py
@@ -1,0 +1,44 @@
+"""Backend mixins for Ralph"""
+
+import json
+
+from ralph.defaults import HISTORY_FILE
+
+
+class HistoryMixin:
+    """Handle backend download history to avoid fetching same files multiple
+    times if they are already available."""
+
+    @property
+    def history(self):
+        """Get backend history"""
+        if not hasattr(self, "_history"):
+            try:
+                with HISTORY_FILE.open() as history_file:
+                    self._history = json.load(history_file)
+            except FileNotFoundError:
+                self._history = []
+        return self._history
+
+    # pylint: disable=no-self-use
+    def write_history(self, history):
+        """Write given history as a JSON file"""
+
+        if not HISTORY_FILE.parent.exists():
+            HISTORY_FILE.parent.mkdir(parents=True)
+
+        with HISTORY_FILE.open("w") as history_file:
+            json.dump(history, history_file)
+
+    def clean_history(self, selector):
+        """Clean selected events from the history.
+
+        selector: a callable that selects events that need to be removed
+        """
+        self._history = list(filter(lambda event: not selector(event), self.history))
+        self.write_history(self._history)
+
+    def append_to_history(self, event):
+        """Append event to history"""
+
+        self.write_history(self.history + [event])
--- a/src/ralph/backends/storage/__init__.py
+++ b/src/ralph/backends/storage/__init__.py
@@ -1,0 +1,1 @@
+"""Storage backends for Ralph"""
--- a/src/ralph/backends/storage/ldp.py
+++ b/src/ralph/backends/storage/ldp.py
@@ -1,7 +1,6 @@
 """OVH's LDP storage backend for Ralph"""
 
 import logging
-import re
 import sys
 
 import ovh
@@ -64,16 +63,20 @@ class LDPStorage(HistoryMixin, BaseStorage):
 
         return download_url
 
-    def list(self):
+    def list(self, details=False):
         """List archives for a given stream"""
 
         list_archives_endpoint = self._archive_endpoint
         logger.debug("List archives endpoint: %s", list_archives_endpoint)
+        logger.debug("List archives details: %s", str(details))
 
         archives = self.client.get(list_archives_endpoint)
         logger.debug("Found %d archives", len(archives))
 
-        return archives
+        for archive in archives:
+            yield self.client.get(
+                f"{self._archive_endpoint}/{archive}"
+            ) if details else archive
 
     def read(self, name, chunk_size=4096):
         """Read the `name` archive file and stream its content"""
--- a/src/ralph/backends/storage/ldp.py
+++ b/src/ralph/backends/storage/ldp.py
@@ -1,8 +1,8 @@
 """OVH's LDP storage backend for Ralph"""
 
+import datetime
 import logging
 import sys
-from datetime import datetime, timezone
 
 import ovh
 import requests
--- a/src/ralph/backends/storage/ldp.py
+++ b/src/ralph/backends/storage/ldp.py
@@ -2,6 +2,7 @@
 
 import logging
 import sys
+from datetime import datetime, timezone
 
 import ovh
 import requests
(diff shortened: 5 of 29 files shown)


src/ralph/cli.py
```diff
@@ -0,0 +1,124 @@
...
+    "--parser",
+    type=click.Choice(PARSERS),
+    required=True,
+    help="Container format parser used to extract events",
+)
+@click.option(
+    "-c",
+    "--chunksize",
+    type=int,
+    default=50000,
```
slandrault, line 80: According to the docstring in the GELFParser.parse method, the recommend benchmarked chunksize is below 10k. Maybe this default should be adjusted?

src/ralph/parsers.py
```diff
@@ -36,45 +37,31 @@ class GELFParser(BaseParser):
     documentation: https://docs.graylog.org/en/latest/pages/gelf.html
     """
 
-    def parse(self, input_file, filters=None, chunksize=50000):
+    name = "gelf"
+
+    def parse(self, input_file, chunksize=50000):
```
slandrault, line 42: According to the docstring the recommend benchmarked chunksize is <10k. Maybe this default should be adjusted?

docker-compose.yml
```diff
@@ -9,6 +9,8 @@ services:
         DOCKER_USER: ${DOCKER_USER:-1000}
     user: ${DOCKER_USER:-1000}
     image: ralph:development
+    env_file:
+      - .env
```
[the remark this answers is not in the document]
jmaupetit: Good point. This is what I wanted to document for the project's bootstrapping.

src/ralph/backends/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+
+import json
+import logging
+
+from ralph.defaults import HISTORY_FILE
+
+logger = logging.getLogger(__name__)
+
+
+class HistoryMixin:
```
sampaccoud, line 11: Do we need to a way to clean history without knowing where the file is?

src/ralph/cli.py
```diff
@@ -0,0 +1,139 @@
...
+    logger.debug("Backend parameters: %s", options)
+
+    storage = get_instance_from_class(
+        get_class_from_name(backend, StorageBackends), **options
+    )
+    archives = storage.list(details=details, new=new)
+
+    counter = 0
+    for archive in archives:
+        click.echo(json.dumps(archive) if details else archive)
```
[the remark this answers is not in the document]
sampaccoud: I understood this later and I think it's ok. That's like 2 modes for the tool. It requires documentation though.

src/ralph/backends/mixins.py
jmaupetit: Good idea!

src/ralph/backends/storage/ldp.py
```diff
@@ -0,0 +1,137 @@
...
+        # Stream response (archive content)
+        with requests.get(self.url(name), stream=True) as result:
+            result.raise_for_status()
+            chunk_size = 4096
+            for chunk in result.iter_content(chunk_size=chunk_size):
+                sys.stdout.buffer.write(chunk)
+
+        # Archive is supposed to have been fully fetched, add a new entry to
+        # the history.
+        self.write_history(
```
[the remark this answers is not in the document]
jmaupetit: In my design, each backend may write to this history file. Current implementation allows to list only "new" archives from a storage backend, but we can imagine the same mechanism for the database backend and also prevent writing entries already pushed.

bin/ralph
```diff
@@ -3,4 +3,4 @@
 declare DOCKER_USER
 DOCKER_USER="$(id -u):$(id -g)"
 
-DOCKER_USER=${DOCKER_USER} docker-compose run ralph
+DOCKER_USER=${DOCKER_USER} docker-compose run --rm -T ralph ralph ${@}
```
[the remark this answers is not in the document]
jmaupetit: That's what I did in a first attempt, but doing so prevent running other commands in the container. This not ideal for development. I think I will add our standard entrypoint using ralph only for the `production` target.

src/ralph/cli.py
```diff
@@ -0,0 +1,139 @@
...
+
+@click.argument("archive")
+@backends_options(backends=STORAGE_BACKENDS)
+def fetch(backend, archive, **options):
+    """Fetch an archive from a configured storage backend"""
+
+    logger.info("Fetching archive %s from the configured %s backend", archive, backend)
+    logger.debug("Backend parameters: %s", options)
+
+    storage = get_instance_from_class(
```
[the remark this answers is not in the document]
jmaupetit: Not sure about this as this utility has been developed to automate backends integration in the CLI. Adding this to the class' init would make the class usage more confusing/convoluted. WDYT?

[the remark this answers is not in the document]
sampaccoud: Up to you to consider. I'm just making sure you considered the option ;-)

bin/ralph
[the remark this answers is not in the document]
sampaccoud: What I understand is that the container is not only ralph then... it should be named differently?

src/ralph/cli.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+    archives = storage.list(details=details, new=new)
+
+    counter = 0
+    for archive in archives:
+        click.echo(json.dumps(archive) if details else archive)
+        counter += 1
+
+    if counter == 0:
+        logger.warning("Configured %s backend contains no archive", backend)
```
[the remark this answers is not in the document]
jmaupetit: `archives` is an iterator, not a list. The following is never `True`.

```python
    if not archive:
        logger.warning("Configured %s backend contains no archive", backend)
```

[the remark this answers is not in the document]
sampaccoud: ah ok. Maybe don't count but just set a boolean?

src/ralph/backends/mixins.py
jmaupetit: Done.

docker-compose.yml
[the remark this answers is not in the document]
jmaupetit: Done!

bin/ralph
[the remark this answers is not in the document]
jmaupetit: It can be named `app` then.

src/ralph/cli.py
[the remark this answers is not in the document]
jmaupetit: Yes I did. :pray:

bin/ralph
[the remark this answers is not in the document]
jmaupetit: Done.

src/ralph/cli.py
jmaupetit: I've added a `DEFAULT_GELF_PARSER_CHUNCK_SIZE` variable for this.

src/ralph/parsers.py
jmaupetit: ditto

src/ralph/backends/storage/__init__.py
```diff
@@ -0,0 +1 @@
+"""Storage backends for Ralph"""
```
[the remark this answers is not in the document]
jmaupetit: I think you are totally right. Let's move to storage backends only. Some may be read-only by not implementing a `write` method.

src/ralph/backends/storage/ldp.py
```diff
@@ -0,0 +1,139 @@
...
+            self.history
```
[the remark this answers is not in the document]
jmaupetit: Good idea! Let's do this.

src/ralph/backends/storage/ldp.py
```diff
@@ -0,0 +1,139 @@
...
+
+        logger.debug("Getting archive: %s", name)
+
+        # Get detailled information about the archive to fetch
+        details = self._details(name)
+
+        # Stream response (archive content)
+        with requests.get(self.url(name), stream=True) as result:
+            result.raise_for_status()
+            chunk_size = 4096
```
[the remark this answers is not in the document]
jmaupetit: Well spotted. I'll make it configurable.

jmaupetit:
@sampaccoud @madmatah I've addressed all your issues. I'm ready for a final review :pray:

tests/test_parsers.py
```diff
@@ -2,27 +2,31 @@
...
 
-import pandas as pd
 import pytest
 
-from ralph import filters
 from ralph.parsers import BaseParser, GELFParser
 
 
 def test_baseparser_parse():
     """Test that calling parse on the BaseParser raises a NotImplementedError."""
```
lunika, line 13: IMO this test is useless. You can remove it.

tests/test_parsers.py
```diff
@@ -2,27 +2,31 @@
...
+        match="Can't instantiate abstract class BaseParser with abstract methods parse",
+    ):
         BaseParser().parse("foo.log")
 
     # pylint: disable=abstract-method
     class MyParser(BaseParser):
         """Dummy parser."""
 
-    with pytest.raises(NotImplementedError):
+    with pytest.raises(
```
lunika, line 26: Same here, you can remove this test. All of this is in the python engine, you can I think trust it.

src/ralph/backends/storage/ldp.py
```diff
@@ -0,0 +1,90 @@
...
+
+        self.client = ovh.Client(
+            endpoint=self.endpoint,
+            application_key=self.application_key,
+            application_secret=self.application_secret,
+            consumer_key=self.consumer_key,
+        )
+
+    @property
+    def _archive_endpoint(self):
```
lunika, line 47: I don't understand why you are creating a property function here ? This property can be created in the constructor like you did for the `client` property.

tests/backends/storage/test_base.py
```diff
@@ -0,0 +1,44 @@
+"""Tests for Ralph base storage backend"""
+
+import pytest
+
+from ralph.backends.storage.base import BaseStorage
+
+
+def test_abstract_interface_with_missing_abstract_methods():
```
lunika, line 8: IMO this test is useless.

src/ralph/backends/storage/ldp.py
jmaupetit: Because it should be dynamical as it depends on the `service_name` and `stream_id` attributes that may change after `LDPStorage` class instantiation.

tests/test_parsers.py
jmaupetit: You are right.

src/ralph/backends/storage/ldp.py
madmatah: I see your point @jmaupetit, but I agree with @lunika on this.

How can `service_name` and `stream_id` attributes change after instantiation, but not `application_key`, `application_secret` and `consumer_key` ? They are declared the same way and there is no special setter methods for `service_name` and `stream_id`.

jmaupetit: As discussed, I have moved client instantiation arguments as private to clarify object usage.

src/ralph/backends/storage/ldp.py
```diff
@@ -0,0 +1,142 @@
...
+class LDPStorage(HistoryMixin, BaseStorage):
+    """OVH's LDP storage backend"""
+
+    # pylint: disable=too-many-arguments
+
+    name = "ldp"
+
+    def __init__(
+        self,
+        endpoint=None,
```
lunika, line 25: Why providing a default value ? The endpoint value looks like required and stored in a "private" property.

jmaupetit: This is tricky but storage-specific parameters should not be arguments but keywords.

lunika: Yes but semantically this is wrong.

jmaupetit: In fact, making them arguments is not breaking the parameters discovery mechanism as you can call arguments as keywords in python.

madmatah (approved):

lunika (approved):

PR #12, opened 2020-10-19
⬆️(project) move to a python3 compatible GELF library

jmaupetit:
### Purpose

Djehouty is a legacy library theoretically only compatible with python2. Hat tip to @pyup-bot 

### Proposal

Djehouty has been rewritten by the OVH team to be fully compatible with python3. Let's switch to it even if we only use the GELF formatter for testing.


madmatah (approved):

slandrault:
So its a "one to one" replacement? That's sweet!

PR #25, opened 2020-11-06
Add file system storage and push command

Sifflex:
## Purpose

The purpose of ralph is to be able to store archives in many formats, and the filesystem one was lacking.

## Proposal
- [x] add a FileSystem storage
- [x] add a `push` command to interact implemented backends

Related to #16

PR #27, opened 2020-11-12
✨(backends) add Elasticsearch database backend

### LINKED ISSUE #16: Add push command
### Purpose

Ralph should be able to write to storage backends.

### Proposal

Implement a generic `push` command that takes a storage backend (and its configuration) as arguments and streams the standard input to this backend.

### LINKED ISSUE #17: Add ElasticSearch backend
### Purpose

Ralph should be able to read/write from ES indexes.

### Proposal

Implement a new storage backend with read and write features to ES clusters.

jmaupetit:
## Purpose

Ralph should be able to publish or get statements from database backends; our first implementation of such support should be Elasticsearch since its used in production to mine collected xAPI statements.

## Proposal

- [x] add `ESDatabase` backend
- [x] adapt the `fetch` command to support both database & storage backends ; to get all database records (from an index, a collection, a table depending on supported backend)
- [x] adapt the `push` command to support both database & storage backends
- [x] add an `elasticsearch` service in the CI to run tests

Fix #16 
Fix #17

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,6 +170,12 @@ jobs:
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
+      - image: elasticsearch:7.8.1
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+        environment:
+          discovery.type: single-node
     working_directory: ~/fun
     steps:
       - checkout
@@ -178,7 +184,11 @@ jobs:
             - v1-dependencies-{{ .Revision }}
       - run:
           name: Run tests
-          command: ~/.local/bin/pytest
+          command: |
+            dockerize \
+              -wait tcp://${HOSTNAME}:9200 \
+              -timeout 60s \
+              ~/.local/bin/pytest
 
   # ---- Packaging jobs ----
   package:
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,13 @@ COMPOSE_RUN          = $(COMPOSE) run --rm
 COMPOSE_TEST_RUN     = $(COMPOSE_RUN)
 COMPOSE_TEST_RUN_APP = $(COMPOSE_TEST_RUN) app
 
+# -- Elasticsearch
+ES_PROTOCOL = http
+ES_HOST     = localhost
+ES_PORT     = 9200
+ES_INDEX    = statements
+ES_URL      = $(ES_PROTOCOL)://$(ES_HOST):$(ES_PORT)
+
 # ==============================================================================
 # RULES
 
@@ -21,7 +28,8 @@ bootstrap: ## bootstrap the project for development
 bootstrap: \
   .env \
   build \
-  dev
+  dev \
+  es-index
 .PHONY: bootstrap
 build: ## build the app container
 	@$(COMPOSE) build app
@@ -30,6 +38,16 @@ build: ## build the app container
 dev: ## perform editable install from mounted project sources
 	DOCKER_USER=0 docker-compose run --rm app pip install -e ".[dev]"
 .PHONY: dev
+
+down: ## stop and remove backend containers
+	@$(COMPOSE) down
+.PHONY: down
+
+es-index:  ## create elasticsearch index and sample documents
+es-index: run
+	@echo "Creating $(ES_INDEX) index"
+	bin/es index $(ES_INDEX)
+.PHONY: es-index
 
 # Nota bene: Black should come after isort just in case they don't agree...
 lint: ## lint back-end python sources
@@ -70,11 +88,22 @@ logs: ## display app logs (follow mode)
 	@$(COMPOSE) logs -f app
 .PHONY: logs
 
+run: ## starts supported backends servers
+	@$(COMPOSE) up -d
+	@echo "Waiting for elasticsearch to be up and running..."
+	@$(COMPOSE_RUN) dockerize -wait tcp://elasticsearch:9200 -timeout 60s
+.PHONY: run
+
 status: ## an alias for "docker-compose ps"
 	@$(COMPOSE) ps
 .PHONY: status
 
+stop: ## stops backend servers
+	@$(COMPOSE) stop
+.PHONY: stop
+
 test: ## run back-end tests
+test: run
 	bin/pytest
 .PHONY: test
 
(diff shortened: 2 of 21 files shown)


jmaupetit:
This PR is blocked while #25 has not been merged.

src/ralph/backends/database/es.py
```diff
@@ -0,0 +1,66 @@
...
+    name = "es"
+
+    def __init__(self, hosts, index):
+        """Instanciate the Elasticsearch client.
+
+        hosts is supposed to be a list
+
+        """
+
+        self._hosts = hosts
```
SergioSim, line 27: self._hosts is not used ?
```suggestion

```

bin/es
```diff
@@ -0,0 +1,53 @@
...
+  local index="${1:-${ES_INDEX}}"
+
+  # timestamp in ms (as required by elasticsearch)
+  local now
+  now=$(echo "$(date +%s.%N) * 1000" | bc)
+
+  for i in $(seq 100); do
+    uuid=$(uuidgen)
+    # Simulate one event per minute
+    at=$(echo "${now} - (60 * 1000 * ${i})" | bc)
```
madmatah, line 36: Another alternative here, you can remove the `$now` variable and use the following :

```suggestion
    at=$(date -d "now -${i}min" +"%s.%N000")
```

The only difference is that `now` will not be a fixed value and will be relative to the time when the command is executed. In your use case i don't think it's a problem.

madmatah (approved):

Makefile
```diff
@@ -70,16 +90,24 @@ logs: ## display app logs (follow mode)
 	@$(COMPOSE) logs -f app
 .PHONY: logs
 
+run: ## starts supported backends servers
+	@$(COMPOSE) up -d
```
lunika, line 94: You could also call dockerize to wait ES to be ready ?

src/ralph/cli.py
```diff
@@ -91,41 +103,76 @@ def extract(parser, chunksize):
...
+        archive,
+        chunk_size,
+    )
     logger.debug("Backend parameters: %s", options)
 
-    storage = get_instance_from_class(
-        get_class_from_name(backend, StorageBackends), **options
-    )
-    storage.read(archive)
+    klass = get_class_from_name(backend, BACKENDS)
```
lunika, line 132: the typo is intended ? `class` is a reserved name but you can maybe found something more relevant ?

src/ralph/utils.py
```diff
@@ -29,11 +33,27 @@ def import_string(dotted_path):
         ) from err
 
 
+def get_backend_type(klass):
```
lunika, line 36: same here, `klass` is not relevant

src/ralph/backends/database/es.py
```diff
@@ -0,0 +1,66 @@
...
+
+from .base import BaseDatabase
+
+logger = logging.getLogger(__name__)
+
+
+class ESDatabase(BaseDatabase):
+    """Elasticsearch database backend"""
+
+    name = "es"
```
lunika, line 18: If you want to explore more OOP, you can declare a `@property` method returning this value ? You are sure this value will never change and you can also declare it as an abstract method in the abstract class. WDYT ?

tests/backends/database/test_es.py
```diff
@@ -0,0 +1,128 @@
...
+        (
+            "http://elasticsearch:9200" in database._hosts,
+            "http://localhost:9200" in database._hosts,
+        )
+    )
+    assert database.index == ES_TEST_INDEX
+    assert isinstance(database.client, Elasticsearch)
+
+
+def test_to_documents_method(es):
```
lunika, line 41: IMO a private method should not be test. You will test it through the public one using it.

Sifflex (approved):

src/ralph/backends/database/es.py
jmaupetit: It's a class attribute. It is not supposed to change.

tests/backends/database/test_es.py
jmaupetit: Isolated a piece of code in a private method is to make it easily reusable AND testable. I'm not sure it shouldn't be tested.

src/ralph/cli.py
jmaupetit: This is not a typo :sweat_smile: I'll make it more explicit.

.circleci/config.yml
```diff
@@ -178,7 +184,11 @@ jobs:
             - v1-dependencies-{{ .Revision }}
       - run:
           name: Run tests
-          command: ~/.local/bin/pytest
+          command: |
+            dockerize \
+              -wait tcp://${HOSTNAME}:9200 \
```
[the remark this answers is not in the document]
jmaupetit: It is a shell built-in variable as the `PATH` variable for example.

bin/es
jmaupetit: Wow didn't know about this pattern, thanks.

src/ralph/backends/database/es.py
jmaupetit: You are right, I should use it to instantiate the Elasticsearch client.

bin/es
```diff
@@ -0,0 +1,53 @@
...
+
+
+function index(){
+
+  local data
+  local index="${1:-${ES_INDEX}}"
+
+  # timestamp in ms (as required by elasticsearch)
+  local now
+  now=$(echo "$(date +%s.%N) * 1000" | bc)
```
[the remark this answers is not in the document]
jmaupetit: I might be wrong but this add zeros at the end of the timestamp instead of multiplying it by 10^3.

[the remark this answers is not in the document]
madmatah: You're right, I misinterpreted your code.

jmaupetit:
Thank you all for your reviews :pray: @lunika waiting for your response/approval before merging this.

src/ralph/backends/database/es.py
lunika: well maybe but nothing prevent you to a change of this attribute.

tests/backends/database/test_es.py
lunika: > solated a piece of code in a private method is to make it easily reusable and testable
I disagree, why not implementing a public method in that case ? 

You are testing twice.

lunika: A unit test is a client of the object you are testing. Never a client will access your private method, if python implement real private method you will see how difficult it is to test a private method.

jmaupetit: Ok, I've made this method public instead.

lunika (approved):

PR #29, opened 2020-11-19
✨(converters) add edx server event xapi converter

SergioSim:
## Purpose

The edx server events  are the most simple and common events in the edx tracking logs. They are logged at almost all requests made to the edx lms. We want to convert them into xAPI format.

## Proposal

`ServerEventSchema` validates the edx server event.

`ServerEventToXapi` uses the ServerEventSchema to validate the edx server event and then converts it to xAPI format.

`XapiConverterSelector` deserializes the edx events and tries to match a known converter for the event conversion.


src/ralph/converters/xapi/base.py
```diff
@@ -0,0 +1,126 @@
...
+        extensions = {
+            XAPI_EXTENSION_ACCEPT_LANGUAGE: self.edx_event["accept_language"],
+            XAPI_EXTENSION_AGENT: self.edx_event["agent"],
+            XAPI_EXTENSION_COURSE_ID: context["course_id"],
+            XAPI_EXTENSION_COURSE_USER_TAGS: context.get("course_user_tags", {}),
+            XAPI_EXTENSION_HOST: self.edx_event["host"],
+            XAPI_EXTENSION_IP: event_ip,
+            XAPI_EXTENSION_ORG_ID: context["org_id"],
+            XAPI_EXTENSION_REFERER: self.edx_event["referer"],
+            XAPI_EXTENSION_USER_ID: context["user_id"],
```
SergioSim, line 115: I haven't found a specific place where to put this information in the xAPI statement so I put it in the context.extentions

Although the [spec](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#requirements-18) states that excessive usage of extensions is meaningless - It's preferable to keep those details somewhere in the xAPI statement (to be able to convert back to edx format for example). WDYT?

SergioSim, line 115: XAPI_EXTENSION_USER_ID can be null and null values are removed from xApi statements.
Do you prefer to have an empty XAPI_EXTENSION_USER_ID value instead or we can keep this behaviour?

src/ralph/converters/xapi/server.py
```diff
@@ -0,0 +1,40 @@
...
+
+class ServerXapiConverter(ServerEventSchema, BaseXapiConverter):
+    """Converts a common edx server event to Xapi
+    See ServerEventSchema for info about the Edx server event
+    Example Statement: John viewed https://www.fun-mooc.fr/ Web page
+    """
+
+    def get_verb_kwargs(self):  # pylint: disable=no-self-use
+        """Returns tincan.Verb kwargs for the xapi_event"""
+        return {"id": XAPI_VERB_VIEWED, "display": LanguageMap({EN: VIEWED})}
```
SergioSim, line 28: May be http://id.tincanapi.com/verb/viewed is not the best option to describe the user action (verb) ?
(as edx logs server events for almost any kind of request - example the requests for the translations script: /i18n.js)

Other options could be:
http://adlnet.gov/expapi/verbs/launched
https://w3id.org/xapi/adb/verbs/requested
https://w3id.org/xapi/adb/verbs/demanded

WDYT?

src/ralph/converters/xapi_converter.py
```diff
@@ -0,0 +1,67 @@
...
+    """Uses xapi converters to convert to xAPI format"""
+
+    def __init__(self, event, platform):
+        """Initialise Xapi Converter
+
+        Args:
+            event (str): event to convert
+            platform (str): url of the platform to which the event belongs
+        """
+        BaseXapiConverter._platform = platform
```
SergioSim, line 30: I don't know why but pylint don't complain about protected-access here. Any Ideas ? May be we could keep it this way or it's better to extract this in a public function ?

tests/converters/test_base_converter.py
```diff
@@ -0,0 +1,244 @@
...
+    assert converter.convert(context_module) == {"baz": "change"}
+
+    # We may alter the destination path depending on other fields values
+    context_module["display_name"] = "display_name_value"
+    converter.display_name = GoTo(
+        Link(
+            [KwargPath("usage_key_value", ["usage_key"])],
+            lambda x, usage_key_value: [x, usage_key_value],
+        )
+    )
```
SergioSim, line 201: Some suggestions to change the name of the Link class

```suggestion
    converter.display_name = GoTo(
        With(
            [KwargPath("usage_key_value", ["usage_key"])],
            lambda x, usage_key_value: [x, usage_key_value],
        )
    )
```
```suggestion
    converter.display_name = GoTo(
        Dependency(
            [KwargPath("usage_key_value", ["usage_key"])],
            lambda x, usage_key_value: [x, usage_key_value],
        )
    )
```
```suggestion
    converter.display_name = GoTo(
        Including(
            [KwargPath("usage_key_value", ["usage_key"])],
            lambda x, usage_key_value: [x, usage_key_value],
        )
    )
```
```suggestion
    converter.display_name = GoTo(
        Attach(
            [KwargPath("usage_key_value", ["usage_key"])],
            lambda x, usage_key_value: [x, usage_key_value],
        )
    )
```

tests/converters/test_base_converter.py
```diff
@@ -0,0 +1,244 @@
...
+    ):
+        converter.convert(context_module)
+
+    converter.display_name = GoTo(["foo"])
+    # Now only usage_key is missing
+    with pytest.raises(
+        AttributeError, match="'BaseConverter' object has no attribute 'usage_key'"
+    ):
+        converter.convert(context_module)
+    converter.usage_key = GoTo(["bar"])
```
SergioSim, line 156: Some suggestions to change the name of the GoTo class

```suggestion
    converter.usage_key = MoveTo(["bar"])
```
```suggestion
    converter.usage_key = At(["bar"])
```
```suggestion
    converter.usage_key = FieldAt(["bar"])
```


```suggestion
    converter.usage_key = PathValue(["bar"])
```

SergioSim:
### Some Benchmarks:

I have generated a file "log.txt" of size 1,1G (1 million lines) repeating the gelf log containing a simple server event:

> {"version": "1.1", "host": "example.org", "short_message": "{\"username\":\"someUsername\",\"event_type\":\"/courses/course-v1:organisationCourse+numeroCours+sessiondCours/courseware/b5722732abed4eb7ba823f0b4239d50d/e2fa42fa1ca645389ee4618c86e982ff/\",\"ip\":\"172.20.0.1\",\"agent\":\"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0\",\"host\":\"7a4a1c08e030\",\"referer\":\"http://localhost:8072/courses/course-v1:organisationCourse+n…32abed4eb7ba823f0b4239d50d/e2fa42fa1ca645389ee4618c86e982ff/\",\"accept_language\":\"en-US,en;q=0.5\",\"event\":\"{\\\"POST\\\": {}, \\\"GET\\\": {}}\",\"event_source\":\"server\",\"context\":{\"course_user_tags\":{},\"user_id\":2,\"org_id\":\"organisationCourse\",\"course_id\":\"course-v1:organisationCourse+numeroCours+sessiondCours\",\"path\":\"/courses/course-v1:organisationCourse+numeroCours+sessiondCours/courseware/b5722732abed4eb7ba823f0b4239d50d/e2fa42fa1ca645389ee4618c86e982ff/\"},\"time\":\"2020-04-18T13:23:03.357329+00:00\",\"page\":null}", "timestamp": 1385053862.3072, "level": 1, "_user_id": 9001, "_some_info": "foo", "_some_env_var": "bar"}

And then used the "perf" command to run the performance tests:

**1) extract command without resultcallback vs. extract command with resultcallback**

- ralph extract -p gelf (the original version without using resultcallback)

`$ sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat log.txt | ralph -v CRITICAL extract -p gelf > /dev/null 2>&1";`

    PARSE without resultcallback
 
    280,27 msec task-clock          #    0,014 CPUs utilized
    849         context-switches    #    0,003 M/sec
    10          cpu-migrations      #    0,036 K/sec
    21 480      page-faults         #    0,077 M/sec
    429 201 469 L1-dcache-loads     # 1531,410 M/sec

    19,944762601 seconds time elapsed

    PARSE with resultcallback

    278,75 msec task-clock          #    0,014 CPUs utilized
    728         context-switches    #    0,003 M/sec
    20          cpu-migrations      #    0,072 K/sec
    21 458      page-faults         #    0,077 M/sec
    415 022 513 L1-dcache-loads     # 1488,870 M/sec

    20,136412941 seconds time elapsed

= It seems that using resultcallback makes ralph run a bit slower (difference of 0.19 sec = ~1% slower)

**2) extract to-xapi command vs extract | convert command:**

`$ echo "TO-XAPI"; sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat log.txt | ralph -v CRITICAL extract -p gelf to-xapi -p https://fun-mooc.fr > /dev/null 2>&1"; echo "CONVERT" ; sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat log.txt | ralph -v CRITICAL extract -p gelf | ralph -v CRITICAL convert -p https://fun-mooc.fr > /dev/null 2>&1"`

    TO-XAPI

    271,03 msec task-clock          #    0,001 CPUs utilized
    1 608       context-switches    #    0,006 M/sec
    31          cpu-migrations      #    0,114 K/sec
    21 572      page-faults         #    0,080 M/sec
    440 827 545 L1-dcache-loads     # 1626,474 M/sec

    406,022736256 seconds time elapsed

    CONVERT

    289,43 msec task-clock          #    0,001 CPUs utilized
    1 895       context-switches    #    0,007 M/sec
    33          cpu-migrations      #    0,114 K/sec
    21 505      page-faults         #    0,074 M/sec
    424 223 608 L1-dcache-loads     # 1465,721 M/sec

    545,113409858 seconds time elapsed

= It seems that using extract to-xapi is a bit faster than using extract | convert (difference of 139.09 sec = ~25% faster)

**3) parallel extract to-xapi command vs extract | parallel convert command:**

`$ echo "TO-XAPI PARALLEL"; sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat log.txt | parallel --pipe ralph -v CRITICAL extract -p gelf to-xapi -p https://fun-mooc.fr > /dev/null 2>&1"; echo "CONVERT PARALLEL ONLY CONVERT STEP" ; sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat log.txt | ralph -v CRITICAL extract -p gelf | parallel --pipe ralph -v CRITICAL convert -p https://fun-mooc.fr > /dev/null 2>&1"`

    TO-XAPI PARALLEL

    270,76 msec task-clock                #    0,001 CPUs utilized          
    1 052       context-switches          #    0,004 M/sec                  
    23          cpu-migrations            #    0,085 K/sec                  
    21 459      page-faults               #    0,079 M/sec                  
    436 399 752 L1-dcache-loads           # 1611,733 M/sec

    239,929169634 seconds time elapsed

    CONVERT PARALLEL

    260,28 msec task-clock                #    0,001 CPUs utilized          
    1 084       context-switches          #    0,004 M/sec                  
    20          cpu-migrations            #    0,077 K/sec                  
    21 412      page-faults               #    0,082 M/sec                  
    416 210 201 L1-dcache-loads           # 1599,095 M/sec

    223,509672211 seconds time elapsed


Note: Suprisingly "parallel extract | parallel convert" had worse performances (411 seconds)

= It seems that using extract | parallel convert is a bit faster than using extract to-xapi (difference of 16 sec = ~7% faster)

**4) Bonus: Comparing with first implementation of the converter (version using Marshmallow load/dump without GoTo/Link/KwargPath):**

`echo "TO-XAPI (FIRST COMMIT VERSION)"; sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat log.txt | ralph -v CRITICAL extract -p gelf to-xapi -p https://fun-mooc.fr > /dev/null 2>&1";`

    TO-XAPI (FIRST COMMIT VERSION)

    276,52 msec task-clock          #    0,001 CPUs utilized
    1 294       context-switches    #    0,005 M/sec
    23          cpu-migrations      #    0,083 K/sec
    21 573      page-faults         #    0,078 M/sec
    486 933 236 L1-dcache-loads     # 1760,904 M/sec

    398,295653421 seconds time elapsed

= It seems that the first implementation of the converter is a bit faster (difference of 7.7 sec = ~2% faster)

#### Conclusion:

The extract | parallel convert command performs better than the extract to-xapi command as soon as we execute it in parallel.

But I think this might point out another problem - the conversion is about 20 times slower than the extracting part. Should we do it better ?

jmaupetit:
Let's talk about this IRL!

SergioSim:
WIP on Benchmarks:

**1)  extract | convert command (new version of the converter using conversion_dict)**

`sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | ralph -v CRITICAL extract -p gelf | ralph -v CRITICAL convert -p https://fun-mooc.fr > /dev/null 2>&1"`

    294,17 msec task-clock          #    0,001 CPUs utilized
    1 340       context-switches    #    0,005 M/sec
    23          cpu-migrations      #    0,078 K/sec
    21 495      page-faults         #    0,073 M/sec
    417 865 199 L1-dcache-loads     # 1420,490 M/sec

    381,934024065 seconds time elapsed

= It seems the new version of the converter performs better (381 vs. 545 sec)(difference of 163.18 sec = ~30% faster)

**2) extract | convert command (new version of the converter using conversion_dict with flat_conversion_array )**

`sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | ralph -v CRITICAL extract -p gelf | ralph -v CRITICAL convert -p https://fun-mooc.fr > /dev/null 2>&1"`

    301,51 msec task-clock          #    0,001 CPUs utilized          
    1 526       context-switches    #    0,005 M/sec                  
    18          cpu-migrations      #    0,060 K/sec                  
    21 519      page-faults         #    0,071 M/sec                  
    428 266 972 L1-dcache-loads     # 1420,396 M/sec

     351,270590901 seconds time elapsed

= It seems that with the flat_conversion_array the converter performs a bit better (351 vs. 381 sec)(difference of 30.66 sec = ~8% faster)

**2) extract | convert command (new version of the converter using conversion_dict with flat_conversion_array and json.loads instead of pandas.from_json) + last comparaison with to-xapi command (I promise)**

`echo "TO-XAPI"; sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | ralph -v CRITICAL extract -p gelf to-xapi -p https://fun-mooc.fr > /dev/null 2>&1"; echo "CONVERT" ; sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | ralph -v CRITICAL extract -p gelf | ralph -v CRITICAL convert -p https://fun-mooc.fr > /dev/null 2>&1"`

    TO-XAPI

    266,43 msec task-clock          #    0,001 CPUs utilized          
    1 057       context-switches    #    0,004 M/sec                  
    20          cpu-migrations      #    0,075 K/sec                  
    21 447      page-faults         #    0,080 M/sec                  
    476 452 651 L1-dcache-loads     # 1788,300 M/sec

    214,656165536 seconds time elapsed

    CONVERT

    244,33 msec task-clock          #    0,001 CPUs utilized          
    1 059       context-switches    #    0,004 M/sec                  
    12          cpu-migrations      #    0,049 K/sec                  
    21 505      page-faults         #    0,088 M/sec                  
    427 298 813 L1-dcache-loads     # 1748,842 M/sec

    215,917903286 seconds time elapsed

= Convert and to-xapi finnaly line up! So it was the pandas.from_json part that slowed down the convert command. The differece with the previous run: (351 vs. 215 sec) (difference of 135.35 sec = ~39% faster)

### Running cProfile and visualising base_converter with runsnakerun
 
runsnakerun

= It's seems most of the time is spend on the marshmallow validation part

**Test run without schema validation**

`sudo chrt -f 99 perf stat -d docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | ralph -v CRITICAL extract -p gelf | ralph -v CRITICAL convert -p https://fun-mooc.fr > /dev/null 2>&1"`

    1 447,58 msec   task-clock          #    0,023 CPUs utilized          
    815             context-switches    #    0,563 K/sec                  
    16              cpu-migrations      #    0,011 K/sec                  
    25 462          page-faults         #    0,018 M/sec                  
    2 101 772 221   L1-dcache-loads     # 1451,921 M/sec

    62,368962790 seconds time elapsed

SergioSim:
**I have tried to reproduce the benchmark of February on the extract command but wasn't able to get the same results, may be I did something wrong ?**

`docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | mprof run ralph -v CRITICAL extract -p gelf -c 100000 > /dev/null 2>&1";`
`docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | mprof run ralph -v CRITICAL extract -p gelf -c 50000 > /dev/null 2>&1";`
`docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | mprof run ralph -v CRITICAL extract -p gelf -c 5000 > /dev/null 2>&1";`
`docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | mprof run ralph -v CRITICAL extract -p gelf -c 3000 > /dev/null 2>&1";`
`docker-compose run --rm -T --no-deps app bash -c "cat bench/log.txt | mprof run ralph -v CRITICAL extract -p gelf -c 500 > /dev/null 2>&1"`

`mprof plot -t "Ralph extract with different chunk sizes" mprofile_20201204085924.dat mprofile_20201204085945.dat mprofile_20201204090006.dat mprofile_20201204090027.dat mprofile_20201204090049.dat`

compare_chunk_sizes_1

That's really strange, the memory usage should decrease considerably with smaller chunksizes but for some reason it's not the case for me :disappointed: 

**Comparing other json parsers (they read the input line by line) with default pandas read_json.**

compare_parsers_1

**Zoom on the tree fastest:**
compare_parsers_2

=This benchmark suggest that using ujson / orjson packages could increase the performance of json parsing (reduce time by ~30%) (for extract and potentially convert command) and reduce memory consumption

jmaupetit:
Thanks for those benchmarks. My conclusion is that we can remove Pandas from project dependencies. :wink:

src/ralph/filters.py
```diff
@@ -1,22 +1,24 @@
...
+    """Remove anonymous event.
 
     Args:
-        events (DataFrame): events to filter
+        event (dict): event to filter
 
     Returns:
-        Filtered pandas DataFrame.
+        event (dict): when event is not anonymous
+        None: otherwise
```
SergioSim, line 16: As we removed the pandas dependency some changes had to be made for the anonymous filter, is this ok?

tests/test_parsers.py
```diff
@@ -3,33 +3,22 @@
...
 
     parser = GELFParser()
 
-    # Input log file is empty
-    empty_file_path = "/var/log/empty"
-    fs.create_file(empty_file_path)
+    empty_file = StringIO()
     with pytest.raises(StopIteration):
-        next(parser.parse(empty_file_path))
+        next(parser.parse(empty_file))
```
SergioSim, line 21: Removed the ability to parse files by specifying a path as the cli parse command only reads from standart input.
May be I could add an option to the parse command to specify a file instead of the standart input ? Then we could keep this tests unchanged?

src/ralph/converters/xapi/base.py
SergioSim: The current behaviour is: we keep the XAPI_EXTENSION_USER_ID and set it to an empty string when user_id is null

src/ralph/cli.py
```diff
@@ -92,23 +92,33 @@ def wrapper(command):
...
+)
+def convert(platform):
+    """Convert extracted events to xApi"""
+
+    logger.info("Converting events to xAPI")
+
+    converter_selector = XapiConverterSelector(platform)
+
+    for event in converter_selector.convert(sys.stdin):
         click.echo(event)
```
SergioSim, line 122: Apropos, I tried a version of the converter using `json.dump(event, sys.stout, default=default)` where the default function did the transformation for GetFromField instances instead - but it ended up being slower.
I thought, as json.dump is a C function wrapper it should be faster to let the json library do the transform & writing but unfortunately I was wrong.

src/ralph/filters.py
jmaupetit: yes

tests/test_parsers.py
jmaupetit: For now, we suppose to have a handle on a file.

src/ralph/schemas/edx/converters/xapi/server_event_to_xapi.py
```diff
@@ -0,0 +1,65 @@
...
+            "extensions": {
+                const.XAPI_EXTENSION_ACCEPT_LANGUAGE: GetFromField("accept_language"),
+                const.XAPI_EXTENSION_AGENT: GetFromField("agent"),
+                const.XAPI_EXTENSION_COURSE_ID: GetFromField("context>course_id"),
+                const.XAPI_EXTENSION_COURSE_USER_TAGS: GetFromField(
+                    "context>course_user_tags",
+                    lambda course_user_tags: course_user_tags
+                    if course_user_tags
+                    else {},
+                ),
```
SergioSim, line 52: One question, the [xAPI spec](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Data.md#requirements) states `Statements and other objects SHOULD NOT include properties with a value of an empty object.` 

What should we do if course_user_tags is not present (or equal to an empty object) ? It seems we can't set it to {} by default

src/ralph/filters.py
```diff
@@ -1,22 +1,24 @@
...
 
     """
 
-    if events.get("username", None) is None:
-        raise EventKeyError(
-            "Cannot filter anonymous filters without 'username' column."
-        )
-    return events.loc[lambda df: df["username"] != "", :]
+    if (username := event.get("username", None)) is None:
```
jmaupetit, line 20: Last week I made a small poll in the team, and we decided that this pattern is hard to maintain or understand by juniors. We prefer having something more explicit/classical:

```suggestion
    username = event.get("username", None)
    if username is None:
```

src/ralph/parsers.py
```diff
@@ -41,29 +36,24 @@ class GELFParser(BaseParser):
...
-
-        chunks = pd.read_json(input_file, lines=True, chunksize=chunksize)
-        for chunk in chunks:
-            for event in chunk["short_message"].values:
-                yield event
+        for event in input_file:
+            try:
+                yield json.loads(event)["short_message"]
+            except (json.JSONDecodeError, TypeError, KeyError) as err:
+                logger.info(
```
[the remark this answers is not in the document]
SergioSim: Good catch, indeed logging an error makes more sens, thanks. Updated the PR.

src/ralph/filters.py
SergioSim: Good to know, thanks, will then avoid this pattern for now. Updated the PR.

src/ralph/converters/xapi/base.py
jmaupetit: It should be in the `actor` section.

SergioSim: This make sens, but it seems the [xAPI spec](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Data.md#41-extensions) has foreseen extensions only as part of `Activity Definitions`,  `context` property, or `result` property. [The actor property ](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Data.md#242-actor) includes only optional `objectType`(string), optional `name` (string) and a [unique identifier ](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Data.md#2423-inverse-functional-identifier) (mailto IRI or String or URI or [account object](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Data.md#2424-account-object)) Where the account object only holds a `homePage` and `name` property.

jmaupetit: For Masha's project we decided to add edx's anonynous user key in the `actor.account.name` field.

jmaupetit: Anonymous users are referred as "student".

SergioSim: Ah, ok, so we put edx `context.user_id` in `actor.account.name` instead of edx `username` ?
And what about the rest of the extensions above ?

jmaupetit: This is subject to changes in the coming weeks. We'll discuss about this IRL soon.

src/ralph/schemas/edx/base.py
```diff
@@ -0,0 +1,170 @@
...
+
+        path = data["path"]
+        valid_path = (
+            f"/courses/"
+            f"{data['course_id']}"
+            f"/xblock/"
+            f"{data['module']['usage_key']}/"
+            f"handler/"
+        )
+        if path[: len(valid_path)] != valid_path:
```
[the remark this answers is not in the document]
SergioSim: Good point) `not path.startswith(valid_path)` is much better ! Thanks)

Here we check only the start of the {context>path} because we know that the context has a module field so the {context>path} should match something like:

`/courses/{context>course_id}/xblock/{context>module>usage_id}/handler/{hanlder_name}{?/suffix}`

But at this point we don't know what the handler_name is (Example: "xmodule_handler" / "peer_assess") and if there is a suffix (Example: "problem_check" / "hint_button")

src/ralph/schemas/edx/base.py
```diff
@@ -0,0 +1,170 @@
...
+            return ""
+        return value.exploded
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if not value:
+            return ""
+        try:
+            return IPv4Address(value)
+        except ValueError as error:
+            raise ValidationError("Invalid IPv4 Address") from error
```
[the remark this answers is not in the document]
SergioSim: Good Idea, thanks, updated

src/ralph/schemas/edx/converters/base_converter.py
```diff
@@ -0,0 +1,102 @@
+"""Converter Base Class"""
```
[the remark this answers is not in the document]
SergioSim: Good Idea. Although we have a lot of "base.py", this seems to be more consistent. Updated the PR.
```
 >find . -name base.py
./tests/fixtures/edx/base.py
./src/ralph/schemas/edx/converters/xapi/base.py
./src/ralph/schemas/edx/converters/base.py
./src/ralph/schemas/edx/base.py
./src/ralph/backends/storage/base.py
./src/ralph/backends/database/base.py
```

src/ralph/schemas/edx/converters/base_converter.py
```diff
@@ -0,0 +1,102 @@
...
+    Converters define a conversion dictionnary to convert from one schema to another
+    """
+
+    _schema = Schema()
+    conversion_dict = {}
+
+    def __init__(self):
+        """Initialize Base Converter"""
+
+        self.conversion_dict = copy.deepcopy(type(self).conversion_dict)
```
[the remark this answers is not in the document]
SergioSim: I'm surprised that this worked! Thanks!
This is the same as:
```
def __init__(self):
        """Initialize Base Converter"""
        self.conversion_dict = copy.deepcopy(self.conversion_dict)
```
But we can't do the same trick at the line 69 (in comment below) as we want the "Class variable" and self.conversion_dict becomes an "Instance variable" after the `__init__`.

Now this part has changed  - conversion_dict is now retrieved from a method (get_conversion_dictionary) and we don't have the static conversion_dict anymore.

src/ralph/schemas/edx/base.py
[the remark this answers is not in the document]
SergioSim: Re-updated this part - now only validating the IP address [here](https://github.com/openfun/ralph/pull/29/files#diff-b2d0d66fb47826c80e0ab78de7bb4c8da90072d3908a11e12215ff19a8a25a71R140)

src/ralph/schemas/edx/converters/xapi/constants.py
```diff
@@ -0,0 +1,30 @@
...
+XAPI_EXTENSION_AGENT = "https://www.edx.org/extension/agent"
+XAPI_EXTENSION_COURSE_ID = "https://www.edx.org/extension/course_id"
+XAPI_EXTENSION_COURSE_USER_TAGS = "https://www.edx.org/extension/course_user_tags"
+XAPI_EXTENSION_HOST = "https://www.edx.org/extension/host"
+XAPI_EXTENSION_IP = "https://www.edx.org/extension/ip"
+XAPI_EXTENSION_ORG_ID = "https://www.edx.org/extension/org_id"
+XAPI_EXTENSION_PATH = "https://www.edx.org/extension/path"
+XAPI_EXTENSION_REFERER = "https://www.edx.org/extension/referer"
+XAPI_EXTENSION_REQUEST = "https://www.edx.org/extension/request"
+XAPI_EXTENSION_USER_ID = "https://www.edx.org/extension/user_id"
```
SergioSim, line 23:
```suggestion
XAPI_EXTENSION_AGENT = "http://id.tincanapi.com/extension/browser-info"
XAPI_EXTENSION_COURSE_ID = "https://www.edx.org/extension/course_id"
XAPI_EXTENSION_COURSE_USER_TAGS = "https://www.edx.org/extension/course_user_tags"
XAPI_EXTENSION_HOST = "https://www.edx.org/extension/host"
XAPI_EXTENSION_IP = "http://id.tincanapi.com/extension/ip-address"
XAPI_EXTENSION_ORG_ID = "https://www.edx.org/extension/org_id"
XAPI_EXTENSION_PATH = "https://www.edx.org/extension/path"
```
WDYT ?
If we remove the referer because of sensitive information may be we should remove the client request arguments too (GET parameters / POST data)
There are extensions in the [tincanapi extentsions page](https://registry.tincanapi.com/#home/extensions) that match some of ours - should we use those instead? (ip / agent)

scripts/tests
```diff
@@ -0,0 +1 @@
+../tests/
```
SergioSim, line 1: I wanted to use the fixtures defined in `ralph/tests` from `ralph/scripts/generate_docs.py` but wasn't able to import it from there.
So I created a symbolic link. I suppose this is bad practice, is there a better way ?

SergioSim:
Closing this pull request as it will be split into smaller pieces.
The first part is #59

PR #31, opened 2020-11-20
Add Arnold tray

### LINKED ISSUE #18: Add Arnold tray
### Purpose

Arnold will soon support an experimental way to package deployable application similarly to Helms for k8s. It's a great opportunity to add a tested tray for Ralph.

### Proposal

- [ ] add tray's openshift objects (customizable cronjob + secret)
- [ ] update the CI to test tray's deployment

jmaupetit:
## Purpose

Now that Ralph is almost ready, we need to be able to deploy it to k8s with our beloved Arnold tool.

## Proposal

- [x] add `verify_certs` option to ESDatabase backend
- [x] add `docker-hub` make rule
- [x] add Arnold tray
- [x] test the tray in the CI
- [x] ~document tray development~ **update:** should be handled in Arnold's documentation

Fix #18

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,11 +18,11 @@ docker-login: &docker-login
     # Nota bene: you'll need to define the following secrets environment vars
     # in CircleCI interface:
     #
-    #   - DOCKER_USER
-    #   - DOCKER_PASS
+    #   - DOCKER_HUB_USER
+    #   - DOCKER_HUB_PASSWORD
     run:
       name: Login to DockerHub
-      command: echo "$DOCKER_PASS" | docker login -u "$DOCKER_USER" --password-stdin
+      command: echo "$DOCKER_HUB_PASSWORD" | docker login -u "$DOCKER_HUB_USER" --password-stdin
 
 version: 2
 jobs:
@@ -32,8 +32,8 @@ jobs:
     docker:
       - image: circleci/python:3.8-buster
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
     working_directory: ~/fun
     steps:
       - checkout
@@ -62,8 +62,8 @@ jobs:
     docker:
       - image: circleci/buildpack-deps:stretch-scm
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
     working_directory: ~/fun
     steps:
       - checkout
@@ -77,8 +77,8 @@ jobs:
     docker:
       - image: debian:stretch
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
     working_directory: ~/fun
     steps:
       - checkout
@@ -94,8 +94,8 @@ jobs:
     docker:
       - image: circleci/buildpack-deps:stretch
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
     working_directory: ~/fun
     steps:
       # Checkout repository sources
@@ -120,8 +120,8 @@ jobs:
     docker:
       - image: circleci/python:3.8-buster
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
     working_directory: ~/fun
     steps:
       - checkout
@@ -140,8 +140,8 @@ jobs:
     docker:
       - image: circleci/python:3.8-buster
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
     working_directory: ~/fun
     steps:
       - checkout
@@ -168,14 +168,14 @@ jobs:
     docker:
       - image: circleci/python:3.8-buster
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
         environment:
           RALPH_APP_DIR: ~/fun/.ralph
       - image: elasticsearch:7.10.1
         auth:
-          username: $DOCKER_USER
-          password: $DOCKER_PASS
+          username: $DOCKER_HUB_USER
+          password: $DOCKER_HUB_PASSWORD
         environment:
           discovery.type: single-node
     working_directory: ~/fun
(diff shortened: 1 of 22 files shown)


src/tray/templates/services/app/cronjob_pipeline.yml.j2
```diff
@@ -0,0 +1,50 @@
...
+      labels:
+        app: ralph
+        service: app
+        version: "{{ ralph_image_tag }}"
+    spec:
+      schedule: "{{ cronjob.schedule }}"
+      successfulJobsHistoryLimit: 2
+      failedJobsHistoryLimit: 1
+      concurrencyPolicy: Forbid
+      suspend: {{ suspend_cronjob | default(false) }}
```
jmaupetit, line 19: This does not work for non-blue-green apps. Should be `false` in this case.

.circleci/config.yml
```diff
@@ -192,13 +192,101 @@ jobs:
...
+            echo "export ARNOLD_DEFAULT_VAULT_PASSWORD=arnold" >> $BASH_ENV
+            echo 'export ANSIBLE_VAULT_PASSWORD="${ARNOLD_DEFAULT_VAULT_PASSWORD}"' >> $BASH_ENV
+            echo "export K8S_AUTH_VERIFY_SSL=no" >> $BASH_ENV
+            source $BASH_ENV
+
+      - run:
+          name: Install arnold CLI
+          command: |
+            wget https://raw.githubusercontent.com/openfun/arnold/master/bin/arnold -O ${HOME}/bin/arnold
+            chmod +x $HOME/bin/arnold
```
jmaupetit, line 246:
```suggestion
            chmod +x ${HOME}/bin/arnold
```

.circleci/config.yml
```diff
@@ -192,13 +192,101 @@ jobs:
...
+            arnold -d -c ralph -e ci create_project
+            docker push 172.30.1.1:5000/ci-ralph/ralph:${CIRCLE_SHA1}
+
+      - run:
+          name: Bootstrap ralph application
+          command: |
+            arnold -d -c ralph -e ci -a ralph init
+            arnold -d -c ralph -e ci -a ralph deploy
+            oc project ci-ralph
+            oc get pods -l app=ralph | grep Running
```
jmaupetit, line 281: Maybe we can also check that at least one cronjob ran?

jmaupetit: In this case we should increase the job frequency to at least 10 sec.

.circleci/config.yml
```diff
@@ -192,13 +192,101 @@ jobs:
...
+    machine:
+      # Prevent cache-related issues
+      docker_layer_caching: false
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - <<: [*generate-version-file, *docker-login]
+
+      - run:
+          name: Build production image (using cached layers)
```
lunika, line 206: cache seems deactivated here.

.circleci/config.yml
madmatah: If you want to check this, I recommend to look [here](https://www.craftypenguins.net/how-to-trigger-a-kubernetes-cronjob-manually/) and [here](https://www.mankier.com/1/oc-create-job). You can manually trigger a cronjob execution, instead of waiting for it to be scheduled.

.circleci/config.yml
```diff
@@ -192,13 +192,101 @@ jobs:
...
+            K8S_DOMAIN=$(hostname -I | awk '{print $1}')
+            export K8S_DOMAIN
+            oc cluster up --server-loglevel=5 --public-hostname=${K8S_DOMAIN}
+            oc login https://${K8S_DOMAIN}:8443 -u developer -p developer --insecure-skip-tls-verify=true
+            # Set environment variables for the CI
+            echo "export K8S_DOMAIN=${K8S_DOMAIN}" >> $BASH_ENV
+            echo "export ARNOLD_DEFAULT_VAULT_PASSWORD=arnold" >> $BASH_ENV
+            echo 'export ANSIBLE_VAULT_PASSWORD="${ARNOLD_DEFAULT_VAULT_PASSWORD}"' >> $BASH_ENV
+            echo "export K8S_AUTH_VERIFY_SSL=no" >> $BASH_ENV
+            source $BASH_ENV
```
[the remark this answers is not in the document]
jmaupetit: Yeah, we are on the same line: I thought a bit about this yesterday. That's the next step.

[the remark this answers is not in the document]
madmatah: Great :+1:

.circleci/config.yml
jmaupetit: I'll follow @madmatah instructions on this.

jmaupetit:
@lunika & @madmatah I've addressed all your comments. Ready for a final review?

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       args:
         DOCKER_USER: ${DOCKER_USER:-1000}
     user: ${DOCKER_USER:-1000}
-    image: ralph:development
+    image: fundocker/ralph:development
```
lunika, line 11: interesting, we never do that... why did you choose to change it ? It's just to ease the use in the CI ?

.circleci/config.yml
```diff
@@ -192,13 +192,107 @@ jobs:
...
+          command: docker build -t "ralph:${CIRCLE_SHA1}" --target production .
+
+      - run:
+          name: Check built images availability
+          command: docker images "ralph:${CIRCLE_SHA1}*"
+
+      - run:
+          name: Install the OC client
+          command: |
+            mkdir /tmp/openshift/
```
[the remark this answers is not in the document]
jmaupetit: It always does.

docker-compose.yml
jmaupetit: To ease locally build image publication via: `docker-compose push`

.circleci/config.yml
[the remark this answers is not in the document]
lunika: so you should use `-p`

.circleci/config.yml
```diff
@@ -192,13 +192,107 @@ jobs:
...
+      - run:
+          name: Bootstrap ralph application
+          command: |
+            arnold -d -c ralph -e ci -a ralph init
+            arnold -d -c ralph -e ci -a ralph deploy
+            oc project ci-ralph
+            oc get pods -l app=ralph | grep Running
+            test $(oc get cj -o name | wc -l) -eq 1
+            oc create job ralph-ci-test-job --from=cronjob/ralph-ci-test
+            test $(oc get jobs -o name | wc -l) -ge 1
```
madmatah, line 287: What is your intention here ?
Testing that the job has been completed successfully ?

jmaupetit: At least that it has been created and ran.

madmatah: > At least that it has been created 

Maybe I'm wrong, but I consider that from the moment the `cronjob` creation has been accepted, the underlying `job` can inevitably be created since its definition has been checked.

> and ran

I'm afraid that just after the job creation, the underlying pod will in most case appear as `Pending`.
And even if the job ran, it can be in `Error` status and the test will also pass. This line does not ensure that the job has been ran.

You can wait for the job to be terminated ( [see here](https://stackoverflow.com/a/55077916) ) but it might be overkill :man_shrugging: 


Regardless of these remarks, I have 2 suggestions to be sure to target the right cronjob / jobs : 
```suggestion
            test $(oc get jobs -o name ralph-ci-test-job | wc -l) -ge 1
```

And for the line above : 

```diff
-test $(oc get cj -o name | wc -l) -eq 1
+test $(oc get cj -o name ralph-ci-test | wc -l) -eq 1
```

jmaupetit: > Maybe I'm wrong, but I consider that from the moment the cronjob creation has been accepted, the underlying job can inevitably be created since its definition has been checked.

I will trust you on this.

> I'm afraid that just after the job creation, the underlying pod will in most case appear as Pending.
And even if the job ran, it can be in Error status and the test will also pass. This line does not ensure that the job has been ran.

True. What would you recommend then? Wait until finished even if it might be overkill?

madmatah: If it was me, I would remove the last 2 lines since we don't check any cronjob execution in any arnold app ;)

But if you want to test it, yes I think the way to go is to wait until the job condition is `complete` or `failed`.

jmaupetit: Ok, I've tried and it's not that easy, as the wait flag seems to wait forever. Let's be pragmatic on this. :wink:

jmaupetit:
@madmatah & @lunika ready for a (final) review?

madmatah (approved):
it !

PR #33, opened 2020-11-25
Add --no-deps argument to ralph

SergioSim:
### Purpose

Some ralph commands do not require additional docker dependencies

Example: extract command or push -b fs command

Also, when docker-compose loads additional dependencies it writes to the standard output, which inferences ralph commands reading from the standard output.

Example: stream | ralph extract ... | read output
```
Starting ralph_elasticsearch_1 ... done
extracted stream
```

### Proposal

- [X] add --no-deps argument to launch ralph without docker dependencies (./bin/ralph --no-deps)

### DIFF
--- a/bin/ralph
+++ b/bin/ralph
@@ -3,4 +3,10 @@
 declare DOCKER_USER
 DOCKER_USER="$(id -u):$(id -g)"
 
-DOCKER_USER=${DOCKER_USER} docker-compose run --rm -T app ralph ${@}
+NO_DEPS=""
+if [ "$1" == "--no-deps" ]; then
+    NO_DEPS="--no-deps"
+    shift
+fi
+
+DOCKER_USER=${DOCKER_USER} docker-compose run --rm -T ${NO_DEPS} app ralph ${@}


jmaupetit (commented):
Nice idea. Why don't you add an option to the `bin/ralph` script instead of creating a new one?

I recommend using `getopt` bash command to do so.

An alternative implementation would be to use an environment variable to toggle this setting.

SergioSim:
> Nice idea. Why don't you add an option to the `bin/ralph` script instead of creating a new one?
> 
> I recommend using `getopt` bash command to do so.
> 
> An alternative implementation would be to use an environment variable to toggle this setting.

Thanks for your advice! Unfortunately I haven't succeeded with the getopt command but found another way using "if" and "shift". Would it be ok if we use that ?

bin/ralph
```diff
@@ -3,4 +3,10 @@
...
 DOCKER_USER="$(id -u):$(id -g)"
 
-DOCKER_USER=${DOCKER_USER} docker-compose run --rm -T app ralph ${@}
+NO_DEPS=""
+if [ "$1" == "--no-deps" ]; then
+    NO_DEPS="--no-deps"
+    shift
+fi
+
+DOCKER_USER=${DOCKER_USER} docker-compose run --rm -T ${NO_DEPS} app ralph ${@}
```
jmaupetit, line 12: A carriage return is missing.

jmaupetit (commented):
Seems fair :+1:

bin/ralph
SergioSim: Good catch, thanks, added the new line)

jmaupetit (approved):

PR #44, opened 2021-01-13
💚(circle) fix docker login step

jmaupetit:
## Purpose

YAML anchors misuse leads to drop the docker login step, so we are not able to publish images to DockerHub.

## Proposal

- [x] use simple YAML anchors


sampaccoud (approved):
seems legit :man_shrugging:

lunika (approved):

PR #47, opened 2021-01-22
Swift storage backend

### LINKED ISSUE #42: Add OVH swift storage backend
## Purpose

OVH Swift object storage is widely used. It's crucial for us to support it.

## Proposal

- [ ] add a new swift storage backend

SergioSim:
### Purpose

OpenStack's Swift object storage is widely used. We want to support it.

### Proposal

Add a new swift storage backend

Fix #42


src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+        os_tenant_id,
+        os_tenant_name,
+        os_username,
+        os_password,
+        os_region_name,
+        os_storage_url,
+        os_user_domain_name=SWIFT_OS_USER_DOMAIN_NAME,
+        os_project_domain_name=SWIFT_OS_PROJECT_DOMAIN_NAME,
+        os_auth_url=SWIFT_OS_AUTH_URL,
+        os_identity_api_version=SWIFT_OS_IDENTITY_API_VERSION,
```
SergioSim, line 55: Here are the most common / necessary options for the swift client.
The swift client support a lot more [options](https://docs.openstack.org/python-swiftclient/latest/service-api.html#options). Do we add (copy/paste) all of them here ?

SergioSim:
### (Mini) Benchmark

Just to make sure we don't use too much memory.

Create  a 1.47G random text file and gzip it (to get about ~1.1G):

`< /dev/urandom tr -dc "\t\n [:alnum:]" | head -c1470000000 > 1GB.txt ; gzip -f 1GB.txt ;`

Push to swift:

`docker-compose run --rm -T --no-deps app bash -c "cat 1GB.txt.gz | mprof run ralph -v CRITICAL push -b swift 1GB.txt.gz";`

swift_push

Fetch from swift:

`docker-compose run --rm -T --no-deps app bash -c "mprof run ralph -v CRITICAL fetch -b swift 1GB.txt.gz > fetched_1GB.txt.gz";`

swift_fetch

P.S. I wondered why fetched_1GB.txt.gz differed with the original 1GB.txt.gz while both had the same size. Later found out that `mprof run` also wrote to sdtout and corrupted the output :man_facepalming:! Running the same test without `mprof run` solved this problem.

tests/backends/storage/test_swift.py
```diff
@@ -0,0 +1,286 @@
...
+    """
+
+    msg = f"Error: {exception.__class__}"
+
+    @throws_swift_error
+    def will_throw_exception():
+        """Throws an instance of exception"""
+        raise exception(msg)
+
+    with TestCase().assertLogs("ralph", level="ERROR") as cm:
```
jmaupetit, line 214: You can use the `caplog` pytest fixture instead. But this test will be removed if we remove tested decorator.

https://docs.pytest.org/en/stable/logging.html#caplog-fixture

jmaupetit (commented):
Thank you for this huge contribution :pray:

Dockerfile
```diff
@@ -28,10 +31,6 @@ FROM core as development
 # Copy all sources, not only runtime-required files
 COPY . /app/
 
-RUN apt-get update && \
-    apt-get install -y gcc && \
-    rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
SergioSim: Not sure, but the python-swiftclient also requires gcc (netifaces don't compile without it)

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+            for entry in filter(self.fetched_from_backend, self.history):
+                archives_to_skip.add(entry["id"])
+        with SwiftService(self.options) as swift:
+            for page in swift.list(self.container):
+                if not page["success"]:
+                    msg = "Failed to list container %s: %s"
+                    logger.error(msg, page["container"], page["error"])
+                    continue
+                for archive in page["listing"]:
+                    if new and archive["name"] in archives_to_skip:
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks, updated with:

`archives_to_skip = self.get_command_history(self.name, "fetch")`

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Remove click_log package dependency
+- Support for Swift storage backend
```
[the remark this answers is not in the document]
SergioSim: Good catch) Thanks)

Dockerfile
```diff
@@ -11,7 +11,10 @@ WORKDIR /build
 
 COPY . /build/
 
-RUN python setup.py install
+RUN apt-get update && \
+    apt-get install -y gcc python3-netifaces && \
```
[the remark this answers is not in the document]
SergioSim: Good question)

[python-swiftclient](https://opendev.org/openstack/python-swiftclient/src/tag/3.11.0/requirements.txt) depends on
[python-keystoneclient](https://opendev.org/openstack/python-keystoneclient/src/tag/4.2.0/requirements.txt) which depends on
[oslo.utils](https://opendev.org/openstack/oslo.utils/src/tag/3.33.0/requirements.txt) which depends on
[netifaces](https://github.com/al45tair/netifaces/blob/8b0d75750543453cdc8680bc52d4240eb4de2636/setup.py#L112) which depends on (quoting official documentation) "relevant developer tools".

Installing gcc and python3-netifaces enables us to compile the netifaces dependency.
Thanks for this question! It made me double-check and actually we don't need python3-netifaces, libc6-dev is enough.

src/ralph/backends/mixins.py
```diff
@@ -53,3 +54,9 @@ def append_to_history(self, event):
         """Append event to history"""
 
         self.write_history(self.history + [event])
+
+    @staticmethod
+    def get_current_iso_time():
+        """Return the current UTC time in ISO format"""
+
+        return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
```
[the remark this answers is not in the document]
SergioSim: This makes sense, unfortunatly this creates a circular import: `utils` depends on `ralph.backends.storage` which depends on `utils`.
Removing the content of ralph.backends.storage.__ init __.py solves this (current updated pull request). Do we keep it this way ?

src/ralph/backends/mixins.py
```diff
@@ -53,3 +54,9 @@ def append_to_history(self, event):
         """Append event to history"""
 
         self.write_history(self.history + [event])
+
+    @staticmethod
+    def get_current_iso_time():
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Simpler is better) updated

setup.cfg
```diff
@@ -26,11 +26,13 @@ install_requires =
     click==7.1.2
     click-option-group==0.5.2
     elasticsearch==7.10.1
+    python-keystoneclient==4.2.0
     ovh==0.5.0
     pandas==1.2.0
     pyyaml==5.3.1
     requests==2.25.1
     sentry_sdk==0.19.5
+    python-swiftclient==3.11.0
```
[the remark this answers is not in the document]
SergioSim: Good catch) yes indeed)

src/ralph/backends/storage/base.py
```diff
@@ -23,3 +23,7 @@ def read(self, name, chunk_size=4096):
     @abstractmethod
     def write(self, name, chunk_size=4096, overwrite=False):
         """Write content to the `name` target"""
+
+    def fetched_from_backend(self, entry):
```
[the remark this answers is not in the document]
SergioSim: Good point! I was hesitating about that because this utility would depend on both `HistoryMixin` (self.history) and `BaseStorage` (self.name).
For now I put it in the `HistoryMixin` and adding the backend name argument `get_command_history(self, backend_name, command)`

src/ralph/backends/storage/fs.py
```diff
@@ -61,10 +61,7 @@ def list(self, details=False, new=False):
         if new:
             archives = set(archives) - set(
                 entry.get("id")
-                for entry in filter(
-                    lambda e: e["backend"] == self.name and e["command"] == "fetch",
-                    self.history,
-                )
+                for entry in filter(self.fetched_from_backend, self.history)
```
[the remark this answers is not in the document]
SergioSim: Thanks, updated with:
`archives = set(archives) - self.get_command_history(self.name, "fetch")`

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+            return func(*args, **kwargs)
+        except (SwiftError, ClientException) as error:
+            logger.error(error.value)
+            raise error
+
+    return wrapper
+
+
+class SwiftStorage(HistoryMixin, BaseStorage):
+    """OVH's Swift storage backend"""
```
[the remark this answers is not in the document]
SergioSim: Good catch) thanks) updated)

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+            "os_username": os_username,
+            "os_password": os_password,
+            "os_region_name": os_region_name,
+            "os_storage_url": os_storage_url,
+        }
+        self._check_options()
+        self.container = os_storage_url.split("/")[-1]
+        self.options["os_storage_url"] = os_storage_url.replace(
+            f"/{self.container}", ""
+        )
```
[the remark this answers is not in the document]
SergioSim: Good Idea)
1. Updated with instance properties
2. Updated
3. Now when initializing we make a stat request to check the connection

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+    """Decorator catches SwiftError or ClientException, then logs and re-raises it"""
+
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except (SwiftError, ClientException) as error:
+            logger.error(error.value)
+            raise error
+
+    return wrapper
```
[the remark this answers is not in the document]
SergioSim: Good point) Thanks) added the `swiftclient` logger in  `ralph.defaults` (but it behaves only partly when we use the verbosity command line option - is this a problem ?)

src/ralph/backends/storage/swift.py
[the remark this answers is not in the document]
SergioSim: Good question) I don't know) Both works for me) What approach do you prefer?

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+
+        Why chunks of (max) 2 ** 16 ?
+            Because SwiftService opens a file to stream the object into:
+            See swiftclient.service.py:2082 open(filename, 'rb', DISK_BUFFER)
+            Where filename = "/dev/stdout" and DISK_BUFFER = 2 ** 16
+        """
+
+        logger.debug("Getting archive: %s", name)
+
+        options = {"options": {"out_file": "/dev/stdout"}}
```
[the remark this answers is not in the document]
SergioSim: Good catch) Thanks) Indeed this will not work on other platforms) updated with the "-" option)

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+        with SwiftService(self.options) as swift:
+            download = next(swift.download(self.container, [name], **options))
```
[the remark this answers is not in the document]
SergioSim: This is true, Thanks) Updated

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+            raise FileExistsError(msg % name)
+
+        logger.debug("Creating archive: %s", name)
+
+        swift_object = SwiftUploadObject(sys.stdin.buffer, object_name=name)
+        with SwiftService(self.options) as swift:
+            for upload in swift.upload(self.container, [swift_object]):
+                if not upload["success"]:
+                    logger.error(upload["error"])
+                    raise BackendParameterException(upload["error"])
```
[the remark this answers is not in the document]
SergioSim: Thanks, yes indeed) updated this (created a BackendException)

tests/backends/storage/test_swift.py
SergioSim: Good point) Wasn't aware of caplog, thanks)

src/ralph/backends/storage/swift.py
```diff
@@ -0,0 +1,170 @@
...
+            "os_project_domain_name": os_project_domain_name,
+            "os_tenant_id": os_tenant_id,
+            "os_tenant_name": os_tenant_name,
+            "os_username": os_username,
+            "os_password": os_password,
+            "os_region_name": os_region_name,
+            "os_storage_url": os_storage_url,
+        }
+        self._check_options()
+        self.container = os_storage_url.split("/")[-1]
```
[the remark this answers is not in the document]
SergioSim: Good point) Updated using the `urllib.parse.urlparse`. But don't know how/why to make this a cached property as it depends on the initial `os_storage_url` value which then is modified afterwards.

src/ralph/backends/storage/swift.py
```diff
@@ -56,50 +46,54 @@ def __init__(
...
-            # pylint: disable=protected-access
-            if option is not None and not isinstance(option, inspect._empty):
-                continue
-            msg = "SwiftStorage backend instance requires the %s option to be set"
-            logger.error(msg, option_name)
-            raise BackendParameterException(msg % option_name)
+        with SwiftService(self.options) as swift:
+            stats = swift.stat()
+            if not stats["success"]:
+                msg = "SwiftStorage backend unable to connect: %s"
```
jmaupetit, line 64: The class/method is in the logger context (depending on the log format, you can choose to display it or not), we don't need to repeat it.

```suggestion
                msg = "Unable to connect to the requested container: %s"
```

src/ralph/backends/storage/swift.py
```diff
@@ -56,50 +46,54 @@ def __init__(
...
+        self.os_tenant_name = os_tenant_name
+        self.os_username = os_username
+        self.os_password = os_password
+        self.os_region_name = os_region_name
+        self.os_user_domain_name = os_user_domain_name
+        self.os_project_domain_name = os_project_domain_name
+        self.os_auth_url = os_auth_url
+        self.os_identity_api_version = os_identity_api_version
+        self.container = urlparse(os_storage_url).path.rpartition("/")[-1]
+        self.os_storage_url = os_storage_url.replace(f"/{self.container}", "")
```
jmaupetit, line 59:
```suggestion
        self.os_storage_url = os_storage_url.removesuffix(f"/{self.container}")
```

tests/fixtures/backends.py
```diff
@@ -47,12 +47,17 @@ def es():
...
-    """Returns an instance of SwiftStorage"""
-    return SwiftStorage(
-        os_tenant_id="os_tenant_id",
-        os_tenant_name="os_tenant_name",
-        os_username="os_username",
-        os_password="os_password",
-        os_region_name="os_region_name",
-        os_storage_url="os_storage_url/ralph_logs_container",
-    )
+    """Returns an SwiftStorage generator"""
```
jmaupetit, line 50:
```suggestion
    """Returns a SwiftStorage generator"""
```

tests/fixtures/backends.py
```diff
@@ -47,12 +47,17 @@ def es():
...
+
+    def swift_storage_generator():
+        """Returns an instance of SwiftStorage"""
+        return SwiftStorage(
```
jmaupetit, line 54: If it's a generator shouldn't we `yield` instead?

tests/backends/storage/test_swift.py
```diff
@@ -0,0 +1,291 @@
...
+        return {"success": True}
+
+    monkeypatch.setattr(SwiftService, "list", mock_list_with_pages)
+    monkeypatch.setattr(SwiftService, "stat", mock_successful_stat)
+    fs.create_file(HISTORY_FILE, contents=json.dumps([]))
+    caplog.set_level(logging.ERROR)
+    swift = swift()
+    msg = "Failed to list container ralph_logs_container: Container not found"
+    with pytest.raises(BackendException, match=msg):
+        assert list(swift.list()) == []
```
jmaupetit, line 108: You don't need to assert something as once the exception has been raised you're out of the context.

```suggestion
        swift.list()
```

tests/backends/storage/test_swift.py
```diff
@@ -0,0 +1,291 @@
+"""Tests for Ralph swift storage backend"""
+
+import datetime
+import json
+import logging
+
+import pytest
+
+from ralph.backends.storage import swift as swift_module
+from ralph.backends.storage.swift import SwiftService
```
jmaupetit, line 10:
```suggestion
from swiftclient.service import SwiftService
```

tests/backends/storage/test_swift.py
```diff
@@ -0,0 +1,291 @@
...
+    def mock_successful_download(*args, **kwargs):  # pylint:disable=unused-argument
+        yield {"contents": [b"some", b"contents"]}
+
+    def mock_successful_stat(*args, **kwargs):  # pylint:disable=unused-argument
+        return {"success": True}
+
+    freezed_now = datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+    monkeypatch.setattr(SwiftService, "download", mock_successful_download)
+    monkeypatch.setattr(SwiftService, "stat", mock_successful_stat)
+    monkeypatch.setattr(swift_module, "now", lambda: freezed_now)
```
jmaupetit, line 134: Why not patching `ralph.utils` instead?

```suggestion
    monkeypatch.setattr(ralph.utils, "now", lambda: freezed_now)
```

tests/backends/storage/test_swift.py
```diff
@@ -0,0 +1,291 @@
+"""Tests for Ralph swift storage backend"""
+
+import datetime
+import json
+import logging
+
+import pytest
+
+from ralph.backends.storage import swift as swift_module
```
jmaupetit, line 9: I would remove this as it's only used to patch `ralph.utils.now` and import `ralph.utils` instead.

Dockerfile
```diff
@@ -11,7 +11,10 @@ WORKDIR /build
 
 COPY . /build/
 
-RUN python setup.py install
+RUN apt-get update && \
+    apt-get install -y gcc libc6-dev && \
+    rm -rf /var/lib/apt/lists/* && \
+    python setup.py install
```
jmaupetit, line 17: I would write apt and python setup steps in separated layers:

```suggestion
    rm -rf /var/lib/apt/lists/*
    
RUN python setup.py install
```

jmaupetit (commented):
Almost ready to be merged! Yay!

src/ralph/backends/storage/swift.py
SergioSim: Thanks for the suggestion) was not aware of this method)

src/ralph/backends/storage/swift.py
SergioSim: Indeed, Thanks)

tests/fixtures/backends.py
SergioSim: Good catch) thanks)

Dockerfile
SergioSim: Makes sense, thanks)

tests/backends/storage/test_swift.py
SergioSim: This makes sense, unfortunately in the `swift.py` I use `from ralph.utils import now` and it seems that this gives the now function a different name. https://stackoverflow.com/questions/31306080/pytest-monkeypatch-isnt-working-on-imported-function

Do you prefer that we change the import to `import ralph.utils as utils` in swift.py or do we change the test monkeypatch ?:
from this:
`monkeypatch.setattr(swift_module, "now", lambda: freezed_now)`
to this:
`monkeypatch.setattr("ralph.backends.storage.swift.now", lambda: freezed_now)`

tests/fixtures/backends.py
SergioSim: Good point, thanks) Indeed, it's not consistent) changed the name to get_swift_storage

tests/backends/storage/test_swift.py
SergioSim: Good catch) thanks)

tests/backends/storage/test_swift.py
SergioSim: Thanks) updated)

src/ralph/backends/mixins.py
```diff
@@ -55,8 +54,12 @@ def append_to_history(self, event):
...
+    def get_command_history(self, backend_name, command):
+        """Returns a set of entry ids from the history for a command and backend_name"""
 
-        return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+        def from_backend(entry):
+            """Return if the entry was fetched by the backend"""
+
+            return entry["backend"] == backend_name and entry["command"] == command
+
+        return set(entry["id"] for entry in filter(from_backend, self.history))
```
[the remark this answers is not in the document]
SergioSim: Ok, updated) thanks)

src/ralph/backends/mixins.py
```diff
@@ -55,8 +54,12 @@ def append_to_history(self, event):
...
-    def get_current_iso_time():
-        """Return the current UTC time in ISO format"""
+    def get_command_history(self, backend_name, command):
+        """Returns a set of entry ids from the history for a command and backend_name"""
 
-        return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+        def from_backend(entry):
+            """Return if the entry was fetched by the backend"""
+
+            return entry["backend"] == backend_name and entry["command"] == command
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks) updated)

src/ralph/backends/mixins.py
```diff
@@ -57,9 +57,7 @@ def append_to_history(self, event):
...
-        def from_backend(entry):
-            """Return if the entry was fetched by the backend"""
-
-            return entry["backend"] == backend_name and entry["command"] == command
-
-        return set(entry["id"] for entry in filter(from_backend, self.history))
+        from_backend = (
+            lambda item: item["backend"] == backend_name and item["command"] == command
+        )
+        return [entry["id"] for entry in filter(from_backend, self.history)]
```
jmaupetit, line 63:
```suggestion
        return [
            entry["id"] for entry in filter(
                lambda e: e["backend"] == backend_name and e["command"] == command, 
                self.history
            )
        ]
```

tests/backends/storage/test_swift.py
```diff
@@ -105,11 +104,11 @@ def mock_successful_stat(*args, **kwargs):  # pylint:disable=unused-argument
     swift = swift()
     msg = "Failed to list container ralph_logs_container: Container not found"
     with pytest.raises(BackendException, match=msg):
-        assert list(swift.list()) == []
+        list(swift.list())
```
jmaupetit, line 107: There is no need to cast this to a list. We only need to call the `list()` method.

README.md
```diff
@@ -46,6 +46,33 @@ Ralph is distributed along with its tray (a deployable package for Kubernetes
...
+RALPH_SWIFT_OS_REGION_NAME=RegionOne
+RALPH_SWIFT_OS_TENANT_ID=cd238e84310a46e58af7f1d515887d88
+RALPH_SWIFT_OS_TENANT_NAME=RegionOne
+RALPH_SWIFT_OS_STORAGE_URL=http://swift:8080/v1/KEY_cd238e84310a46e58af7f1d515887d88/test_container
+```
+
+2) Push a first object (forcefully) to create the test container
+
+```
+echo "some content" | ./bin/ralph push -b swift -f some_archive_name
```
jmaupetit, line 71: I usually add the shell prompt for shell commands in the docs:

```
$ echo "some content" | ./bin/ralph push -b swift -f some_archive_name
```

tests/backends/storage/test_swift.py
jmaupetit: I prefer the later (that you've implemented).

src/ralph/backends/mixins.py
SergioSim: Thanks)

README.md
SergioSim: That's good practice) Thanks) Will update this)

tests/backends/storage/test_swift.py
jmaupetit: As the `list()` method returns a generator, we agreed to use `next()` instead.

jmaupetit (approved):

PR #51, opened 2021-02-02
✨(cli) allow to force push for database backends

jmaupetit:
## Purpose

When bulk importing documents in ES, if a document is badly formatted (e.g. with mistyped fields), the bulk importation stops after the first chunk importation failure.

## Proposal

We now allow to force the importation regardless of indexation errors.

### DIFF
--- a/src/ralph/backends/database/es.py
+++ b/src/ralph/backends/database/es.py
@@ -47,7 +47,7 @@ class ESDatabase(BaseDatabase):
                 "_source": item,
             }
 
-    def put(self, chunk_size=500):
+    def put(self, chunk_size=500, ignore_errors=False):
         """Write documents streamed from the standard input to the instance index"""
 
         logger.debug(
@@ -59,6 +59,7 @@ class ESDatabase(BaseDatabase):
             client=self.client,
             actions=self.to_documents(sys.stdin, lambda d: d.get("id", None)),
             chunk_size=chunk_size,
+            raise_on_error=(not ignore_errors),
         ):
             documents += success
             logger.debug(
--- a/tests/backends/database/test_es.py
+++ b/tests/backends/database/test_es.py
@@ -1,13 +1,15 @@
 """Tests for Ralph es database backend"""
 
 import json
+import random
 import sys
 from collections.abc import Iterable
 from io import BytesIO, StringIO
 from pathlib import Path
 
+import pytest
 from elasticsearch import Elasticsearch
-from elasticsearch.helpers import bulk
+from elasticsearch.helpers import BulkIndexError, bulk
 
 from ralph.backends.database.es import ESDatabase
 from ralph.defaults import APP_DIR, HISTORY_FILE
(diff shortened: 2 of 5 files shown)


src/ralph/backends/database/es.py
```diff
@@ -47,7 +47,7 @@ def to_documents(self, stream, get_id):
                 "_source": item,
             }
 
-    def put(self, chunk_size=500):
+    def put(self, chunk_size=500, force=False):
```
lunika, line 50: `silent_errors` or `ignore_errors` should be more relevant IMO. I understand `force` like a `git push -f` and it's not the same meaning for me.

tests/backends/database/test_es.py
```diff
@@ -126,3 +128,62 @@ def test_write_method(es, fs, monkeypatch):
...
+    )
+
+    assert len(es.search(index=ES_TEST_INDEX)["hits"]["hits"]) == 0
+
+    database = ESDatabase(
+        hosts=ES_TEST_HOSTS,
+        index=ES_TEST_INDEX,
+    )
+
+    # By default, we sould raise an error and stop the importation
```
SergioSim, line 153:
```suggestion
    # By default, we should raise an error and stop the importation
```

SergioSim (commented):
This looks great!

tests/backends/database/test_es.py
```diff
@@ -126,3 +128,62 @@ def test_write_method(es, fs, monkeypatch):
...
+    """Test ES put method with badly formatted data when the force mode is active"""
+    # pylint: disable=invalid-name,unused-argument
+
+    records = [{"id": idx, "count": random.randint(0, 100)} for idx in range(10)]
+    # Patch a record with a non-expected type for the count field (should be
+    # assigned as long)
+    records[2].update({"count": "wrong"})
+
+    monkeypatch.setattr(
+        "sys.stdin", StringIO("\n".join([json.dumps(record) for record in records]))
```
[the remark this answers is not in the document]
jmaupetit: As JSONs are generated using the `json.dumps` method, I think we can consider generated data as valid no?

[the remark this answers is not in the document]
SergioSim: Good point, yes indeed, if we always read what `json.dumps` outputs then it's fine

src/ralph/backends/database/es.py
jmaupetit: When forcing a `git push`, it means push this regardless of potential errors occurring without the `--force` flag. So it's basically the same to me.

lunika: No with `git push --force` you bypass a security, you force to overwrite a remote ref that is not an ancestor of the local ref. It has nothing to do with ignoring errors.

quitterie-lcs (approved):
Modifications have been reviewed, ready for merging!

src/ralph/backends/database/es.py
jmaupetit: After IRL conversation we decided to introduce a new `ignore-errors` option to the `push` command.

jmaupetit:
@lunika I've addressed your comment. Feel free to give this PR a new try. Thanks! :pray:

lunika (approved):

PR #52, opened 2021-02-03
✨(backends:es) add support for supported client options

jmaupetit:
## Purpose

The elasticsearch client supports various options to support a lot of connection scenarii with an elasticsearch cluster. Unfortunately current implementation of ralph is restricted to the `verify_certs` option only.

## Proposal

We now also support passing those options from the CLI using coma-separated key=value pairs:

```
$ ralph push --backend es --es-client-options ca_certs=/path/to/ca/certs,client_cert=/path/to/client/cert.pem
```

### DIFF
--- a/src/ralph/cli.py
+++ b/src/ralph/cli.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import re
 import sys
 from inspect import signature
 
@@ -35,6 +36,40 @@ DATABASE_BACKENDS = (lambda: [backend.value for backend in DatabaseBackends])()
 PARSERS = (lambda: [parser.value for parser in Parsers])()
 STORAGE_BACKENDS = (lambda: [backend.value for backend in StorageBackends])()
 BACKENDS = (lambda: DATABASE_BACKENDS + STORAGE_BACKENDS)()
+
+
+class CommaSeparatedKeyValueParamType(click.ParamType):
+    """Comma separated key=value parameter type"""
+
+    name = "key=value,key=value"
+
+    def convert(self, value, param, ctx):
+        """Split value by comma and equal sign to return a dict build with key/value pairs"""
+
+        # Parse options string
+        try:
+            options = dict(option.split("=") for option in value.split(","))
+        except ValueError:
+            self.fail(
+                "You should provide key=value pairs separated by commas, e.g. foo=bar,bar=2",
+                param,
+                ctx,
+            )
+
+        # Cast simple types
+        for key, value_ in options.items():
+            if value_ == "":
+                options.update({key: None})
+            elif re.match("^true$", value_, re.IGNORECASE):
+                options.update({key: True})
+            elif re.match("^false$", value_, re.IGNORECASE):
+                options.update({key: False})
+            elif re.match(r"^[0-9]+\.[0-9]+$", value_):
+                options.update({key: float(value_)})
+            elif re.match("^[0-9]+$", value_):
+                options.update({key: int(value_)})
+
+        return options
 
 
 @click.group(name="ralph")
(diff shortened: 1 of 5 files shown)


src/ralph/cli.py
```diff
@@ -37,6 +37,24 @@
...
+class ComaSeparatedKeyValueParamType(click.ParamType):
+    """Coma separated key=value parameter type"""
+
+    name = "key=value,key=value"
+
+    def convert(self, value, param, ctx):
+        """Split value by coma and equal sign to return a dict build with key/value pairs"""
+
+        try:
+            return dict(option.split("=") for option in value.split(","))
```
jmaupetit, line 49: I think I'd better try to infer simple types to the value here (at least `bool`). WDYT?

madmatah (commented):
Nice! I just hope that no Elasticsearch option requires a comma separated list of values :smile:

SergioSim:
I like the approach with comma separated key values!

Although with `click` it's possible to [pass unspecified options](https://stackoverflow.com/a/36522299) unfortunately then the unspecified options are always at the end of the command and there is no hint for unspecified options with `--help`

src/ralph/cli.py
```diff
@@ -37,6 +37,24 @@
 BACKENDS = (lambda: DATABASE_BACKENDS + STORAGE_BACKENDS)()
 
 
+class ComaSeparatedKeyValueParamType(click.ParamType):
+    """Coma separated key=value parameter type"""
```
SergioSim, line 41:
```suggestion
    """Comma separated key=value parameter type"""
```

src/ralph/cli.py
quitterie-lcs: Good idea with bools but what would be the meaning of using `int` or other simple types in value?

quitterie-lcs (commented):
The concept of using key-value pair is super relevant and keeps the tool simple to use. Really like it!

src/ralph/cli.py
jmaupetit: Good catch! I'll fix this everywhere I made the typo :sweat_smile:

src/ralph/cli.py
jmaupetit: When the `elasticsearch` client will be instantiated, it expects to receive python types for some arguments, not always strings. Some of them are flags (booleans) but some of them may be integers or floats (such as the `timeout` parameter).

quitterie-lcs: Ok! Got it ;)

jmaupetit: And it's done.

src/ralph/cli.py
```diff
@@ -45,15 +46,31 @@ class CommaSeparatedKeyValueParamType(click.ParamType):
...
+        # Cast simple types
+        for key, value_ in options.items():
+            if value_ == "":
+                options.update({key: None})
+            elif re.match("^(T|t)rue$", value_):
+                options.update({key: True})
+            elif re.match("^(F|f)alse$", value_):
+                options.update({key: False})
+            elif re.match(r"^[0-9]+\.[0-9]+$", value_):
+                options.update({key: float(value_)})
```
madmatah, line 68: Are you sure you want to deal with floats ?
Floating point arithmetic and its precision can lead to errors, like in any programming language.
So, to be sure, I would personally avoid dealing with this type and keep it as a string.

madmatah (approved):
Just a few non-blocking suggestions ;)

src/ralph/cli.py
jmaupetit: I hesitated to do so, but in Ralph's particular case, I don't think it's an issue with expected floating point values (1 or 2 digits).

src/ralph/cli.py
```diff
@@ -45,15 +46,31 @@ class CommaSeparatedKeyValueParamType(click.ParamType):
...
             )
 
+        # Cast simple types
+        for key, value_ in options.items():
+            if value_ == "":
+                options.update({key: None})
+            elif re.match("^(T|t)rue$", value_):
+                options.update({key: True})
+            elif re.match("^(F|f)alse$", value_):
+                options.update({key: False})
```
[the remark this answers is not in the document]
jmaupetit: Good idea.

PR #54, opened 2021-02-04
📝(env) fix documentation of environment variables

quitterie-lcs:
## Purpose

The environment variables are defined in `.env.dist` file and are commented. Users have to uncomment the definition line for running the application. 

## Proposal

The indication has been documented on the head of the file so that user are aware of doing it before running ralph.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -1,6 +1,9 @@
 # Configuration
 RALPH_APP_DIR=/app/.ralph
 
+# Uncomment lines (by removing # characters at the beginning of target lines)
+# to define environment variables associated to the backend(s) you need.
+
 # LDP storage backend
 #
 # You need to generate an API token for your OVH's account and fill the service


.env.dist
```diff
@@ -1,6 +1,9 @@
 # Configuration
 RALPH_APP_DIR=/app/.ralph
 
+# Uncomment lines (by removing # at the head of the line)
```
jmaupetit, line 4:
```suggestion
# Uncomment lines (by removing # characters at the beginning of target lines)
```

jmaupetit (commented):
Nice one. A simple suggestion and we are good to go!

quitterie-lcs:
@jmaupetit I have updated with your suggestion, feel free to approve it ;)

jmaupetit (approved):

PR #55, opened 2021-02-05
🐛(tray) fix pods anti-affinity and missing secrets

jmaupetit:
## Purpose

Swift credentials loaded as environment variables are missing from the Tray's main secret.

## Proposal

- [x] add `RALPH_SWIFT_OS_*` variables in Raph app secret

During test deployments, I also fixed the anti-affinity selector of Ralph's DC as it's a non blue-green application.


src/tray/templates/services/app/secret.yml.j2
```diff
@@ -7,6 +7,17 @@ metadata:
   name: "{{ ralph_secret_name }}"
   namespace: "{{ project_name }}"
 data:
+  # Swift
+  RALPH_SWIFT_OS_AUTH_URL: "{{ RALPH_VAULT.RALPH_SWIFT_OS_AUTH_URL | b64encode }}"
```
lunika, line 11: All variable are mandatory ?

jmaupetit: To connect to swift yes. Maybe we should introduce more flexibility in backends configuration and allow to make them optional. Is this what you have in mind?

lunika: If they are mandartory, they should not be commented in the main file. WDYT ?

jmaupetit: They are mandatory only if you use this backend.

jmaupetit: Sorry you mean mandatory from a jinja/ansible perspective, not Ralph! You are perfectly right, this is not consistent.

jmaupetit: This is now fixed in a way simpler way. Thanks @lunika :muscle:

lunika (approved):

PR #59, opened 2021-02-10
✨(models) add edx server event model

SergioSim:
## Purpose

The edx server events are the most simple and common events in the edx tracking logs. They are logged at almost all requests made to the edx lms. We want to validate them.

## Proposal

- [x] ServerEventModel
- [x] ServerEventFactory
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- EdX server event pydantic model and factory
+
 ### Fixed
 
 - Tray: add missing Swift variables in the secret
--- a/src/ralph/models/edx/base.py
+++ b/src/ralph/models/edx/base.py
@@ -1,0 +1,143 @@
+"""Base event model definitions"""
+
+import re
+from datetime import datetime
+from ipaddress import IPv4Address
+from pathlib import Path
+from typing import Literal, Optional, Union
+
+from pydantic import AnyUrl, BaseModel, constr, root_validator
+
+
+class BaseModelWithConfig(BaseModel):
+    """Base model defining configuration shared among all models"""
+
+    class Config:  # pylint: disable=missing-class-docstring
+        extra = "forbid"
+
+
+class BaseContextModel(BaseModelWithConfig):
+    """Represents the base context model inherited by all event contexts.
+
+    Attributes:
+        course_user_tags (dict of str: str): Content from `user_api_usercoursetag` table.
+            Retrieved with:
+                `dict(
+                    UserCourseTag.objects.filter(
+                        user=request.user.pk, course_id=course_key
+                    ).values_list('key', 'value')
+                )`
+            Note:
+                Is only present when a course page is requested.
+                Is an empty dictionary when the user is not logged in or not found in the
+                `user_api_usercoursetag` table.
+        user_id (int or str or None): Consists of the ID of the authenticated user.
+            Retrieved with:
+                `request.user.pk` querying the `auth_user` table.
+            Note:
+                Is an integer when the user is found in the `auth_user` table.
+                Is an empty string when an exception is raised while retrieving the ID.
+                Is `None` when the user is not logged in.
+        org_id (str): Consists of the organization name that lists the course.
+            Retrieved with:
+                `course_id.org` where `course_id` is an `opaque_keys.edx.locator.CourseLocator`
+                which is created using the URL of the requested page.
+            Note:
+                Is an empty string when the requested page is not a course page.
+        course_id (str): Consists of the unique identifier for the visited course page.
+            Retrieved with:
+                `course_id.to_deprecated_string()` where `course_id` is an
+                `opaque_keys.edx.locator.CourseLocator` which is created using the URL
+                of the requested page.
+            Note:
+                Is an empty string when the requested page is not a course page.
+        path (Path): Consist of the relative URL (without the hostname) of the requested page.
+            Retrieved with:
+                `request.META['PATH_INFO']`
+    """
+
+    course_user_tags: Optional[dict[str, str]]
+    user_id: Union[int, Literal[""], None]
+    org_id: str
+    course_id: str
+    path: Path
+
+    @root_validator
+    def validate_course_id(
+        cls, values
+    ):  # pylint: disable=no-self-argument, no-self-use
+        """The course_id must match for example `course-v1:org+course+any` or an empty string."""
+
+        org_id = values.get("org_id")
+        course_id = values.get("course_id")
+        regex = f"course-v1:{org_id}\\+.+\\+.+"
+        if (org_id or course_id) and not re.match(regex, course_id):
+            raise ValueError(f"course_id must match regex `{regex}`")
+        return values
+
+
+class BaseEventModel(BaseModelWithConfig):
+    """Represents the base event model all events inherit from.
+
+    WARNING: it does not define the event and event_type fields.
+
+    Attributes:
+        username (str): Consists of the unique username identifying the logged in user.
+            Retrieved with:
+                `request.user.username` querying the `auth_user` table.
+            Note:
+                Is an empty string when the user is not logged in.
+                If an exception is raised when retrieving the username from the table then
+                the value is `anonymous`.
+                Usernames are made of 2-30 ASCII letters / numbers / underscores (_) / hyphens (-)
+        ip (IPv4Address or str): Consists of the public IPv4 address of the user.
+            Retrieved with:
+                `get_ip(request)` cf. https://github.com/un33k/django-ipware/tree/1.1.0
+            Note:
+                Can be an empty string if the IP address is not found.
+        agent (str): Consists of the `User-Agent` HTTP request header.
+            Retrieved with:
+                `request.META[HTTP_USER_AGENT]`
+            Note:
+                Can be an empty string if the header is not present in the request.
+                Contains information about:
+                    Browser name and version
+                    Operating system name and version
+                    Default language
+        host (str): Consists of the hostname of the server.
+            Retrieved with:
+                `request.META[SERVER_NAME]`
+        referer (Path): Consists of the `Referer` HTTP request header.
+            Retrieved with:
+                `request.META[HTTP_REFERER]`
+            Note:
+                Can be an empty string if the header is not present in the request.
+                Contains the referring URL (previous URL visited by the user).
+        accept_language (str): Consists of the `Accept-Language` HTTP request header.
+            Retrieved with:
+                `request.META[HTTP_ACCEPT_LANGUAGE]`
+            Note:
+                Can be an empty string if the header is not present in the request.
+                Contains the default language settings of the user.
+        event_source (str): Consists of the value `server`.
+            Note:
+                Specifies the source of the interaction that triggered the event.
+        context (BaseContextModel): see BaseContestModel.
+        time (datetime): Consists of the UTC time in ISO format at which the event was emitted.
+            Retrieved with:
+                `datetime.datetime.utcnow()`
+        page (None): Consists of the value `None`
+            Note:
+                In JSON the value is `null` instead of `None`.
+    """
+
+    username: Union[constr(min_length=2, max_length=30), Literal[""]]
+    ip: Union[IPv4Address, Literal[""]]
+    agent: str
+    host: str
+    referer: Union[AnyUrl, Literal[""]]
+    accept_language: str
+    event_source: Literal["server"]
+    context: BaseContextModel
+    time: datetime
+    page: Union[Literal[None], None]
--- a/src/ralph/models/edx/server.py
+++ b/src/ralph/models/edx/server.py
@@ -1,0 +1,47 @@
+"""Server event model definitions"""
+
+from pathlib import Path
+
+from pydantic import Json, root_validator
+
+from .base import BaseEventModel, BaseModelWithConfig
+
+
+class ServerEventField(BaseModelWithConfig):
+    """Represents the `event` field of the ServerEventModel"""
+
+    GET: dict
+    POST: dict
+
+
+class ServerEventModel(BaseEventModel):
+    """Represents a common server event.
+
+    This type of event is triggered from the django middleware on each request excluding:
+    `/event`, `login`, `heartbeat`, `/segmentio/event` and `/performance`.
+
+    Attributes:
+        event_type (str): Consist of the relative URL (without the hostname) of the requested page.
+            Retrieved with:
+                `request.META['PATH_INFO']`
+        event (str): Consist of a JSON string holding the content of the GET or POST request.
+            Retrieved with:
+                `json.dumps({'GET': dict(request.GET), 'POST': dict(request.POST)})[:512]`
+            Note:
+                Values for ['password', 'newpassword', 'new_password', 'oldpassword',
+                'old_password', 'new_password1', 'new_password2'] are replaced by `********`.
+                The JSON string is truncated at 512 characters resulting in invalid JSON.
+    """
+
+    event_type: Path
+    event: Json[ServerEventField]  # pylint: disable=unsubscriptable-object
+
+    @root_validator
+    def validate_event_type(
+        cls, values
+    ):  # pylint: disable=no-self-argument, no-self-use
+        """Check that the event_type and context.path values are equal"""
+
+        if values.get("event_type") != values.get("context").path:
+            raise ValueError("event_type should be equal to context.path")
+        return values
--- a/tests/backends/database/test_es.py
+++ b/tests/backends/database/test_es.py
@@ -17,8 +17,8 @@ from ralph.defaults import APP_DIR, HISTORY_FILE
 from tests.fixtures.backends import ES_TEST_HOSTS, ES_TEST_INDEX
 
 
-def test_es_database_instanciation(es):
-    """Test the ES backend instanciation"""
+def test_es_database_instantiation(es):
+    """Test the ES backend instantiation"""
     # pylint: disable=invalid-name,unused-argument,protected-access
 
     assert ESDatabase.name == "es"
@@ -41,7 +41,7 @@ def test_es_database_instanciation(es):
 
 
 def test_es_client_kwargs(es):
-    """Test the ES backend client instanciation using client_options that must be
+    """Test the ES backend client instantiation using client_options that must be
     passed to the http(s) connection pool"""
     # pylint: disable=invalid-name,unused-argument,protected-access
 
@@ -125,7 +125,7 @@ def test_put_method(es, fs, monkeypatch):
 
     # Prepare fake file system
     fs.create_dir(str(APP_DIR))
-    # Force Path instanciation with fake FS
+    # Force Path instantiation with fake FS
     history_file = Path(str(HISTORY_FILE))
     assert not history_file.exists()
 
--- a/tests/fixtures/edx/base.py
+++ b/tests/fixtures/edx/base.py
@@ -1,0 +1,44 @@
+"""Base event factory definitions"""
+
+from datetime import timezone
+
+from factory import Factory, LazyAttribute, SubFactory
+from faker import Faker
+
+from ralph.models.edx.base import BaseContextModel, BaseEventModel
+
+Faker.seed(1)
+FAKE = Faker()
+
+
+class BaseContextFactory(Factory):
+    """Base context factory inherited by all context factories"""
+
+    class Meta:  # pylint: disable=missing-class-docstring
+        model = BaseContextModel
+
+    course_user_tags = FAKE.pydict(value_types=str)
+    path = f"/{FAKE.uri_path()}"
+    user_id = FAKE.random_element(elements=(FAKE.random_int(), "", None))
+    org_id = FAKE.word()
+    course_id = LazyAttribute(
+        lambda self: f"course-v1:{self.org_id}+{FAKE.word()}+{FAKE.word()}"
+    )
+
+
+class BaseEventFactory(Factory):
+    """Base event factory inherited by all event factories"""
+
+    class Meta:  # pylint: disable=missing-class-docstring
+        model = BaseEventModel
+
+    username = FAKE.user_name() if FAKE.boolean() else ""
+    ip = FAKE.ipv4_public()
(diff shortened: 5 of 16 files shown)


src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,171 @@
...
+    course_id: str
+    path: Path
+
+    @root_validator
+    def validate_course_id(
+        cls, values
+    ):  # pylint: disable=no-self-argument, no-self-use
+        """The course_id must match the regex `course-v1:{org_id}\\+.+\\+.+` or an empty string."""
+
+        org_id, course_id = values.get("org_id"), values.get("course_id")
```
jmaupetit, line 81: Not a big fan of this shortener, I would prefer being explicit:
```suggestion
        org_id = values.get("org_id")
        course_id = values.get("course_id")
```

src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,171 @@
...
+        regex = f"course-v1:{org_id}\\+.+\\+.+"
+        if (org_id or course_id) and not re.match(regex, course_id):
+            raise ValueError(f"course_id must match regex `{regex}`")
+        return values
+
+
+class BaseEventModel(BaseModelWithConfig):
+    """Represents the base event model all events inherit from.
+    Does not define event and event_type fields.
```
jmaupetit, line 90:
```suggestion
    
    WARNING: it does not define the event and event_type fields.
```

src/ralph/models/edx/server.py
```diff
@@ -0,0 +1,51 @@
...
+                `request.META['PATH_INFO']`
+            Source:
+                /common/djangoapps/track/views/__init__.py#L106
+        event (str): Consist of a JSON string holding the content of the GET or POST request.
+            Retrieved with:
+                `json.dumps({'GET': dict(request.GET), 'POST': dict(request.POST)})[:512]`
+            Note:
+                Values for ['password', 'newpassword', 'new_password', 'oldpassword',
+                'old_password', 'new_password1', 'new_password2'] are replaced by `********`.
+                The JSON string is truncated at 512 characters resulting in invalid JSON.
```
jmaupetit, line 35: WTF? I think that's an issue... we should investigate how many broken events it generates.

src/ralph/models/edx/server.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    event_type: Path
+    event: Json[ServerEventField]  # pylint: disable=unsubscriptable-object
+
+    @root_validator
+    def validate_event_type(
+        cls, values
+    ):  # pylint: disable=no-self-argument, no-self-use
+        """Check that the event_type and context>path values are equal"""
```
jmaupetit, line 47: More pythonic suggestion :wink: 
```suggestion
        """Check that the event_type and context.path values are equal"""
```

src/ralph/models/edx/server.py
```diff
@@ -0,0 +1,51 @@
...
+
+        if values.get("event_type") != values.get("context").path:
+            raise ValueError("event_type should be equal to context>path")
```
jmaupetit, line 50:
```suggestion
            raise ValueError("event_type should be equal to context.path")
```

tests/backends/database/test_es.py
```diff
@@ -125,7 +125,7 @@ def test_put_method(es, fs, monkeypatch):
 
     # Prepare fake file system
     fs.create_dir(str(APP_DIR))
-    # Force Path instanciation with fake FS
+    # Force Path instantiation with fake FS
```
jmaupetit, line 128: Thank you for fixing my typos :sweat_smile:

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- EdX server event model and factory
```
jmaupetit, line 13:
```suggestion
- EdX server event pydantic model and factory
```

tests/fixtures/edx/server.py
```diff
@@ -0,0 +1,25 @@
...
+    event_type = LazyAttribute(lambda self: self.context.path)
+
+    @lazy_attribute
+    def event(self):  # pylint: disable=no-self-use
+        """Return a randomly empty/filled POST/GET value"""
+
+        empty = '{"POST": {}, "GET": {}}'
+        post = '{"POST": {"input_e8bc718966e9441abc3ccf1a6429ee8b_2_1": ["choice_2"]}, "GET": {}}'
+        get = '{"POST": {}, "GET": {"rpp": ["50"], "page": ["1"]}}'
+        return FAKE.random_element([empty, post, get])
```
jmaupetit, line 25: Why not using a Param instead of a random choice?

tests/fixtures/edx/base.py
```diff
@@ -0,0 +1,58 @@
...
+    accept_language = FAKE.locale()
+    event_source = "server"
+    time = FAKE.iso8601(tzinfo=timezone.utc)
+    page = None
+
+    @lazy_attribute
+    def context(self):
+        """Returns the context field"""
+
+        # pylint: disable=no-member, protected-access
```
jmaupetit, line 54: When you should bypass `protected-access` warnings, it often means you are doing something wrong :grimacing:

tests/fixtures/edx/base.py
```diff
@@ -0,0 +1,58 @@
...
+        field = "course_user_tags"
+        exclude = () if field in self.context_ or FAKE.boolean() else (field,)
+        BaseContextFactory._meta.exclude = exclude
+        return BaseContextFactory(**self.context_)
```
jmaupetit, line 58: Why not using a SubFactory here?

tests/fixtures/logs.py
```diff
@@ -24,3 +24,21 @@ def gelf_logger():
...
+        kwargs: Pass additional keyword arguments to the factory
+
+    Returns:
+        A dict representing the event when `size` is 1
+        A `size` long list of events when `size` is more than 1
+    """
+
+    if size == 1:
+        return factory_class.create(**kwargs).dict()
+    return [event.dict() for event in factory_class.create_batch(size, **kwargs)]
```
jmaupetit, line 44: What do you think of always returning a generator instead?

tests/models/edx/test_base.py
```diff
@@ -0,0 +1,70 @@
...
+def test_valid_base_event(kwargs):
+    """Test that a valid base event does not raise ValidationError."""
+
+    try:
+        event_factory(BaseEventFactory, **kwargs)
+    except ValidationError:
+        pytest.fail(f"Valid base event with {kwargs} should not raise exceptions")
+
+
+REGEX = "course_id must match regex"
```
jmaupetit, line 38: Does not look like a regex to me :wink: 

```suggestion
INVALID_COURSE_ID_ERROR = "course_id must match regex"
```

or repeat the expected error message in your test parameters as the variable name is not that short :stuck_out_tongue:

tests/models/edx/test_base.py
```diff
@@ -0,0 +1,70 @@
...
+        {"referer": "https://www.fun-mooc.fr/"},
+        {"referer": "https://www.fun-mooc.fr/event?event_type=page_close"},
+        {"context_": {"user_id": ""}},
+        {"context_": {"user_id": None}},
+        {"context_": {"user_id": 213456}},
+        {"context_": {"org_id": "", "course_id": ""}},
+        {"context_": {"org_id": "org", "course_id": "course-v1:org+any+any"}},
+    ],
+)
+def test_valid_base_event(kwargs):
```
jmaupetit, line 29: Think of test naming pattern as a namespace:

```suggestion
def test_models_edx_base_event_with_valid_content(kwargs):
```

tests/models/edx/test_base.py
```diff
@@ -0,0 +1,70 @@
...
+        {"agent": ""},
+        {"host": ""},
+        {"referer": ""},
+        {"referer": "https://www.fun-mooc.fr/"},
+        {"referer": "https://www.fun-mooc.fr/event?event_type=page_close"},
+        {"context_": {"user_id": ""}},
+        {"context_": {"user_id": None}},
+        {"context_": {"user_id": 213456}},
+        {"context_": {"org_id": "", "course_id": ""}},
+        {"context_": {"org_id": "org", "course_id": "course-v1:org+any+any"}},
```
jmaupetit, line 26: Can you also provide more substantial events with more than one field?

tests/fixtures/edx/base.py
```diff
@@ -0,0 +1,58 @@
...
+    page = None
+
+    @lazy_attribute
+    def context(self):
+        """Returns the context field"""
+
+        # pylint: disable=no-member, protected-access
+        field = "course_user_tags"
+        exclude = () if field in self.context_ or FAKE.boolean() else (field,)
+        BaseContextFactory._meta.exclude = exclude
```
jmaupetit, line 57: :warning: Think of dependency injection instead of monkeypatching your factory here!

tests/models/edx/test_base.py
```diff
@@ -0,0 +1,70 @@
...
+        ({"context_": {"course_id": "+org+any"}}, REGEX),
+        ({"context_": {"course_id": "course-v1:+course+not_empty"}}, REGEX),
+        ({"context_": {"course_id": "course-v1:NOT_org+course+session"}}, REGEX),
+        ({"context_": {"org_id": "org", "course_id": ""}}, REGEX),
+        ({"context_": {"org_id": "org", "course_id": "course-v1:org"}}, REGEX),
+        ({"context_": {"org_id": "org", "course_id": "course-v1:org+course"}}, REGEX),
+        ({"context_": {"org_id": "org", "course_id": "course-v1:org+course+"}}, REGEX),
+    ],
+)
+def test_invalid_base_event(kwargs, error):
```
jmaupetit, line 66: ditto:

```suggestion
def test_models_edx_base_event_with_invalid_content(kwargs, error):
```

tests/models/edx/test_server.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.mark.parametrize(
+    "kwargs",
+    [
+        {"context_": {"path": "/"}, "event_type": "/"},
+        {"context_": {"path": "/a/valid/path"}, "event_type": "/a/valid/path"},
+        {"event": '{"POST": {"key": ["value"]}, "GET": {}}'},
+        {"event": '{"POST": {}, "GET": {}}'},
+    ],
+)
+def test_valid_server_event(kwargs):
```
jmaupetit, line 19:
```suggestion
def test_models_edx_server_event_with_valid_content(kwargs):
```

tests/models/edx/test_server.py
```diff
@@ -0,0 +1,50 @@
...
+        ({"event": '{"GET": {}}'}, "POST\n  field required"),
+        ({"event": '{"GET": {}, "NOT_POST": {}}'}, "POST\n  field required"),
+        ({"event": '{"POST": {}, "GET": {}, "NOR": {}}'}, "extra fields not permitted"),
+        ({"event": '{"POST": "not_object", "GET": {}}'}, "not a valid dict"),
+        ({"event": '{"POST": {}, "GET": "not_object"}'}, "not a valid dict"),
+        ({"event": '{"POST": 123, "GET": {}}'}, "not a valid dict"),
+        ({"event": '{"POST": {}, "GET": 123}'}, "not a valid dict"),
+    ],
+)
+def test_invalid_base_event(kwargs, error):
```
jmaupetit, line 46:
```suggestion
def test_models_edx_server_event_with_invalid_content(kwargs, error):
```

jmaupetit (commented):
I'm glad to see the beauty and concision of `pydantic` models coming in this project :muscle:

jmaupetit:
One last suggestion concerning the scope of your commit message: maybe prefer `core` to `library`. WDYT?

src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,171 @@
...
+            Retrieved with:
+                `course_id.org` where `course_id` is an `opaque_keys.edx.locator.CourseLocator`
+                which is created using the URL of the requested page.
+            Note:
+                Is an empty string when the requested page is not a course page.
+            Source:
+                /common/djangoapps/track/contexts.py#L55
+        course_id (str): Consists of the unique identifier for the visited course page.
+            Retrieved with:
+                `course_id.to_deprecated_string()` where `course_id` is an
```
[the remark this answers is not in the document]
SergioSim: Good point) Yes, indeed, it seems before the course_id was slash separated `org/course/run` but at some point it was changed to `course-v1:org+course+session`.
The [to_deprecated_string](https://github.com/edx/opaque-keys/blame/addcb74adaed6dd9e6d10f01d01414cbe9f572a6/opaque_keys/edx/locations.py#L42-L49) method was deprecated about 7 years ago and removed in [April 2019](https://github.com/edx/opaque-keys/commit/da30c3854c100d593a1a6a6725ac9f3ab075ccb4#diff-b233917bd11b79f38722cb751e329f152bffe87146c75a800d4cd7d45b27a750L41-L49).

All the method does is redirect the call to `locator.CourseLocator.from_string` which is the new/current (course-v1:org+course+session) way to get the course_id string.

I suppose (hope) this change affects only quite old events.

src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,171 @@
...
+                    UserCourseTag.objects.filter(
+                        user=request.user.pk, course_id=course_key
+                    ).values_list('key', 'value')
+                )`
+            Note:
+                Is only present when a course page is requested.
+                Is an empty dictionary when the user is not logged in or not found in the
+                `user_api_usercoursetag` table.
+            Source:
+                /openedx/core/djangoapps/user_api/middleware.py#L38-L46
```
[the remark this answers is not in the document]
SergioSim: This is true! Thanks) Removing them)

src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,176 @@
+"""Base event model definitions"""
+
+import re
+from datetime import datetime
+from ipaddress import IPv4Address
+from pathlib import Path
+from typing import Literal, Optional, Union
+
+from pydantic import (  # pylint: disable=no-name-in-module
```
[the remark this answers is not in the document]
SergioSim: Thanks) whitelisting looks cleaner)

src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,171 @@
...
+    user_id: Union[int, Literal[""], None]
+    org_id: str
+    course_id: str
+    path: Path
+
+    @root_validator
+    def validate_course_id(
+        cls, values
+    ):  # pylint: disable=no-self-argument, no-self-use
+        """The course_id must match the regex `course-v1:{org_id}\\+.+\\+.+` or an empty string."""
```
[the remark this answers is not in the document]
SergioSim: Good Idea! Thanks) updated)

src/ralph/models/edx/server.py
SergioSim: Good Idea) The current schema tries to parse the event string to JSON and when it fails it throws a ValidationError. When we will have the validate command we could group the errors by error type ? Maybe if this happens often we could skip the `event` field to JSON validation ?

tests/fixtures/edx/server.py
SergioSim: Good point, yes indeed.
This made me think that actually this part is not used in the tests. Maybe we could make it just one string ?

tests/fixtures/edx/base.py
SergioSim: You are right :sweat_smile: I was trying to control, if the course_user_tags are present or excluded, but didn't found a right way to do it. Actually this part is not used in the tests - removing it

tests/fixtures/edx/base.py
SergioSim: Good idea! Thanks) Updated)

tests/fixtures/logs.py
SergioSim: Good idea) Thanks) As this method would become a single line, I have removed it)

tests/models/edx/test_base.py
SergioSim: Yes, indeed it's not a regex :sweat_smile:
To be honest I tried to make pylint keep one test per line, but with the subfactory syntax and long error messages, pylint takes over. Updated)

tests/models/edx/test_base.py
SergioSim: Thanks) good point) updated

tests/models/edx/test_base.py
SergioSim: It's done by the factory, the tests parameters only specify which fields we alter.

tests/fixtures/edx/base.py
SergioSim: Thanks for the advise! Unfortunately I wasn't able to find a right way to exclude fields from the factory. But as it's not used in the test I removed this part)

tests/models/edx/test_base.py
SergioSim: Thanks, updated the test name

SergioSim:
> One last suggestion concerning the scope of your commit message: maybe prefer `core` to `library`. WDYT?

Good point, yes, may be `models`, as we add event models ?

jmaupetit:
> Good point, yes, may be models, as we add event models ?

Good idea.

tests/models/edx/test_base.py
```diff
@@ -0,0 +1,84 @@
...
+        {"referer": "https://www.fun-mooc.fr/event?event_type=page_close"},
+        {"context__user_id": ""},
+        {"context__user_id": None},
+        {"context__user_id": 213456},
+        {"context__org_id": "", "context__course_id": ""},
+        {"context__org_id": "org", "context__course_id": "course-v1:org+any+any"},
+    ],
+)
+def test_models_edx_base_event_with_valid_content(kwargs):
+    """Test that a valid base event does not raise ValidationError."""
```
jmaupetit, line 29:
```suggestion
    """Test that a valid base event does not raise a ValidationError"""
```

tests/models/edx/test_base.py
```diff
@@ -0,0 +1,84 @@
...
+            "course_id must match regex",
+        ),
+        (
+            {"context__org_id": "org", "context__course_id": "course-v1:org+course+"},
+            "course_id must match regex",
+        ),
+    ],
+)
+def test_models_edx_base_event_with_invalid_content(kwargs, error):
+    """Test that an invalid base event raise ValidationError."""
```
quitterie-lcs, line 81:
```suggestion
    """Test that an invalid base event raises ValidationError."""
```

tests/models/edx/test_server.py
```diff
@@ -0,0 +1,49 @@
...
+        ({"event": '{"GET": {}, "NOT_POST": {}}'}, "POST\n  field required"),
+        ({"event": '{"POST": {}, "GET": {}, "NOR": {}}'}, "extra fields not permitted"),
+        ({"event": '{"POST": "not_object", "GET": {}}'}, "not a valid dict"),
+        ({"event": '{"POST": {}, "GET": "not_object"}'}, "not a valid dict"),
+        ({"event": '{"POST": 123, "GET": {}}'}, "not a valid dict"),
+        ({"event": '{"POST": {}, "GET": 123}'}, "not a valid dict"),
+    ],
+)
+def test_models_edx_server_event_with_invalid_content(kwargs, error):
+    """Test that an invalid base event raise ValidationError."""
```
quitterie-lcs, line 46:
```suggestion
    """Test that an invalid base event raises ValidationError."""
```

jmaupetit (commented):
Almost good to me. I'll merge it once squashed and your commit title amended.

src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,143 @@
...
+            Note:
+                Is only present when a course page is requested.
+                Is an empty dictionary when the user is not logged in or not found in the
+                `user_api_usercoursetag` table.
+        user_id (int or str or None): Consists of the ID of the authenticated user.
+            Retrieved with:
+                `request.user.pk` querying the `auth_user` table.
+            Note:
+                Is an integer when the user is found in the `auth_user` table.
+                Is an empty string when an exception is raised while retrieving the id.
```
quitterie-lcs, line 39:
```suggestion
                Is an empty string when an exception is raised while retrieving the ID.
```

src/ralph/models/edx/base.py
```diff
@@ -0,0 +1,143 @@
...
+                    Default language
+        host (str): Consists of the hostname of the server.
+            Retrieved with:
+                `request.META[SERVER_NAME]`
+        referer (Path): Consists of the `Referer` HTTP request header.
+            Retrieved with:
+                `request.META[HTTP_REFERER]`
+            Note:
+                Can be an empty string if the header is not present in the request.
+                Contains the referring url (previous url visited by the user).
```
quitterie-lcs, line 115:
```suggestion
                Contains the referring URL (previous URL visited by the user).
```

SergioSim: Good catch) Thanks)

src/ralph/models/edx/base.py
SergioSim: Good catch) Thanks)

src/ralph/models/edx/base.py
```diff
@@ -66,7 +66,7 @@ class BaseContextModel(BaseModelWithConfig):
     def validate_course_id(
         cls, values
     ):  # pylint: disable=no-self-argument, no-self-use
-        """The course_id must match the regex `course-v1:{org_id}\\+.+\\+.+` or an empty string."""
+        """The course_id must match for example `course-v1:org+course+session` or empty string."""
```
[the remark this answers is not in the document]
SergioSim: Thanks, yes indeed) I had removed it to make the line shorter but then it's wrong) added the **an** back, but shortened `course-v1:org+course+session` to `course-v1:org+course+any` to keep the docstring on one line.

tests/models/edx/test_base.py
SergioSim: Thanks) Updated this but kept the last `.`

SergioSim:
Thanks a lot for you reviews) :pray: squashed the commits into one, a backup is [here](https://github.com/SergioSim/ralph/commits/server-event-bu) for the latest fixup commits made.

jmaupetit (approved):
LGTM, you just need to change your commit emoji :wink:

PR #61, opened 2021-02-12
✨(models) add edX page_close browser event model

SergioSim:
## Purpose

The edX page_close browser events are the most simple and common browser events in the edX tracking logs. We want to validate them.

## Proposal

- [x] PageCloseBrowserEventModel
- [x] PageCloseBrowserEventFactory
- [x] Tests

### DIFF
--- a/src/ralph/models/edx/browser.py
+++ b/src/ralph/models/edx/browser.py
@@ -1,0 +1,45 @@
+"""Browser event model definitions"""
+
+from pathlib import Path
+from typing import Literal, Union
+
+from pydantic import AnyUrl, constr
+
+from .base import BaseEventModel
+
+
+class BaseBrowserEventModel(BaseEventModel):
+    """Represents the base browser event model all browser events inherit from.
+
+    This type of event is triggered on (XHR) POST/GET requests to the `/event` URL.
+
+    Attributes:
+        event_source (str): Consists of the value `browser`.
+        page (Path): Consists of the URL (with hostname) of the visited page.
+            Retrieved with:
+                `window.location.href` from the JavaScript front-end.
+        session (str): Consists of the md5 encrypted Django session key or an empty string
+    """
+
+    event_source: Literal["browser"]
+    page: Union[AnyUrl, Path]
+    session: Union[constr(regex=r"^[a-f0-9]{32}$"), Literal[""]]  # noqa: F722
+
+
+class PageCloseBrowserEventModel(BaseBrowserEventModel):
+    """Represents the page_close browser event.
+
+    This type of event is triggered when the user navigates to the next page
+    or closes the browser window (when the JavaScript `window.onunload` event
+    is called).
+
+    Attributes:
+        name (str): Consists of the value `page_close`
+        event_type (str): Consists of the value `page_close`
+        event (str): Consists of the string value `{}`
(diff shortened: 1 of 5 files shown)


jmaupetit:
Looks great, can you please rebase it on the `master` branch and change the commit emoji :pray:

tests/fixtures/backends.py
```diff
@@ -36,11 +36,11 @@ class NamedClassEnum(Enum):
 
 @pytest.fixture
 def es():
-    """Create / delete an ElasticSearch test index and yield an instanciated client"""
+    """Create / delete an ElasticSearch test index and yield an instantiated client"""
     # pylint: disable=invalid-name
 
     client = Elasticsearch(ES_TEST_HOSTS)
-    client.indices.create(index=ES_TEST_INDEX, ignore=400)
```
SergioSim, line 43: pylint is complaining here about: `Unexpected keyword argument 'ignore' in method call (unexpected-keyword-arg)`.
But this is only a pylint issue.
I propose to remove this argument as the tests pass without it.
Would it be better if we keep the `ignore` and add a `pylint: disable=unexpected-keyword-arg` ?

jmaupetit: :thinking: no feel free to remove non-required argument. Thanks!

src/ralph/models/edx/browser.py
```diff
@@ -0,0 +1,49 @@
...
+        page (Path): Consists of the URL (with hostname) of the visited page.
+            Retrieved with:
+                `window.location.href` from the JavaScript front-end.
+        session (str): Consists of the md5 encrypted Django session key or an empty string
+    """
+
+    event_source: Literal["browser"]
+    page: Union[AnyUrl, Path]
+    session: Union[
+        constr(min_length=MD5_HASH_LEN, max_length=MD5_HASH_LEN), Literal[""]
```
jmaupetit, line 29: I would use a regular expression instead:

```suggestion
        constr(regex=r'^[a-f0-9]{32}$'), Literal[""]
```

(and remove the `MD5_HASH_LEN` definition).

jmaupetit (approved):
I only have minor comments. Great :tada:

src/ralph/models/edx/browser.py
SergioSim: Good Idea) Thanks)

src/ralph/models/edx/browser.py
```diff
@@ -25,9 +23,7 @@ class BaseBrowserEventModel(BaseEventModel):
 
     event_source: Literal["browser"]
     page: Union[AnyUrl, Path]
-    session: Union[
-        constr(regex=r'^[a-f0-9]{32}$'), Literal[""]
-    ]
+    session: Union[constr(regex=r"^[a-f0-9]{32}$"), Literal[""]]  # noqa: F722
```
SergioSim, line 26: It seems flake-8 finds a false positive here https://github.com/PyCQA/pyflakes/issues/542
added # noqa: F722

SergioSim:
Thanks a lot for the review) updated the PR and squashed commits, a backup is [here](https://github.com/SergioSim/ralph/commits/browser-event-page-close-bu) for the latest changes.

PR #66, opened 2021-02-25
ModelSelector

SergioSim:
## Purpose

The ModelSelector selects a matching pydantic model for a given event.

## Proposal

We use the `ModelSelector.add` decorator to collect selection `Rule`s for each Model.
Then we  construct a decision tree based on these rules.
Finally we check the rules in the order defined by the decision tree for each event in the `ModelSelector().get_model(event)` method to retrieve the matching model.

- [x] Add ModelSelector
- [x] Add Tests
  - [x] Decision tree
  - [x] ModelSelector usage
  - [x] Detect/Prohibit model_rules sets that are a subset of another model_rules set.
- [x] ~~Update documentation (add visualization for the decision tree)~~ (will be done in a separate way)


src/ralph/models/edx/server.py
```diff
@@ -14,6 +16,12 @@ class ServerEventField(BaseModelWithConfig):
     POST: dict
 
 
+@ModelSelector.add(
+    [
+        Rule(ModelField("event_source"), "server"),
+        Rule(ModelField("event_type"), ModelField("context>path")),
```
jmaupetit, line 22: Using dots notation would feel more natural for python developers?

```suggestion
        Rule(ModelField("event_type"), ModelField("context.path")),
```

src/ralph/utils.py
```diff
@@ -84,3 +84,16 @@ def now():
     """Return the current UTC time in ISO format"""
 
     return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+
+
+def nested_get(dictionary, keys):
```
jmaupetit, line 89: Secure the implementation by using type hints!

```suggestion
def nested_get(dict_: dict, keys: list[str]):
```

src/ralph/utils.py
```diff
@@ -84,3 +84,16 @@ def now():
...
+    """Get the nested dict value by keys array."""
+
+    if keys is None:
```
jmaupetit, line 92:
```suggestion
    if keys is None or not len(keys):
```

src/ralph/utils.py
```diff
@@ -84,3 +84,16 @@ def now():
...
+        return None
+    for key in keys:
+        try:
+            dictionary = dictionary[key]
+        except (KeyError, TypeError):
```
jmaupetit, line 97:
```suggestion
        except KeyError:
```

src/ralph/utils.py
```diff
@@ -84,3 +84,16 @@ def now():
...
+
+
+def nested_get(dictionary, keys):
+    """Get the nested dict value by keys array."""
+
+    if keys is None:
+        return None
+    for key in keys:
+        try:
+            dictionary = dictionary[key]
```
jmaupetit, line 96:
```suggestion
            dict_ = dict_[key]
```

src/ralph/utils.py
```diff
@@ -84,3 +84,16 @@ def now():
...
+        except (KeyError, TypeError):
+            return None
+    return dictionary
```
jmaupetit, line 99:
```suggestion
    return dict_
```

jmaupetit, line 99: What about using a more concise functional programing implementation?

```python
import operator
from functools import reduce

def get_dict_value_from_path(dict_: dict, path: str, sep: str = "."):
    return reduce(operator.getitem, path.split(sep), dict_)
```

src/ralph/models/edx/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""edX pydantic models"""
+
+from .browser import PageCloseBrowserEventModel  # noqa: F401
+from .server import ServerEventModel  # noqa: F401
```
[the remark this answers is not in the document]
SergioSim: Good Idea, thanks)

src/ralph/utils.py
SergioSim: Looks much better this way! Thanks)

src/ralph/utils.py
SergioSim: Good catch! I haven't foreseen this case. Thanks

src/ralph/utils.py
SergioSim: I have added the `TypeError` in case the `dictionary` is not  a `dict`. But now, with the type hints, I agree that this is not needed anymore. Thanks)

src/ralph/utils.py
SergioSim: This is beautiful ! Thanks)
It is a bit slower https://pastebin.com/hG0F0bab (~18%) but much more readable.
Will update this

src/ralph/utils.py
SergioSim: Thanks)

src/ralph/utils.py
SergioSim: Thanks)

src/ralph/models/edx/server.py
SergioSim: Good point! Yes, indeed, and with the edX events this should work too, but xAPI events use sometimes URI's as keys (containing the `.`)

Some propositions:

- use `.` for edX and `>` for xAPI
- keep `>` for both
- put the keys in a list `[key1, key2]`

WDYT ? Are there other/better solutions ?

src/ralph/utils.py
SergioSim: I have put back the `TypeError`, this happens in case we try to access fields that are not present in the `dict_`.
Example: `keys="a.b.c"` and `dict_={"a": "b"}`, then `nested_get` will try to access `"b"["c"]` which raises a `TypeError`

src/ralph/utils.py
SergioSim: This part has been undone as by using `reduce`, it's returning `None` for empty lists.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Model selector to retrieve the associated model for a given event
```
jmaupetit, line 13:
```suggestion
- Model selector to retrieve the associated pydantic model of a given event
```

src/ralph/models/edx/server.py
jmaupetit: Oh crap! You are right, it's preferable to use the `>` path notation.

src/ralph/utils.py
```diff
@@ -84,3 +85,14 @@ def now():
     """Return the current UTC time in ISO format"""
 
     return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+
+
+def nested_get(dict_: dict, keys: list[str]):
```
jmaupetit, line 90: What about changing the function name and also it's signature?

```suggestion
def get_dict_value_from_path(dict_: dict, path: str, sep: str = ">"):
    keys = path.split(sep)
    # [...]
```

CHANGELOG.md
SergioSim: Good catch) Thanks)

src/ralph/utils.py
SergioSim: Good for the name change) For the signature change I'm hesitating, this function will be called many times per event and having to split the path at every time can be avoided by splitting the path once at the beginning.
WDYT ?

jmaupetit (commented):
This is a wonderful work :muscle: 

We will need to discuss IRL some points, but it's really nice.

src/ralph/models/selector.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@dataclass(frozen=True)
+class ModelField:
+    """Represents a model field."""
+
+    path: tuple[str]
+
+    def __init__(self, path: str):
+        object.__setattr__(self, "path", tuple(path.split(">")))
```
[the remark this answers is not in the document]
SergioSim: I like this Idea) Thanks) This looks a lot more standard) FactoryBoy uses also the `__` separator)

src/ralph/models/selector.py
```diff
@@ -0,0 +1,141 @@
...
+    value: Union[ModelField, Any]
+    operation: Callable[[Any, Any], bool] = operator.eq
+
+    @staticmethod
+    def get_field_value(field, event, known_fields):
+        """Retrieves the value of a rule field.
+
+        Args:
+            field (ModelField or Any): The rule field to retrieve its value.
+            known_fields (dict): A mapping of already retrieved fields of the event.
```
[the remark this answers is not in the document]
SergioSim: Thanks)

tests/test_utils.py
```diff
@@ -94,3 +94,25 @@ def __init__(self, force=False, verbose=0):
...
+def test_get_dict_value_from_path_should_return_given_value():
+    """The get_dict_value_from_path function should return the value when it's present."""
+
+    dictionary = {"foo": {"bar": "bar_value"}}
+    assert (
+        ralph_utils.get_dict_value_from_path(dictionary, ["foo", "bar"]) == "bar_value"
+    )
+
+
+def test_get_dict_value_from_path_should_return_null_when_value_does_not_exists():
```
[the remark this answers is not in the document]
SergioSim: Good catch) Indeed it's `none`)

tests/test_utils.py
```diff
@@ -94,3 +94,25 @@ def __init__(self, force=False, verbose=0):
...
+    """The get_dict_value_from_path function should return null if the value is not found."""
+
+    dictionary = {"foo": {"bar": "bar_value"}}
+    assert (
+        ralph_utils.get_dict_value_from_path(dictionary, ["foo", "bar", "baz", "qux"])
+        is None
+    )
+    assert ralph_utils.get_dict_value_from_path(dictionary, ["foo", "not_bar"]) is None
+    assert ralph_utils.get_dict_value_from_path(dictionary, ["not_foo", "bar"]) is None
+    assert ralph_utils.get_dict_value_from_path(dictionary, None) is None
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks)

src/ralph/utils.py
```diff
@@ -84,3 +86,14 @@ def now():
     """Return the current UTC time in ISO format"""
 
     return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+
+
+def get_dict_value_from_path(dict_: dict, path: list[str]):
+    """Get the nested dict value by keys array."""
```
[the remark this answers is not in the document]
SergioSim: Thanks, updated the docstring similarly (limit the length to 100 chars):
`"""Get a nested dictionary value using an array of keys representing the path to the value."""`

tests/test_utils.py
```diff
@@ -94,3 +94,28 @@ def __init__(self, force=False, verbose=0):
...
+        ["foo", "not_bar"],
+        ["not_foo", "bar"],
+        None,
+    ],
+)
+def test_get_dict_value_from_path_should_return_none_when_value_does_not_exists(path):
+    """The get_dict_value_from_path function should return None if the value is not found."""
+
+    dictionary = {"foo": {"bar": "bar_value"}}
+    print("the path is:", path)
```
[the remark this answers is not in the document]
SergioSim: Nice catch) thanks)

src/ralph/models/selector.py
```diff
@@ -0,0 +1,141 @@
...
+    def add(cls, rules: list[Rule]):
+        """Event model decorator used to add event models to the model_rules list."""
+
+        def __inner__(model):
+            """Add model and rules for a given event."""
+
+            cls.model_rules[model] = rules
+            return model
+
+        return __inner__
```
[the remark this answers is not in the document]
SergioSim: I like this new approach! Thanks) Indeed the previous method made the rule association implicit and might introduce bugs.
But I still load the edx module while instantiating to avoid circular imports. `server_event depends on Rule, Rule depends on model_selector, model_selector depends on server_event`.

tests/test_utils.py
```diff
@@ -94,3 +94,27 @@ def __init__(self, force=False, verbose=0):
     )
     assert test.force
     assert test.verbose == 2
+
+
+def test_get_dict_value_from_path_should_return_given_value():
+    """The get_dict_value_from_path function should return the value when it's present."""
+
+    dictionary = {"foo": {"bar": "bar_value"}}
+    path = ["foo", "bar"]
```
quitterie-lcs, line 103: To have a uniform test structure for `get_dict_value_from_path`, why don't parametrize this test for `path` values?

scripts/generate_docs.py
```diff
@@ -0,0 +1,71 @@
...
+            "",
+        ))
+
+    def add_line(self, line):
+        """Adds the line to the markdown content"""
+
+        self.content += f"{line}\n\n"
+
+    def add_header(self, header, level=1):
+        """Adds a header to the markdown document"""
```
quitterie-lcs, line 29:
```suggestion
        """Adds a header to the markdown document."""
```

scripts/generate_docs.py
```diff
@@ -0,0 +1,71 @@
...
+    def save(self, file_name):
+        """Save MarkdownWriter.content to file"""
+
+        os.makedirs(os.path.dirname(file_name), exist_ok=True)
+        with open(file_name, "w") as file:
+            file.write(self.content)
+
+
+def write_decision_tree(decision_tree, line="", intend=1):
+    """Recursively write the if statements representing the decision tree"""
```
quitterie-lcs, line 42:
```suggestion
    """Recursively writes the if statements representing the decision tree."""
```

scripts/generate_docs.py
```diff
@@ -0,0 +1,71 @@
...
+                if isinstance(decision_tree[rule][value], dict):
+                    depth_change = intend + int(value)
+                    line += write_decision_tree(decision_tree[rule][value], "", depth_change)
+                elif decision_tree[rule][value]:
+                    line += f"{'    ' * (intend + 1)}{decision_tree[rule][value].__name__}\n"
+        return line
+
+
+def generate_model_selector_docs(file_name):
+    """Generate Markdown documentation for ModelSelector"""
```
quitterie-lcs, line 61:
```suggestion
    """Generates Markdown documentation for ModelSelector."""
```

src/ralph/models/selector.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Attributes:
+        model_rules (dict): Stores the list of rules for each model.
+        decision_tree (dict): Stores the rule checking order for model selection.
+    """
+
+    model_rules: dict[BaseModel, list[Rule]] = {}
+
+    def __init__(self, module="ralph.models.edx"):
+        """Instantiate ModelSelector."""
```
quitterie-lcs, line 59:
```suggestion
        """Instantiates ModelSelector."""
```

src/ralph/models/selector.py
```diff
@@ -0,0 +1,132 @@
...
+            position = self.decision_tree
+        rule = next(iter(position))
+        direction = rule.check(event, known_fields)
+        next_position = position[rule][direction]
+        if not isinstance(next_position, dict):
+            return next_position
+        return self.get_model(event, known_fields, next_position)
+
+    def get_decision_tree(self, model_rules):
+        """Recursively constructs the decision tree"""
```
quitterie-lcs, line 87:
```suggestion
        """Recursively constructs the decision tree."""
```

src/ralph/utils.py
```diff
@@ -84,3 +86,14 @@ def now():
     """Return the current UTC time in ISO format"""
 
     return datetime.datetime.now(tz=datetime.timezone.utc).isoformat()
+
+
+def get_dict_value_from_path(dict_: dict, path: list[str]):
+    """Get a nested dictionary value using an array of keys representing the path to the value."""
```
quitterie-lcs, line 92:
```suggestion
    """Gets a nested dictionary value using an array of keys representing the path to the value."""
```

tests/models/test_selector.py
```diff
@@ -0,0 +1,113 @@
...
+                            False: None,
+                        }
+                    },
+                }
+            },
+        ),
+    ],
+)
+def test_models_selector_model_selector_decision_tree(model_rules, decision_tree):
+    """Given `model_rules` the ModelSelector should create the expected decision tree"""
```
quitterie-lcs, line 83:
```suggestion
    """Given `model_rules` the ModelSelector should create the expected decision tree."""
```

tests/models/test_selector.py
```diff
@@ -0,0 +1,113 @@
...
+@pytest.mark.parametrize(
+    "event,model",
+    [
+        (ServerEventFactory(), ServerEventModel),
+        (PageCloseBrowserEventFactory(), PageCloseBrowserEventModel),
+        ({"invalid": "event"}, None),
+    ],
+)
+def test_models_selector_model_selector_get_model(event, model):
+    """Given an event the ModelSelector.get_model method should return the corresponding model"""
```
quitterie-lcs, line 98:
```suggestion
    """Given an event the ModelSelector.get_model method should return the corresponding model."""
```

tests/models/test_selector.py
```diff
@@ -0,0 +1,113 @@
...
+
+    event = json.loads(event.json()) if not isinstance(event, dict) else event
+    assert ModelSelector().get_model(event) is model
+
+
+def test_models_selector_model_selector_model_rules():
+    """Check that no model_rules set is included in another model_rules set"""
```
quitterie-lcs, line 105:
```suggestion
    """Checks that no model_rules set is included in another model_rules set."""
```

quitterie-lcs (commented):
Congrats for this great job! 
I will have to ask for clarification in live to better understand some choices you've made
I took the liberty of making suggestions about docstrings typo in the perspective of the ralph documentation (to get a uniform one)

src/ralph/models/selector.py
```diff
@@ -0,0 +1,141 @@
...
+            known_fields (dict): A mapping of already retrieved fields of the event.
+        """
+
+        if not isinstance(field, ModelField):
+            return field
+        if field.path in known_fields:
+            return known_fields[field.path]
+        value = get_dict_value_from_path(event, field.path)
+        known_fields[field.path] = value
+        return value
```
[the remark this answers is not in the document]
SergioSim: Good point) Thanks, yes indeed this is confusing for the get_field_value method - removing it.

For the known_fields cache - it's true that for now there is no added value, it might be useful (later) when we will have a lot of rules checking for nested values - but this can be done later if needed - removing it too)

scripts/generate_docs.py
```diff
@@ -0,0 +1,71 @@
...
+
+        self.content += f"{line}\n\n"
+
+    def add_header(self, header, level=1):
+        """Adds a header to the markdown document"""
+
+        self.add_line(f"{'#' * level} {header}")
+
+    def save(self, file_name):
+        """Save MarkdownWriter.content to file"""
```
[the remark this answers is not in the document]
SergioSim: Thanks)

src/ralph/utils.py
SergioSim: Thanks) This will make exactly 100 character for this line)

tests/test_utils.py
SergioSim: Good Idea) Thanks) Indeed this will make it more uniform)

src/ralph/models/selector.py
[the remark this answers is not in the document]
jmaupetit: Okay. that's perfectly fine this way.

src/ralph/models/edx/server.py
```diff
@@ -33,6 +35,11 @@ class ServerEventModel(BaseEventModel):
                 The JSON string is truncated at 512 characters resulting in invalid JSON.
     """
 
+    __selector__ = (
+        Rule(ModelField("event_source"), "server"),
+        Rule(ModelField("event_type"), ModelField("context__path")),
```
jmaupetit, line 40: From an API user perspective, I would love to write rules as ORM-like queries:

```suggestion
        Rule(event_type=LazyModelField("context__path")),
```

We'll discuss this IRL, but you got the idea.

docs/models/selector.md
```diff
@@ -0,0 +1,26 @@
...
+## Decision Tree
+
+    If event.event_source == browser:
+        If event.event_type == page_close:
+            PageCloseBrowserEventModel
+    If event.event_source == server:
+        If event.event_type == event.context.path:
+            ServerEventModel
+
+
```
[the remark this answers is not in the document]
SergioSim: Good Point) Thanks) Yes indeed, the documentation for `ModelSelector` is much better documented using `mkdocstring`)

Updated the PR to use `mkdocstring` for the `ModelSelector` documentation.

The only part that is missing is the decision tree visualization which is made with the `generate_docs` script.
I think it might be useful to be able to see what happens with the decision tree when we add a new model ? WDYT ?

src/ralph/models/edx/server.py
SergioSim: This looks interesting) Thanks) I will have a look how to make this work)

SergioSim: Updated the PR with this change) But I kept the optional `_operation` argument to be able to use other operations than equality for the checks (hope no field will be named `_operation` in the future models).

src/ralph/models/selector.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+
+        field = get_dict_value_from_path(event, self.field.path)
+        value = self.value
+        if isinstance(value, LazyModelField):
+            value = get_dict_value_from_path(event, value.path)
+        return field == value
+
+
+def selector(**fields_values):
```
jmaupetit, line 46:
```suggestion
def selector(**filters):
```

src/ralph/models/selector.py
```diff
@@ -0,0 +1,126 @@
...
+    """Returns a list of rules that should match in order to select an event.
+
+    Args:
+        fields_values (keyword arguments): the keywords represent the path to a
+            field and the values represent the value the field should match.
+    """
+
+    return [
+        Rule(LazyModelField(field), value) for field, value in fields_values.items()
```
jmaupetit, line 55: That's smart :wink:

src/ralph/models/selector.py
```diff
@@ -0,0 +1,126 @@
...
+        decision_tree (dict): Stores the rule checking order for model selection.
+    """
+
+    model_rules: dict[BaseModel, list[Rule]] = {}
+
+    def __init__(self, module="ralph.models.edx"):
+        """Instantiates ModelSelector."""
+
+        for model in import_module(module).__dict__.values():
+            if isinstance(model, type) and model.__selector__:
```
jmaupetit, line 73: Can we be more specific?

```suggestion
            if isinstance(model, BaseModel) and model.__selector__:
```

src/ralph/models/selector.py
SergioSim: Thanks)

src/ralph/models/selector.py
SergioSim: Good Idea) Thanks)

src/ralph/models/selector.py
SergioSim: Good catch) thanks, yes indeed, have added `issubclass(model, BaseModel)`, maybe I should rename model to `model_class` ?

scripts/generate_docs.py
```diff
@@ -0,0 +1,73 @@
...
+    doc = MarkdownDocument()
+    doc.add_header("Model Selector")
+    doc.add_line(model_selector.__doc__)
+    doc.add_header("Decision Tree", 2)
+    doc.add_line(write_decision_tree(model_selector.decision_tree))
+    doc.save(file_name)
+
+
+if __name__ == "__main__":
+    generate_model_selector_docs("./docs/models/selector.md")
```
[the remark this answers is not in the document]
SergioSim: Thanks, removed this part for now

Makefile
```diff
@@ -46,6 +46,7 @@ docker-hub: build
 .PHONY: docker-hub
 
 docs-build: ## build documentation site
+	@$(COMPOSE_RUN) app python3 ./scripts/generate_docs.py
```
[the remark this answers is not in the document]
SergioSim: Thanks, removed this part)

src/ralph/models/selector.py
SergioSim: Indeed the typing was misleading, model is of class type, updated the typing to Type[BaseModel] and replaced model with model_class name.

src/ralph/models/selector.py
```diff
@@ -70,7 +68,11 @@ def __init__(self, module="ralph.models.edx"):
         """Instantiates ModelSelector."""
 
         for model in import_module(module).__dict__.values():
-            if isinstance(model, type) and model.__selector__:
+            if (
+                isinstance(model, type)
+                and issubclass(model, BaseModel)
+                and model.__selector__
+            ):
```
jmaupetit: In fact it should be `object` instead of `model` as it could be anything from the module.

SergioSim: You are right, will update that, thanks)

src/ralph/models/selector.py
```diff
@@ -0,0 +1,128 @@
...
+            true_model_rules = next(iter(true_model_rules))
+        elif true_model_rules:
+            true_model_rules = self.get_decision_tree(true_model_rules)
+
+        if false_model_rules:
+            false_model_rules = self.get_decision_tree(false_model_rules)
+        else:
+            false_model_rules = None
+
+        return {root_rule: {True: true_model_rules, False: false_model_rules}}
```
jmaupetit, line 128: A general comment on the decision-tree like implementation: I think some variable names are not relevant and I find it hard to understand those pieces of code. I propose the following alternative algorithm:

1. Set/build the `decision_tree` during instantiation (and drop the `model_rules`) along with a rules registry (all `Rule` instances for a module, _e.g._ `edx`),
2. Given 3 models `A`, `B` and `C`, responding respectively to the following rules (`alpha`, `beta`), (`beta`), (`gamma`, `delta`), I propose the following decision tree structure:

```python
decision_tree = {
  alpha: {
    beta: A
  }, 
  beta: B, 
  gamma: {
    delta: C
  }, 
  delta: None
}
```

3. Recursively walking through this tree would stop when the node value is a model class, meaning that we've found the model corresponding to the input event.

WDYT?

src/ralph/models/selector.py
```diff
@@ -0,0 +1,128 @@
...
+    value: Union[LazyModelField, Any]
+
+    def check(self, event):
+        """Checks if event matches the rule.
+
+        Args:
+            event (dict): The event to check.
+        """
+
+        field = get_dict_value_from_path(event, self.field.path)
```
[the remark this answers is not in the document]
SergioSim: Good Idea! Thanks)

src/ralph/models/selector.py
```diff
@@ -0,0 +1,128 @@
...
+            value = get_dict_value_from_path(event, value.path)
+        return field == value
+
+
+def selector(**filters):
+    """Returns a list of rules that should match in order to select an event.
+
+    Args:
+        **filters: Each keyword represents the path to a field, the corresponding
+            value represents the value which the field should match.
```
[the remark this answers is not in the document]
SergioSim: Yes sure) Good point) Thanks)

src/ralph/models/selector.py
```diff
@@ -0,0 +1,128 @@
...
+    model_rules: dict[Type[BaseModel], list[Rule]] = {}
+
+    def __init__(self, module="ralph.models.edx"):
+        """Instantiates ModelSelector."""
+
+        for object_ in import_module(module).__dict__.values():
+            if (
+                isinstance(object_, type)
+                and issubclass(object_, BaseModel)
+                and object_.__selector__
```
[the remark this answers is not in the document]
SergioSim: Nice catch) Thanks

tests/models/test_selector.py
```diff
@@ -0,0 +1,113 @@
...
+        (
+            {
+                PageCloseBrowserEventModel: [
+                    Rule("event_source", "browser"),
+                    Rule("event_type", "page_close"),
+                ],
+                ServerEventModel: [
+                    Rule("event_source", "server"),
+                    Rule("event_type", LazyModelField("context__path")),
+                ],
```
jmaupetit, line 56: I think you should add more complex scenario with similar events using common rules.

src/ralph/models/selector.py
jmaupetit: As discussed IRL, this proposal adds too much complexity in its implementation. We will stick to a binary decision tree.

src/ralph/models/selector.py
```diff
@@ -0,0 +1,130 @@
...
+    def get_decision_tree(self, model_rules):
+        """Recursively constructs the decision tree."""
+
+        rule_counter = Counter(chain.from_iterable(model_rules.values()))
+        if not rule_counter:
+            return {}
+
+        root_rule = rule_counter.most_common(1)[0][0]
+        true_model_rules = {}
+        false_model_rules = {}
```
jmaupetit, line 111:
```suggestion
        true_subtree = {}
        false_subtree = {}
```

tests/models/test_selector.py
```diff
@@ -0,0 +1,113 @@
...
+def test_models_selector_model_selector_model_rules():
+    """Checks that no model_rules set is included in another model_rules set."""
+
+    model_rules = ModelSelector().model_rules
+    for model, rules in model_rules.items():
+        rules_set = set(rules)
+        other_model_rules = filter(lambda x, m=model: x[0] != m, model_rules.items())
+        for other_model, other_rules in other_model_rules:
+            if rules_set.issubset(set(other_rules)):
+                pytest.fail(f"The rules of {model} are a subset of {other_model}")
```
jmaupetit, line 113: That's a great idea, but as discussed IRL it should be checked while instantiating the decision tree.

jmaupetit (commented):
As already said, that's a great piece of work. Keep it up :muscle:

src/ralph/models/selector.py
```diff
@@ -0,0 +1,130 @@
...
+        for object_ in import_module(module).__dict__.values():
+            if (
+                isinstance(object_, type)
+                and issubclass(object_, BaseModel)
+                and hasattr(object_, "__selector__")
+            ):
+                self.model_rules[object_] = object_.__selector__
+        self.decision_tree = self.get_decision_tree(self.model_rules)
+
+    def get_model(self, event: dict, position=None):
```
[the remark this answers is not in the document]
SergioSim: Good point) Thanks) yes, indeed, `tree` is more representative for this value

src/ralph/models/selector.py
```diff
@@ -0,0 +1,130 @@
...
+            model (BaseModel): When event matches all rules of the model.
+            None: Otherwise.
+        """
+
+        if position is None:
+            position = self.decision_tree
+        rule = next(iter(position))
+        direction = rule.check(event)
+        next_position = position[rule][direction]
+        if not isinstance(next_position, dict):
```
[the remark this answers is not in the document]
SergioSim: Good Idea, thanks) Added the comment:
`# Here we have found the model or know that the event does not match any model.`

src/ralph/models/selector.py
```diff
@@ -0,0 +1,130 @@
...
+        return self.get_model(event, next_position)
+
+    def get_decision_tree(self, model_rules):
+        """Recursively constructs the decision tree."""
+
+        rule_counter = Counter(chain.from_iterable(model_rules.values()))
+        if not rule_counter:
+            return {}
+
+        root_rule = rule_counter.most_common(1)[0][0]
```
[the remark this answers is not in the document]
SergioSim: Good Idea, thanks) I have to admit that the rules are not sorted - only counted in this case.
Added the comment:
`# We retrieve the most occurred rule and use it to split the decision tree in two:`
`# 1 - true_subtree: a subtree of models that require the most occurred rule to be true.`
`# 2 - false_subtree: a subtree of models that don't use the most occurred rule.`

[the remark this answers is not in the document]
jmaupetit: Suggestion:

> We retrieve the rule with the highest occurrence and use it to split the decision tree in two subtrees:

src/ralph/models/selector.py
```diff
@@ -106,25 +107,26 @@ def get_decision_tree(self, model_rules):
         if not rule_counter:
             return {}
 
+        # We retrieve the most occurred rule and use it to split the decision tree in two:
+        # 1 - true_subtree: a subtree of models that require the most occurred rule to be true
+        # 2 - false_subtree: a subtree of models that don't use the most occurred rule
         root_rule = rule_counter.most_common(1)[0][0]
```
jmaupetit, line 113: Edge case: what if two (or more) rules have the same number of occurrences?

src/ralph/models/selector.py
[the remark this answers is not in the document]
SergioSim: Thanks, good suggestion)

src/ralph/models/selector.py
SergioSim: Good point, indeed this is not covered.
I ran into it just now trying to use sets instead of lists for the rules as there is a lot of "rule" in "rule_list" usage.
With sets it's undefined behavior (they are not ordered) - tests some times pass - sometimes not.
With lists the order is determined by the order we create the `model_rules`

SergioSim: > Good point, indeed this is not covered.
> I ran into it just now trying to use sets instead of lists for the rules as there is a lot of "rule" in "rule_list" usage.
> With sets it's undefined behavior (they are not ordered) - tests some times pass - sometimes not.
> With lists the order is determined by the order we create the `model_rules` (not 100%, I will have to check this)

I was wrong :sweat_smile: the order will be determined by the hash value of the rule, [example](https://trinket.io/python/ad3599ed63) and [docs](https://docs.python.org/3/library/collections.html#collections.Counter)

I will retry using sets instead, just to see why it doesn't work.
Is there any action to be taken in this edge case ?

SergioSim: :facepalm: The example was in python2, python3 does use the order of insertion, that's why using sets didn't worked for me. [Example](https://trinket.io/python3/77bfdc2e26)

tests/models/test_selector.py
SergioSim: Good Idea) Thanks) I have added one more test with 3 models, multiple rules, where one rule occurs twice.

tests/models/test_selector.py
SergioSim: Thanks) You are right) Although it costs a bit of processing at the instantiation it's worth to check the decision tree while building it.

src/ralph/models/selector.py
```diff
@@ -0,0 +1,147 @@
...
+        for object_ in import_module(module).__dict__.values():
+            if (
+                isinstance(object_, type)
+                and issubclass(object_, BaseModel)
+                and hasattr(object_, "__selector__")
+            ):
+                self.add_to_model_rules(object_, object_.__selector__)
+        self.decision_tree = self.get_decision_tree(self.model_rules)
+
+    def add_to_model_rules(self, new_model: Type[BaseModel], new_rules: list[Rule]):
```
SergioSim, line 81: Maybe we should enforce this check by replacing model_rules with a subclass of dict with a __setter__ method ? [Example](https://stackoverflow.com/a/3137768)
WDYT ?

src/ralph/models/selector.py
```diff
@@ -64,20 +65,36 @@ class ModelSelector:
...
         self.decision_tree = self.get_decision_tree(self.model_rules)
 
+    def add_to_model_rules(self, new_model: Type[BaseModel], new_rules: list[Rule]):
+        """Adds a new model with it's list of rules to the model_rules dictionary.
+
+        Raises:
+            ModelRulesException: When new_rules list is a subset or superset of another rules list.
+        """
+
+        rules_set = set(new_rules)
```
jmaupetit, line 88: Let's consider that rules are always sets:

```suggestion
        new_rules = set(new_rules)
```

src/ralph/models/selector.py
jmaupetit: Good idea.

src/ralph/models/selector.py
```diff
@@ -64,20 +65,36 @@ class ModelSelector:
...
             if (
                 isinstance(object_, type)
                 and issubclass(object_, BaseModel)
                 and hasattr(object_, "__selector__")
             ):
-                self.model_rules[object_] = object_.__selector__
+                self.add_to_model_rules(object_, object_.__selector__)
         self.decision_tree = self.get_decision_tree(self.model_rules)
 
+    def add_to_model_rules(self, new_model: Type[BaseModel], new_rules: list[Rule]):
```
jmaupetit, line 81: I would suggest to isolate this utility in a static method and make it more testable by including modules loading:

```suggestion
    @staticmethod
    def build_model_rules(module):
```

src/ralph/models/selector.py
```diff
@@ -64,20 +65,36 @@ class ModelSelector:
...
 
+        self.model_rules: dict[Type[BaseModel], list[Rule]] = {}
         for object_ in import_module(module).__dict__.values():
             if (
                 isinstance(object_, type)
                 and issubclass(object_, BaseModel)
                 and hasattr(object_, "__selector__")
             ):
-                self.model_rules[object_] = object_.__selector__
+                self.add_to_model_rules(object_, object_.__selector__)
```
jmaupetit, line 78:
```suggestion
        self.model_rules = ModelSelector.build_model_rules(module)
```

src/ralph/models/selector.py
SergioSim: Good point) I would really like to use sets for the rules (tried twice) but unfortunately the output of the decision tree depends on the order of the rules. see https://github.com/openfun/ralph/pull/66#discussion_r594695806
The output with rules in a different order will be equally right (Edge case: when two (or more) rules have the same number of occurrences) but different - which breaks the tests :disappointed:

jmaupetit: I just wanted to use sets for rules in this method, not generally.

src/ralph/models/selector.py
SergioSim: Thanks) Updated the PR)

src/ralph/models/selector.py
SergioSim: Good Idea) Thanks) Updated the PR with the build_model_rules method and added a test.

src/ralph/models/selector.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+
+        rules_set = set(new_rules)
+        for model, rules in self.items():
+            if rules_set.issubset(rules) or rules_set.issuperset(rules):
+                raise ModelRulesException(
+                    f"{new_model.__name__} rules are a subset or superset "
+                    f"of {model.__name__} rules"
+                )
+        super().__setitem__(new_model, new_rules)
```
jmaupetit, line 66: As the number of supported events/models grows, we should pay attention on this as it does not scale well... I mean not for now, but in a near future.

tests/models/test_selector.py
```diff
@@ -0,0 +1,223 @@
...
+                    "str_member": "foo",
+                },
+            ),
+            {
+                PageCloseBrowserEventModel: PageCloseBrowserEventModel.__selector__,
+                ServerEventModel: ServerEventModel.__selector__,
+            },
+        ),
+    ],
+)
```
jmaupetit, line 219: Smart usage of `type` in a test!

jmaupetit (approved):
That is a brilliant work! Keep it up.

Now you need to squash your commits and rebase your branch before merging it!

src/ralph/models/selector.py
SergioSim: You are right, yes indeed this will not scale well in future

PR #68, opened 2021-02-25
✨(tray) add support for ES cluster CA certs

jmaupetit:
## Purpose

When requesting an elasticsearch cluster using self-generated CA certificates, the elasticsearch client is raising a warning message as it cannot verify the authenticity of SSL certificates.

To mitigate the issue, one need to copy the CA certificate in ralph's container and specify the CA certificate path for the "es" backend using the `--es-client-options ca_certs=/path/to/ca.pem` option.

## Proposal

If the `ES_CA_CERTIFICATE` is defined in ralph's vault, a dedicated secret will be generated and mounted in the following path `/usr/local/share/ca-certificates/es-cluster.pem` so that execution logs are not polluted with warning messages.

### DIFF
--- a/src/tray/templates/services/app/dc.yml.j2
+++ b/src/tray/templates/services/app/dc.yml.j2
@@ -58,7 +58,16 @@ spec:
           volumeMounts:
             - name: ralph-v-history
               mountPath: /app/.ralph
+{% if ralph_mount_es_ca_secret %}
+            - name: es-ca-certificate
+              mountPath: /usr/local/share/ca-certificates/
+{% endif %}
       volumes:
         - name: ralph-v-history
           persistentVolumeClaim:
             claimName: ralph-pvc-history
+{% if ralph_mount_es_ca_secret %}
+        - name: es-ca-certificate
+          secret:
+            secretName: "{{ ralph_es_ca_secret_name }}"
+{% endif %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - EdX server event pydantic model and factory
 - EdX page_close browser event pydantic model and factory
+- Ralph tray now allows to specify a self-generated elasticsearch cluster CA
+  certificate
 
 ### Fixed
 
--- a/src/tray/templates/services/app/cronjob_pipeline.yml.j2
+++ b/src/tray/templates/services/app/cronjob_pipeline.yml.j2
@@ -42,9 +42,18 @@ items:
                   volumeMounts:
                     - name: ralph-v-history
                       mountPath: /app/.ralph
+{% if ralph_mount_es_ca_secret %}
+                    - name: es-ca-certificate
+                      mountPath: /usr/local/share/ca-certificates/
+{% endif %}
               volumes:
                 - name: ralph-v-history
                   persistentVolumeClaim:
                     claimName: ralph-pvc-history
+{% if ralph_mount_es_ca_secret %}
+                - name: es-ca-certificate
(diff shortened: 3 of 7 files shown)


quitterie-lcs (commented):
This seems good! I like your architecture idea of using `RALPH_VAULT` as a support

src/tray/templates/services/app/dc.yml.j2
```diff
@@ -58,7 +58,17 @@ spec:
           volumeMounts:
             - name: ralph-v-history
               mountPath: /app/.ralph
+{% set has_es_ca_certificate = lookup('k8s', kind='Secret', namespace=project_name, resource_name=ralph_es_ca_secret_name) | length > 0 %}
```
madmatah, line 61: I don't understand : why do you need to query the kubernetes API to check if the secret exist ?
According to `secret-es-ca.yml.j2`, the secret exist if `RALPH_VAULT.ES_CA_CERTIFICATE is defined`.

So, what is wrong, or what am I missing with the following suggestion ? 

```suggestion
{% set has_es_ca_certificate = RALPH_VAULT.ES_CA_CERTIFICATE is defined %}
```

I'm asking because i'm not comfortable with executing a query from a template. For the following reasons : 
 - We assume that k8s entities are created in a specific order by arnold, should we ?
 - Do we know precisely when this code is executed and how many time ? If this is executed during a template `lookup()`, each deployment template code is executed at least twice by arnold (once in `filter_empty_templates_for_app.yml` and once in `manage_app.yml`).

So if we can avoid it, i think it will be better.

jmaupetit: My first implementation was your suggestion, but I forgot that Arnold never loads vaults except when creating secrets (in the corresponding playbook). Hence, this cannot work.

> We assume that k8s entities are created in a specific order by arnold, should we ?

In this case, yes we should: secrets should be created first, prior to any deployment.

> Do we know precisely when this code is executed and how many time ?

We know it is executed at least twice:

- while looking for empty rendered templates, and,
- when creating the DC or the cronjob.

I also do think it's not an ideal situation, but I have no better alternative than:

- considering the CA certificate definition is mandatory, or,
- adding a configuration variable that will manually toggle secret creation and volume mounts.

I found both of them less elegant.

madmatah: > My first implementation was your suggestion, but I forgot that Arnold never loads vaults except when creating secrets (in the corresponding playbook). Hence, this cannot work.

Ok, I understand, thanks for the clarifications.

> > We assume that k8s entities are created in a specific order by arnold, should we ?
> 
> In this case, yes we should: secrets should be created first, prior to any deployment.

Just an extra comment on this point, I disagree.
It's kubernetes job to deal with this. The deployment will stay pending until the secret is created.

That's why we can do, for example, `./generate-k8s-config-from-templates | kubectl apply -f -`

> I also do think it's not an ideal situation, but I have no better alternative than:
> 
> * considering the CA certificate definition is mandatory, or,
> * adding a configuration variable that will manually toggle secret creation and volume mounts.
> 
> I found both of them less elegant.

I find the second proposition (+1 variable) better because it speeds-up deployments, avoid code execution inside templates and therefore follow the `generate-configuration | kubectl apply` philosophy. But there is a drawback for the tray operator, yes.

But it's just my opinion!

I approve this PR, I see nothing blocking.

madmatah (approved):

src/tray/templates/services/app/dc.yml.j2
jmaupetit: Ok, I'll follow your lead on this. Thanks!

PR #70, opened 2021-02-26
🐛(tray) remove version immutable field in selector

jmaupetit:
## Purpose

As Ralph is not a blue-green application, we cannot have a variable `version` field in the DC's label (it should be immutable).

Without this fix, one cannot deploy a new Ralph's release as k8s refuses to patch the object.

## Proposal

- [x] remove the `version` label and selector in the DC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


quitterie-lcs (commented):
Just to be sure I understand correctly, does it mean that we keep the information only in the `spec` data?

madmatah (approved):

PR #79, opened 2021-03-23
✨(models) add edx navigational event pydantic models

quitterie-lcs:
## Purpose

With the `validate` and `select` command newly implemented, it is now possible to implement edx tracking logs events pydantic models to validate incoming events in ralph. 

## Proposal

The first events implemented are the easiest (navigation).

- [x] event models
- [x] event fixtures
- [x] event tests

### DIFF
--- a/src/ralph/models/edx/base.py
+++ b/src/ralph/models/edx/base.py
@@ -9,7 +9,7 @@ from pydantic import BaseModel, HttpUrl, constr
 
 
 class BaseModelWithConfig(BaseModel):
-    """Base model defining configuration shared among all models"""
+    """Base model defining configuration shared among all models."""
 
     class Config:  # pylint: disable=missing-class-docstring
         extra = "forbid"
@@ -72,7 +72,7 @@ class AbstractBaseEventField(BaseModelWithConfig):
 class BaseEvent(BaseModelWithConfig):
     """Represents the base event model all events inherit from.
 
-    WARNING: it does not define the event, event_source and event_type fields.
+    WARNING: it does not define the `event`, `event_type` and `event_source` fields.
 
     Attributes:
         username (str): Consists of the unique username identifying the logged in user.
--- a/src/ralph/models/edx/navigational.py
+++ b/src/ralph/models/edx/navigational.py
@@ -1,0 +1,138 @@
+"""Navigational event model definitions"""
+
+from typing import Literal, Union
+
+from pydantic import Json, constr, validator
+
+from ralph.models.selector import selector
+
+from .base import AbstractBaseEventField
+from .browser import BaseBrowserEvent
+
+
+class UIPageClose(BaseBrowserEvent):
+    """Represents the `page_close` browser event.
+
+    This type of event is triggered when the user navigates to the next page
+    or closes the browser window (when the JavaScript `window.onunload` event
+    is called).
+
+    Attributes:
+        event (str): Consists of the string value `{}`.
+        event_type (str): Consists of the value `page_close`.
+        name (str): Consists of the value `page_close`.
+    """
+
+    __selector__ = selector(event_source="browser", event_type="page_close")
+
+    # pylint: disable=unsubscriptable-object
+    event: Literal["{}"]
+    event_type: Literal["page_close"]
+    name: Literal["page_close"]
+
+
+class NavigationalEventField(AbstractBaseEventField):
+    """Represents the event field of navigational events.
+
+    Note: All navigational events are `browser` events.
+
+    Attributes:
+        id (str): Consists of the edX ID of the sequence.
+        old (int): For `seq_goto`, it consists of the index of the unit being jumped to.
+            For `seq_next` and `seq_prev`, it consists of the index of the unit being navigated to.
+        new (int): For `seq_goto`, it consists of the index of the unit being jumped from.
+            For `seq_next` and `seq_prev`, it consists of the index of the unit being navigated
+            away from.
+    """
+
+    id: constr(
+        regex=(
+            r"^block-v1:[^\/+]+(\/|\+)[^\/+]+(\/|\+)[^\/?]+type"  # noqa : F722
+            r"@sequential\+block@[a-f0-9]{32}$"  # noqa : F722
+        )
+    )
+    new: int
+    old: int
+
+
+class UISeqGoto(BaseBrowserEvent):
+    """Represents the `seq_goto` browser event.
+
+    The browser emits such event when a user selects a navigational control.
+    `seq_goto` is emitted when a user jumps between units in a sequence.
+
+    Attributes:
+        event (obj): Consists of member fields that identify specifics triggered event.
+        event_type (str): Consists of the value `seq_goto`.
+        name (str): Consists of the value `seq_goto`.
+    """
+
+    __selector__ = selector(event_source="browser", event_type="seq_goto")
+
+    # pylint: disable=unsubscriptable-object
+    event: Union[Json[NavigationalEventField], NavigationalEventField]
+    event_type: Literal["seq_goto"]
+    name: Literal["seq_goto"]
+
+
+class UISeqNext(BaseBrowserEvent):
+    """Represents the `seq_next` browser event.
+
+    The browser emits such event when a user selects a navigational control.
+    `seq_next` is emitted when a user navigates to the next unit in a sequence.
+
+    Attributes:
+        event (obj): Consists of member fields that identify specifics triggered event.
+        event_type (str): Consists of the value `seq_next`.
+        name (str): Consists of the value `seq_next`.
+    """
+
+    __selector__ = selector(event_source="browser", event_type="seq_next")
+
+    # pylint: disable=unsubscriptable-object
+    event: Union[Json[NavigationalEventField], NavigationalEventField]
+    event_type: Literal["seq_next"]
+    name: Literal["seq_next"]
+
+    @validator("event")
+    def validate_next_jump_event_field(
+        cls, value
+    ):  # pylint: disable=no-self-argument, no-self-use
+        """Checks that event.new is equal to event.old + 1."""
+
+        if value.new != value.old + 1:
+            raise ValueError("event.new - event.old should be equal to 1")
+
+        return value
+
+
+class UISeqPrev(BaseBrowserEvent):
+    """Represents the `seq_prev` browser event.
+
+    The browser emits such event when a user selects a navigational control.
+    `seq_prev` is emitted when a user navigates to the previous unit in a sequence.
+
+    Attributes:
+        event (obj): Consists of member fields that identify specifics triggered event.
+        event_type (str): Consists of the value `seq_prev`.
+        name (str): Consists of the value `seq_prev`.
+
+    """
+
+    __selector__ = selector(event_source="browser", event_type="seq_prev")
+
+    # pylint: disable=unsubscriptable-object
+    event: Union[Json[NavigationalEventField], NavigationalEventField]
+    event_type: Literal["seq_prev"]
+    name: Literal["seq_prev"]
+
+    @validator("event")
+    def validate_prev_jump_event_field(
+        cls, value
+    ):  # pylint: disable=no-self-argument, no-self-use
+        """Checks that event.new is equal to event.old - 1."""
+
+        if value.new != value.old - 1:
+            raise ValueError("event.old - event.new should be equal to 1")
+
+        return value
--- a/src/ralph/models/edx/server.py
+++ b/src/ralph/models/edx/server.py
@@ -46,8 +46,6 @@ class ServerEvent(BaseServerEvent):
         event_source="server", event_type=LazyModelField("context__path")
     )
 
+    # pylint: disable=unsubscriptable-object
     event_type: Path
-    event: Union[
-        Json[ServerEventField],  # pylint: disable=unsubscriptable-object
-        ServerEventField,
-    ]
+    event: Union[Json[ServerEventField], ServerEventField]
--- a/src/ralph/models/selector.py
+++ b/src/ralph/models/selector.py
@@ -30,7 +30,7 @@ class Rule:
     """Represents a rule used for selection."""
 
     field: LazyModelField
-    value: Union[LazyModelField, Any]
+    value: Union[LazyModelField, Any]  # pylint: disable=unsubscriptable-object
 
     def check(self, event):
         """Checks if event matches the rule.
--- a/tests/models/edx/test_browser.py
+++ b/tests/models/edx/test_browser.py
@@ -8,9 +8,7 @@ from hypothesis import given, provisional, settings
 from hypothesis import strategies as st
 from pydantic.error_wrappers import ValidationError
 
-from ralph.exceptions import UnknownEventException
-from ralph.models.edx.browser import BaseBrowserEvent, PageClose
-from ralph.models.selector import ModelSelector
+from ralph.models.edx.browser import BaseBrowserEvent
 
 
 @settings(max_examples=1)
@@ -44,19 +42,3 @@ def test_models_edx_browser_base_browser_event_with_invalid_content(
 
     with pytest.raises(ValidationError, match=error):
         BaseBrowserEvent(**invalid_event)
-
-
-@settings(max_examples=1)
-@given(st.builds(PageClose, referer=provisional.urls(), page=provisional.urls()))
-def test_models_selector_model_selector_get_model_with_valid_event(event):
-    """Tests given a page_close event the get_model method should return PageClose model."""
-
-    event = json.loads(event.json())
-    assert ModelSelector(module="ralph.models.edx").get_model(event) is PageClose
-
-
-def test_models_selector_model_selector_get_model_with_invalid_event():
-    """Tests given a page_close event the get_model method should raise UnknownEventException."""
-
-    with pytest.raises(UnknownEventException):
-        ModelSelector(module="ralph.models.edx").get_model({"invalid": "event"})
--- a/tests/models/edx/test_navigational.py
+++ b/tests/models/edx/test_navigational.py
@@ -1,0 +1,173 @@
+"""Tests for the navigational event models"""
+
+import json
+import re
+
+import pytest
+from hypothesis import given, provisional, settings
+from hypothesis import strategies as st
+from pydantic.error_wrappers import ValidationError
+
+from ralph.models.edx.base import BaseContextField
+from ralph.models.edx.navigational import (
+    NavigationalEventField,
+    UIPageClose,
+    UISeqGoto,
+    UISeqNext,
+    UISeqPrev,
+)
+from ralph.models.selector import ModelSelector
+
+
+@settings(max_examples=1)
+@given(st.builds(UIPageClose, referer=provisional.urls(), page=provisional.urls()))
+def test_models_edx_ui_page_close_selector_with_valid_event(event):
+    """Tests given a page_close event the get_model method should return UIPageClose model."""
+
+    event = json.loads(event.json())
+    assert ModelSelector(module="ralph.models.edx").get_model(event) is UIPageClose
+
+
+@settings(max_examples=1)
+@given(st.builds(NavigationalEventField))
+def test_fields_edx_navigational_events_event_field_with_valid_content(field):
+    """Tests that a valid NavigationalEventField field does not raise a ValidationError."""
+
+    assert re.match(
+        r"^block-v1:[^\/+]+(\/|\+)[^\/+]+(\/|\+)[^\/?]+type@sequential\+block@[a-f0-9]{32}$",
+        field.id,
+    )
+
+
+@pytest.mark.parametrize(
+    "id",
+    [
+        "block-v2:orgX=CS111+20_T1+type@sequential+block@d0d4a647742943e3951b45d9db8a0ea1",
+        "block-v1:orgX=CS11120_T1+type@sequential+block@d0d4a647742943e3951b45d9db8a0ea1",
+        "block-v1:orgX=CS111=20_T1+tipe@sequential+block@d0d4a647742943e3951b45d9db8a0ea1",
+        "block-v1:orgX=CS111=20_T1+",
+        "type@sequentialblock@d0d4a647742943e3951b45d9db8a0ea1",
+        "block-v1:orgX=CS111=20_T1+type@sequential+block@d0d4a647742943z3951b45d9db8a0ea1",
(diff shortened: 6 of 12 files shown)


src/ralph/models/edx/navigational.py
```diff
@@ -0,0 +1,130 @@
...
+    Attributes:
+        id (int): Consists of the edX ID of the sequence.
+        old (str): For `seq_goto`, it consists of the index of the unit being jumped to.
+        For `seq_next` and `seq_prev`, it consists of the index of the unit being navigated to.
+        new (str): For `seq_goto`, it consists of the index of the unit being jumped from.
+        For `seq_next` and `seq_prev`, it consists of the index of the unit being navigated
+        away from.
+    """
+
+    identifier: str # FIXME with string regex
```
SergioSim, line 47:
```suggestion
    id: str # FIXME with string regex
```
I think it's `id` ?

src/ralph/models/edx/navigational.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Attributes:
+        event (obj): Consists of member fields that identify specifics triggered event
+        event_type (str): Consists of the value `seq_goto`
+        name (str): Consists of the value `seq_goto`
+    """
+
+    __selector__ = selector(event_source="browser", event_type="seq_goto")
+
+    event: NavigationalBrowserEventEventFieldModel
```
SergioSim, line 66: Aye, sorry, there is an error in my documentation claiming that the event field is a dict, it's actually a JSON string containing a dict. I'm really sorry for this mistake :disappointed:

SergioSim:
```suggestion
    event: Union[Json[NavigationalBrowserEventEventFieldModel], NavigationalBrowserEventEventFieldModel]
```
I was hesitating about using `Json` as it changes the output (converts JSON string to dict) but now I think it might be beneficial:
- We can avoid the validation error when validating the event a second time using `Union` (accepting both string + dict)
- It simplifies the conversion step  (we would not have to re-parse the JSON string part)

src/ralph/models/edx/server.py
```diff
@@ -63,3 +71,9 @@ def validate_event_type(
         if values.get("event_type") != values.get("context").path:
             raise ValueError("event_type should be equal to context.path")
         return values
+
+
+class BaseServerEventModel(BaseEventModel):
+    """Represents the base server event model all server events inherit from."""
+
+    event_source: Literal["server"]
```
SergioSim, line 79: Good Idea) Then we don't need to define the `event_source` in the `BaseEventModel` ? Or maybe we could set it to an invalid value like `None` ? And make the `ServerEventModel` inherit the `BaseServerEventModel` ?

src/ralph/models/edx/server.py
```diff
@@ -2,12 +2,20 @@
...
 
-from .base import BaseEventModel
+from .base import BaseEventModel, BaseModelWithConfig
+
+
+class ServerEventField(BaseModelWithConfig):
+    """Represents the `event` field of the ServerEventModel."""
+
+    GET: dict
+    POST: dict
```
SergioSim, line 18:
```suggestion
from .base import BaseEventModel
```

This part was removed in the previous commit (validate command)

tests/fixtures/edx/navigational.py
```diff
@@ -0,0 +1,95 @@
...
+    SeqGotoBrowserEventModel,
+    SeqNextBrowserEventModel,
+    SeqPrevBrowserEventModel,
+)
+
+from .base import BaseEventFieldFactory
+from .browser import BaseBrowserEventFactory
+
+Faker.seed(1)
+FAKE = Faker()
```
SergioSim, line 18:
```suggestion
from .base import FAKE, BaseEventFieldFactory
from .browser import BaseBrowserEventFactory
```
FAKE is already created in `.base`, we could reuse it ?

tests/models/edx/test_navigational.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@pytest.mark.parametrize(
+    "kwargs,error",
+    [
+        ({"name": None}, "name\n  none is not an allowed value"),
+        ({"name": "prev"}, "unexpected value; permitted: 'seq_prev'"),
+        ({"event_type": None}, "event_type\n  none is not an allowed value"),
+        ({"event_type": "prev"}, "unexpected value; permitted: 'seq_prev'"),
+        ({"event": None}, "event\n  none is not an allowed value"),
```
SergioSim, line 86: Could you add some test cases covering the `@validator`s too (validate_next_jump_event_field / validate_prev_jump_event_field)
Example: 
```suggestion
        ({"event": None}, "event\n  none is not an allowed value"),
        ({"event__new": 10, "event__old": 0}, "event.new should be equal to event.old + 1"),
```

SergioSim (commented):
This looks Great! Thank you for your contribution!

src/ralph/models/edx/base.py
```diff
@@ -16,6 +17,22 @@ class Config:  # pylint: disable=missing-class-docstring
...
+
+class EventMode(StrEnum):
+    """Event field modes."""
+
+    AUDIT = "audit"
+    HONOR = "honor"
+    PROFESSIONAL = "professional"
+    VERIFIED = "verified"
+
+
```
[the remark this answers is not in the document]
quitterie-lcs: You're right, it's something I forgot to remove with enrollment events cleaning :+1:

src/ralph/models/edx/server.py
quitterie-lcs: Oops I wrongly fixed the conflict... Change it now

tests/fixtures/edx/navigational.py
quitterie-lcs: You're right, thanks! I did not think of reusing it

tests/models/edx/test_navigational.py
quitterie-lcs: Good idea! Thanks ;)

src/ralph/models/edx/server.py
quitterie-lcs: Yes, I let the server default value at first but as discussed IRL, it would be more logical to overload server value when defining `BaseServerEventModel`

quitterie-lcs: To keep homogeneity, I would prefer to remove it from the base model, as we do not define event_type to None in this model. WDYT?

SergioSim: Sounds good for me) you could then update the warning message in the doc string of `BaseEventModel`, mentioning that `event_source` is not defined too)

src/ralph/models/edx/base.py
```diff
@@ -76,6 +93,13 @@ def validate_course_id(
         return values
 
 
+class BaseEventFieldModel(BaseModelWithConfig):
+    """Represents the base event model inherited by specific event event fiels.
+
+    The base model does not have any attributes as event field does not have common sub-fields.
+    """
```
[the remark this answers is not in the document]
quitterie-lcs: I am not sure it is possible, or we could forbid the import of BaseModelWith Config in other files than the server and browser ones. But that may be dangerous

quitterie-lcs:
@SergioSim it is ready for a new review! :wink:

SergioSim (commented):
Only some questions left for me) :+1:

src/ralph/models/edx/server.py
quitterie-lcs: You're right, I forgot to pay attention it 🤦

src/ralph/models/edx/navigational.py
```diff
@@ -0,0 +1,140 @@
...
+    Attributes:
+        id (int): Consists of the edX ID of the sequence.
+        old (str): For `seq_goto`, it consists of the index of the unit being jumped to.
+            For `seq_next` and `seq_prev`, it consists of the index of the unit being navigated to.
+        new (str): For `seq_goto`, it consists of the index of the unit being jumped from.
+            For `seq_next` and `seq_prev`, it consists of the index of the unit being navigated
+        away from.
+    """
+
+    id: str  # FIXME with string regex
```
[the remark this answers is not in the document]
quitterie-lcs: That's a good idea. I have tried to implement like in your marshmallow implementation so that we can have also course_key and block_id in the instance. Tests are still failing but it should be ready this afternoon

src/ralph/models/selector.py
```diff
@@ -30,7 +30,7 @@ class Rule:
     """Represents a rule used for selection."""
 
     field: LazyModelField
-    value: Union[LazyModelField, Any]
+    value: Union[LazyModelField, Any]  # pylint: disable=unsubscriptable-object
```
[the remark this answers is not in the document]
quitterie-lcs: The linter raised this error actually, that is why I made this modification

jmaupetit (commented):
After reading your work, I have 4 remarks/suggestions in mind:

1. Why not explicitly naming base models that cannot be used directly as `AbstractXXX`?
2. Can we remove the `Model` suffix from models? For example the `PageCloseBrowserEventModel` becomes `PageClose`?
3. Can we separate fields from models in a dedicated module (_e.g._ `fields.py`)?
4. As Pydantic is fully integrated with [`hypothesis`](https://hypothesis.readthedocs.io), shouldn't we remove our factories (and the FactoryBoy dependency) and rewrite our tests using it?

IMO all of this may be addressed in a separated PR that must be evaluated before this proposal that will need to be refactored a bit.

quitterie-lcs:
> After reading your work, I have 4 remarks/suggestions in mind:
> 
> 1. Why not explicitly naming base models that cannot be used directly as `AbstractXXX`?
> 2. Can we remove the `Model` suffix from models? For example the `PageCloseBrowserEventModel` becomes `PageClose`?
> 3. Can we separate fields from models in a dedicated module (_e.g._ `fields.py`)?
> 4. As Pydantic is fully integrated with [`hypothesis`](https://hypothesis.readthedocs.io), shouldn't we remove our factories (and the FactoryBoy dependency) and rewrite our tests using it?
> 
> IMO all of this may be addressed in a separated PR that must be evaluated before this proposal that will need to be refactored a bit.

Modifications have been addressed in #82

SergioSim:
> After reading your work, I have 4 remarks/suggestions in mind:
> 
>     1. Why not explicitly naming base models that cannot be used directly as `AbstractXXX`?
> 
>     2. Can we remove the `Model` suffix from models? For example the `PageCloseBrowserEventModel` becomes `PageClose`?
> 
>     3. Can we separate fields from models in a dedicated module (_e.g._ `fields.py`)?
> 
>     4. As Pydantic is fully integrated with [`hypothesis`](https://hypothesis.readthedocs.io), shouldn't we remove our factories (and the FactoryBoy dependency) and rewrite our tests using it?
> 
> 
> IMO all of this may be addressed in a separated PR that must be evaluated before this proposal that will need to be refactored a bit.

I have some questions regrading your suggestion:
1. By removing the suffix `Browser` / `Server` from the models, how will we distinguish between a `problem_check` server event and a `problem_check` browser event in future ?
2. `hypothesis` is a great tool to test that our models behave the way we have defined them, but I think we are missing tests checking that our Models correspond / match actual edX tracking events. I suggest to add static test fixtures of edX events in addition to `hypothesis` tests. WDYT ?

jmaupetit:
> By removing the suffix Browser / Server from the models, how will we distinguish between a problem_check server event and a problem_check browser event in future ?

Can you paste examples of `problem_check` browser/server events please?

> hypothesis is a great tool to test that our models behave the way we have defined them, but I think we are missing tests checking that our Models correspond / match actual edX tracking events. I suggest to add static test fixtures of edX events in addition to hypothesis tests. WDYT ?

I would like more functional tests. Can you provide an example of static test fixtures you would like to implement?

SergioSim:
> > By removing the suffix Browser / Server from the models, how will we distinguish between a problem_check server event and a problem_check browser event in future ?
> 
> Can you paste examples of `problem_check` browser/server events please?
> 
> > hypothesis is a great tool to test that our models behave the way we have defined them, but I think we are missing tests checking that our Models correspond / match actual edX tracking events. I suggest to add static test fixtures of edX events in addition to hypothesis tests. WDYT ?
> 
> I would like more functional tests. Can you provide an example of static test fixtures you would like to implement?

Here is an example of a Browser problem_check event:

        {
            "username": "sergios",
            "event_source": "browser",
            "name": "problem_check",
            "accept_language": "en-US,en;q=0.5",
            "time": "2020-04-18T14:29:27.162483+00:00",
            "agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0",
            "page": "http://localhost:8072/courses/course-v1:organisationCourse+numeroCours+sessiondCours/courseware/b5722732abed4eb7ba823f0b4239d50d/e2fa42fa1ca645389ee4618c86e982ff/",
            "host": "7a4a1c08e030",
            "session": "cc7ee04f8e4eb7e84f8c4c441cc78f40",
            "referer": "http://localhost:8072/courses/course-v1:organisationCourse+numeroCours+sessiondCours/courseware/b5722732abed4eb7ba823f0b4239d50d/e2fa42fa1ca645389ee4618c86e982ff/",
            "context": {
                "user_id": 2,
                "org_id": "organisationCourse",
                "course_id": "course-v1:organisationCourse+numeroCours+sessiondCours",
                "path": "/event"
            },
            "ip": "172.20.0.1",
            "event": "input_e8bc718966e9441abc3ccf1a6429ee8b_2_1=choice_2",
            "event_type": "problem_check"
        }


Here is an example of a Server problem_check event:

        {
            "username": "sergios",
            "event_type": "problem_check",
            "ip": "172.20.0.1",
            "agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0",
            "host": "7a4a1c08e030",
            "referer": "http://localhost:8072/courses/course-v1:organisationCourse+numeroCours+sessiondCours/courseware/b5722732abed4eb7ba823f0b4239d50d/e2fa42fa1ca645389ee4618c86e982ff/",
            "accept_language": "",
            "event": {
                "submission": {
                    "e8bc718966e9441abc3ccf1a6429ee8b_2_1": {
                        "input_type": "choicegroup",
                        "question": "Which of the following countries has the largest population?",
                        "response_type": "multiplechoiceresponse",
                        "answer": "Indonesia",
                        "variant": "",
                        "correct": true
                    }
                },
                "success": "correct",
                "grade": 1,
                "correct_map": {
                    "e8bc718966e9441abc3ccf1a6429ee8b_2_1": {
                        "hint": "",
                        "hintmode": null,
                        "correctness": "correct",
                        "npoints": null,
                        "answervariable": null,
                        "msg": "",
                        "queuestate": null
                    }
                },
                "state": {
                    "student_answers": {},
                    "seed": 1,
                    "done": true,
                    "correct_map": {
                        "e8bc718966e9441abc3ccf1a6429ee8b_2_1": {
                            "hint": "",
                            "hintmode": null,
                            "correctness": "incorrect",
                            "npoints": null,
                            "answervariable": null,
                            "msg": "",
                            "queuestate": null
                        }
                    },
                    "input_state": {
                        "e8bc718966e9441abc3ccf1a6429ee8b_2_1": {}
                    }
                },
                "answers": {
                    "e8bc718966e9441abc3ccf1a6429ee8b_2_1": "choice_2"
                },
                "attempts": 2,
                "max_grade": 1,
                "problem_id": "block-v1:organisationCourse+numeroCours+sessiondCours+type@problem+block@e8bc718966e9441abc3ccf1a6429ee8b"
            },
            "event_source": "server",
            "context": {
                "course_user_tags": {},
                "user_id": 2,
                "org_id": "organisationCourse",
                "module": {
                    "usage_key": "block-v1:organisationCourse+numeroCours+sessiondCours+type@problem+block@e8bc718966e9441abc3ccf1a6429ee8b",
                    "display_name": "Nom Affichage Multiple Choice"
                },
                "course_id": "course-v1:organisationCourse+numeroCours+sessiondCours",
                "path": "/courses/course-v1:organisationCourse+numeroCours+sessiondCours/xblock/block-v1:organisationCourse+numeroCours+sessiondCours+type@problem+block@e8bc718966e9441abc3ccf1a6429ee8b/handler/xmodule_handler/problem_check"
            },
            "time": "2020-04-18T14:34:00.266685+00:00",
            "page": "x_module"
        }

This two events also represent an example of static test cases I would like to use to check if our problem_check event models correspond / match the actual edX problem_check events.

jmaupetit:
As discussed, browser events can be `UI`-prefixed as reflecting an interaction with the user interface while server events won't have any prefix or suffix, _e.g._ `UIPageClose` or `UIProblemCheck` and `ProblemCheck` for the server event.

src/ralph/models/edx/base.py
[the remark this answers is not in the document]
quitterie-lcs: Resolved in #82

SergioSim (commented):
This looks beautiful! Only some minor questions/suggestions left for me) :+1:

src/ralph/models/edx/navigational.py
```diff
@@ -0,0 +1,144 @@
...
+from pydantic import Json, constr, validator
+
+from ralph.models.selector import selector
+
+from .base import AbstractBaseEventField
+from .browser import BaseBrowserEvent
+
+
+class PageClose(BaseBrowserEvent):
+    """Represents the `page_close` browser event.
```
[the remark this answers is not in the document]
quitterie-lcs: That's not a bad idea, indeed. Let's wait also for @jmaupetit point of view about it

tests/fixtures/edx/navigational.py
```diff
@@ -0,0 +1,100 @@
+"""Navigational event factory definitions"""
```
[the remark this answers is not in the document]
quitterie-lcs: You're right, an error in the rebase I made

tests/models/edx/test_browser.py
```diff
@@ -9,7 +9,8 @@
 from pydantic.error_wrappers import ValidationError
 
 from ralph.exceptions import UnknownEventException
-from ralph.models.edx.browser import BaseBrowserEvent, PageClose
+from ralph.models.edx.browser import BaseBrowserEvent
+from ralph.models.edx.navigational import PageClose
```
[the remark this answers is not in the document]
quitterie-lcs: Ooops, it's something I forgot to move...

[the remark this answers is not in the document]
quitterie-lcs: I prefer `test_models_edx_ui_page_close_selector_with_valid_event` (only the event name) because with enrollment events for example, which names are very long we would exceed the column character limits and it would be difficult to read

tests/models/edx/test_navigational.py
```diff
@@ -0,0 +1,152 @@
...
+            "0",
+            "event\n  event.old - event.new should be equal to 1",
+        ),
+    ],
+)
+@settings(max_examples=1)
+@given(
+    st.builds(
+        UISeqPrev,
+        context=st.builds(BaseContextField),
```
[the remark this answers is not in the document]
quitterie-lcs: As it is a custom made class, we have to specify hypothesis to build it

quitterie-lcs:
The current failing test is the result of a different behaviour, when it is written as a static error message in test_fields_edx_navigational_events_event_field_with_invalid_content, it only describes the fact that str type is expected and did not care about the regex not respected. I think it is due to the `id` which is conflictual. I propose to let the error message as a test parameter

tests/models/edx/test_navigational.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+
+
+@pytest.mark.parametrize(
+    "id",
+    [
+        (
+            "block-v2:universityX=CS111+2020_T1+"
+            "type@sequential+block@d0d4a647742943e3951b45d9db8a0ea1",
+        ),
```
SergioSim, line 48: Sorry my fault, it seems python interprets this as a tuple, but without the coma it's a string and the test works as before)
```suggestion
        (
            "block-v2:universityX=CS111+2020_T1+"
            "type@sequential+block@d0d4a647742943e3951b45d9db8a0ea1"
        ),
```

tests/models/edx/test_navigational.py
```diff
@@ -0,0 +1,190 @@
...
+        (
+            "block-v1:universityX=CS111=2020_T1+"
+            "type@sequential+block@d0d4a647742943z3951b45d9db8a0ea1",
+        ),
+        (
+            "block-v1:universityX=CS111=2020_T1+"
+            "type@sequential+block@d0d4a647742943e3951b45d9db8a0ea13",
+        ),
+    ],  # pylint: disable=invalid-name
+)
```
SergioSim, line 70: Optionally, by shortening the names it's possible to fit each test case in one line:
Exemple:
```suggestion
@pytest.mark.parametrize(
    "id",
    [
        "block-v2:orgX=CS111+20_T1+type@sequential+block@d0d4a647742943e3951b45d9db8a0ea1",
        "block-v1:orgX=CS11120_T1+type@sequential+block@d0d4a647742943e3951b45d9db8a0ea1",
        "block-v1:orgX=CS111=20_T1+tipe@sequential+block@d0d4a647742943e3951b45d9db8a0ea1",
        "block-v1:orgX=CS111=20_T1+",
        "type@sequentialblock@d0d4a647742943e3951b45d9db8a0ea1",
        "block-v1:orgX=CS111=20_T1+type@sequential+block@d0d4a647742943z3951b45d9db8a0ea1",
        "block-v1:orgX=CS111=20_T1+type@sequential+block@d0d4a647742943e3951b45d9db8a0ea13",
    ],  # pylint: disable=invalid-name
)
```

tests/models/edx/test_navigational.py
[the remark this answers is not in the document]
SergioSim: Good to know) Thanks)

tests/models/edx/test_navigational.py
quitterie-lcs: Ah yes! I did not notice that too, I will fix this and see if it works :wink:

quitterie-lcs:
> The current failing test is the result of a different behaviour, when it is written as a static error message in `test_fields_edx_navigational_events_event_field_with_invalid_content`, it only describes the fact that `str` type is expected and did not care about the `regex` not respected. I think it is due to the `id` which is conflictual. We can change the test message to have a valid test but that does not test really what we want. I propose to let the error message as a test parameter to get regex inconsistency test. WDYT?

Resolved with shorter strings

quitterie-lcs:
@SergioSim it has been fixed!

SergioSim (commented):
This is Great :+1:  ! Only one last minor suggestion left for me) @jmaupetit are you available for a review ?

SergioSim (approved):
image

PR #80, opened 2021-03-24
✨(models) add edX server event to xAPI converter

SergioSim:
## Purpose

EdX server events are common events in the edX tracking logs.
We want to convert them to xAPI format.

## Proposal

`models/converter.py` defines the base converter.
`models/edx/converters/xapi/base.py` defines the base converter for edX to xAPI conversion.
`models/edx/converters/xapi/server.py` converts edX server events to xAPI format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to
 
 ### Added
 
+- EdX to xAPI converters for page `viewed` and`page_close` events
+- Implement core event format converter
 - xAPI video `played` pydantic models
 - xAPI page `viewed` and page `terminated` pydantic models
 - Implement edx navigational events pydantic models
--- a/src/ralph/models/converter.py
+++ b/src/ralph/models/converter.py
@@ -1,0 +1,137 @@
+"""Converter methods definition"""
+
+import json
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Callable, Union
+
+from pydantic import BaseModel
+
+from ralph.defaults import MODEL_PATH_SEPARATOR
+from ralph.exceptions import BadFormatException, ConversionException
+from ralph.utils import get_dict_value_from_path, set_dict_value_from_path
+
+
+@dataclass(frozen=True)
+class ConversionItem:
+    """Represents a conversion set item."""
+
+    dest: tuple[str]
+    src: Union[tuple[str], str, None]
+    transformers: tuple[Callable[[Any], Any]]
+    raw_input: bool
+
+    def __init__(self, dest: str, src=None, transformers=lambda _: _, raw_input=False):
+        """Initializes ConversionItem.
+
+        Args:
+            dest (str): The destination path where to place the converted value.
+            src (str or None): The source from where the value to convert is fetched.
+                - When `src` is a path (ex. `context__user_id`) - the value is the item
+                    of the event at the path.
+                - When `src` is `None` - the value is the whole event.
+            transformers (function or tuple of functions): The function(s) to apply on the source
+                value.
+            raw_input (bool): Flag indicating whether `get_value` will receive a raw event string
+                or a parsed event dictionary.
+        """
+
+        object.__setattr__(self, "dest", tuple(dest.split(MODEL_PATH_SEPARATOR)))
+        src = tuple(src.split(MODEL_PATH_SEPARATOR)) if src else None
+        object.__setattr__(self, "src", src)
+        transformers = tuple([transformers]) if callable(transformers) else transformers
+        object.__setattr__(self, "transformers", transformers)
+        object.__setattr__(self, "raw_input", raw_input)
+
+    def get_value(self, data: Union[dict, str]):
+        """Returns fetched source value after having applied all transformers to it.
+
+        Args:
+            data (dict or string): The event to convert.
+
+        Raises:
+            ConversionException: When a field transformation fails.
+        """
+
+        if self.src:
+            data = get_dict_value_from_path(data, self.src)
+        try:
+            for transformer in self.transformers:
+                data = transformer(data)
+        except Exception as err:
+            msg = f"Failed to get the transformed value for field: {self.src}"
+            raise ConversionException(msg) from err
+        return data
+
+
+class BaseConversionSet(ABC):
+    """ConversionSet Base Class.
+
+    ConversionSets are used to convert from one event format to another.
+    """
+
+    __src__: BaseModel
+    __dest__: BaseModel
+
+    def __init__(self):
+        """Initializes BaseConversionSet."""
+
+        self._conversion_items = self._get_conversion_items()
+
+    @abstractmethod
+    def _get_conversion_items(self) -> set[ConversionItem]:
+        """Returns a set of ConversionItems used for conversion."""
+
+    def __iter__(self):
+        return iter(self._conversion_items)
+
+
+def convert_dict_event(
+    event: dict, event_str: str, conversion_set: BaseConversionSet
+) -> BaseModel:
+    """Converts the event dictionary using the provided original event string and conversion_set.
+
+    Args:
+        event (dict): The event to convert.
+        event_str (dict): The original event string.
+        conversion_set (BaseConversionSet): A conversion set used for conversion.
+
+    Returns:
+        event (BaseModel): The converted event pydantic model.
+
+    Raises:
+        ConversionException: When a field transformation fails.
+        ValidationError: When the final converted event is invalid.
+    """
+
+    converted_event = {}
+    for conversion_item in conversion_set:
+        data = event_str if conversion_item.raw_input else event
+        value = conversion_item.get_value(data)
+        if value not in [None, "", {}]:
+            set_dict_value_from_path(converted_event, conversion_item.dest, value)
+    return conversion_set.__dest__(**converted_event)
+
+
+def convert_str_event(event_str: str, conversion_set: BaseConversionSet) -> BaseModel:
+    """Converts the event string using the provided conversion_set.
+
+    Args:
+        event_str (str): The event to convert.
+        conversion_set (BaseConversionSet): A conversion set used for conversion.
+
+    Returns:
+        event (BaseModel): The converted event pydantic model.
+
+    Raises:
+        BadFormatException: When the event_str is not a valid JSON string.
+        ConversionException: When a field transformation fails.
+        ValidationError: When the converted event is invalid.
+    """
+
+    try:
+        event = json.loads(event_str)
+    except (TypeError, json.JSONDecodeError) as err:
+        msg = "Failed to parse the event, invalid JSON string"
+        raise BadFormatException(msg) from err
+    return convert_dict_event(event, event_str, conversion_set)
--- a/src/ralph/models/edx/converters/xapi/base.py
+++ b/src/ralph/models/edx/converters/xapi/base.py
@@ -1,0 +1,79 @@
+"""Base xAPI Converter"""
+
+import re
+from uuid import UUID, uuid5
+
+from ralph.exceptions import ConfigurationException
+from ralph.models.converter import BaseConversionSet, ConversionItem
+from ralph.models.xapi.constants import (
+    EXTENSION_COURSE_ID,
+    EXTENSION_MODULE_ID,
+    EXTENSION_SCHOOL_ID,
+)
+
+
+class BaseXapiConverter(BaseConversionSet):
+    """Base xAPI Converter.
+
+    WARNING: The converter may not include the following edX fields:
+    - context.org_id: When `org_id` is an empty string.
+    - context.course_id: When `course_id` is an empty string.
+
+    WARNING: The converter should not include the following edX fields as they may
+    contain sensitive data: `username`, `referer`, `event`, `event_source`, `ip`,
+    `agent`, `accept_language:`, `context.course_user_tags`.
+    """
+
+    def __init__(self, uuid_namespace: str, home_page: str):
+        """Initializes BaseXapiConverter."""
+
+        self.home_page = home_page
+        try:
+            self.uuid_namespace = UUID(uuid_namespace)
+        except (TypeError, ValueError, AttributeError) as err:
+            raise ConfigurationException("Invalid UUID namespace") from err
+        super().__init__()
+
+    def _get_conversion_items(self):
+        """Returns a set of ConversionItems used for conversion."""
+
+        return {
+            ConversionItem(
+                "id",
+                None,
+                lambda event_str: uuid5(self.uuid_namespace, event_str),
+                True,
+            ),
+            ConversionItem(
+                "actor__account__homePage", transformers=lambda _: self.home_page
+            ),
+            ConversionItem(
+                "actor__account__name",
+                "context__user_id",
+                lambda user_id: user_id if user_id else "anonymous",
+            ),
+            ConversionItem(
+                "object__definition__extensions__" + EXTENSION_SCHOOL_ID,
+                "context__org_id",
+            ),
+            ConversionItem(
+                "object__definition__extensions__" + EXTENSION_COURSE_ID,
+                "context__course_id",
+                (self.parse_course_id, lambda x: x["course"]),
+            ),
+            ConversionItem(
+                "object__definition__extensions__" + EXTENSION_MODULE_ID,
+                "context__course_id",
+                (self.parse_course_id, lambda x: x["module"]),
+            ),
+            ConversionItem("timestamp", "time"),
+        }
+
+    @staticmethod
+    def parse_course_id(course_id: str):
+        """Returns a dictionary with `course` and `module` of edX event's `context.course_id`."""
+
+        match = re.match(r"^course-v1:.+\+(.+)\+(.+)$", course_id)
+        if not match:
+            return {"course": None, "module": None}
+        return {"course": match.group(1), "module": match.group(2)}
--- a/src/ralph/models/edx/converters/xapi/server.py
+++ b/src/ralph/models/edx/converters/xapi/server.py
@@ -1,0 +1,31 @@
+"""Server event xAPI Converter"""
+
+from ralph.models.converter import ConversionItem
+from ralph.models.edx.server import ServerEvent
+from ralph.models.xapi.navigation.statements import PageViewed
+
+from .base import BaseXapiConverter
+
+
+class ServerEventToPageViewed(BaseXapiConverter):
+    """Converts a common edX server event to xAPI.
+
+    Example Statement: John viewed https://www.fun-mooc.fr/ page.
+    """
+
+    __src__ = ServerEvent
+    __dest__ = PageViewed
+
+    def _get_conversion_items(self):
+        """Returns a set of ConversionItems used for conversion."""
+
+        conversion_items = super()._get_conversion_items()
+        return conversion_items.union(
+            {
+                ConversionItem(
+                    "object__id",
+                    "event_type",
+                    lambda event_type: self.home_page + event_type,
+                ),
+            }
+        )
--- a/tests/models/edx/converters/xapi/test_base.py
+++ b/tests/models/edx/converters/xapi/test_base.py
@@ -1,0 +1,59 @@
+"""Tests for the base xAPI Converter"""
+
+from uuid import UUID
+
+import pytest
+
+from ralph.exceptions import ConfigurationException
+from ralph.models.edx.converters.xapi.base import BaseXapiConverter
+
+
+@pytest.mark.parametrize("uuid_namespace", ["ee241f8b-174f-5bdb-bae9-c09de5fe017f"])
+def test_base_xapi_converter_successful_initialization(
+    uuid_namespace,
+):
+    """Tests BaseXapiConverter initialization."""
+
+    class DummyBaseXapiConverter(BaseXapiConverter):
+        """Dummy implementation of abstract BaseXapiConverter."""
+
+        def _get_conversion_items(self):  # pylint: disable=no-self-use
+            """Returns a set of ConversionItems used for conversion."""
+
+            return set()
+
+    converter = DummyBaseXapiConverter(uuid_namespace, "https://fun-mooc.fr")
+    assert converter.home_page == "https://fun-mooc.fr"
+    assert converter.uuid_namespace == UUID(uuid_namespace)
+
+
+def test_base_xapi_converter_unsuccessful_initialization():
+    """Tests BaseXapiConverter failed initialization."""
+
+    class DummyBaseXapiConverter(BaseXapiConverter):
+        """Dummy implementation of abstract BaseXapiConverter."""
+
+        def _get_conversion_items(self):  # pylint: disable=no-self-use
+            """Returns a set of ConversionItems used for conversion."""
+
+            return set()
+
+    with pytest.raises(ConfigurationException, match="Invalid UUID namespace"):
+        DummyBaseXapiConverter(None, "https://fun-mooc.fr")
+
+
+@pytest.mark.parametrize(
+    "course_id,expected",
+    [
+        ("", {"course": None, "module": None}),
+        ("course-v1:+course+not_empty", {"course": None, "module": None}),
+        ("course-v1:org", {"course": None, "module": None}),
+        ("course-v1:org+course", {"course": None, "module": None}),
+        ("course-v1:org+course+", {"course": None, "module": None}),
+        ("course-v1:org+course+module", {"course": "course", "module": "module"}),
+    ],
+)
+def test_base_xapi_converter_parse_course_id(course_id, expected):
+    """Tests that the parse_course_id method returns the expected value."""
+
+    assert BaseXapiConverter.parse_course_id(course_id) == expected
--- a/tests/models/edx/converters/xapi/test_server.py
+++ b/tests/models/edx/converters/xapi/test_server.py
@@ -1,0 +1,106 @@
+"""Tests for the server event xAPI converter"""
+
+import json
+from uuid import UUID, uuid5
+
+import pytest
+from hypothesis import given, provisional, settings
+from hypothesis import strategies as st
+
+from ralph.models.converter import convert_dict_event, convert_str_event
+from ralph.models.edx.base import BaseContextField
+from ralph.models.edx.converters.xapi.server import ServerEventToPageViewed
+from ralph.models.edx.server import ServerEvent, ServerEventField
+
+
+@settings(max_examples=1)
+@given(
+    st.builds(
+        ServerEvent, event=st.builds(ServerEventField), referer=provisional.urls()
+    ),
+    provisional.urls(),
+)
+@pytest.mark.parametrize("uuid_namespace", ["ee241f8b-174f-5bdb-bae9-c09de5fe017f"])
+def test_models_edx_converters_xapi_server_server_event_to_xapi_convert_constant_uuid(
+    uuid_namespace, event, home_page
+):
+    """Tests that ServerEventToPageViewed.convert returns a JSON string with a constant UUID."""
+
+    event_str = event.json()
+    event = json.loads(event_str)
+    xapi_event1 = convert_str_event(
+        event_str, ServerEventToPageViewed(uuid_namespace, home_page)
+    )
+    xapi_event2 = convert_dict_event(
+        event, event_str, ServerEventToPageViewed(uuid_namespace, home_page)
+    )
+    assert xapi_event1.id == xapi_event2.id
+
+
+@settings(max_examples=1)
+@given(
+    st.builds(
+        ServerEvent,
+        event=st.builds(ServerEventField),
+        referer=provisional.urls(),
+        event_type=st.just("/main/blog"),
+        context=st.builds(
+            BaseContextField, user_id=st.just("1"), path=st.just("/main/blog")
+        ),
+    ),
+    provisional.urls(),
+)
+@pytest.mark.parametrize("uuid_namespace", ["ee241f8b-174f-5bdb-bae9-c09de5fe017f"])
+def test_models_edx_converters_xapi_server_server_event_to_xapi_convert_with_valid_event(
+    uuid_namespace, event, home_page
+):
+    """Tests that converting with ServerEventToPageViewed returns the expected xAPI statement."""
+
+    event_str = event.json()
+    event = json.loads(event_str)
+    xapi_event = convert_dict_event(
+        event, event_str, ServerEventToPageViewed(uuid_namespace, home_page)
+    )
+    xapi_event_dict = json.loads(xapi_event.json(exclude_none=True, by_alias=True))
+    assert xapi_event_dict == {
+        "id": str(uuid5(UUID(uuid_namespace), event_str)),
+        "actor": {
+            "account": {"homePage": home_page, "name": "1"},
+        },
+        "object": {
+            "definition": {
+                "name": {"en": "page"},
+                "type": "http://activitystrea.ms/schema/1.0/page",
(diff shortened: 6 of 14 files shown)


jmaupetit (commented):
We want to validate xAPI events too, so it appears that we need to first define xAPI Pydantic models for each edx supported event and a fields conversion set as:

```python
edx_to_xapi = {
  ("platform", "actor__account__homePage"),
  # [...]
}
```

Given those models and a conversion set, the converter signature would be like:

```python
def convert(src: EventModel, dest: EventModelClass, conversion: Set[Tuple]) -> EventModel:
    """Converts src model instance to a dest model instance given a conversion set.

    Returns instantiated model class with converted data.
    """
```

WDYT?

SergioSim:
> We want to validate xAPI events too, so it appears that we need to first define xAPI Pydantic models for each edx supported event and a fields conversion set as:
> 
> ```python
> edx_to_xapi = {
>   ("platform", "actor__account__homePage"),
>   # [...]
> }
> ```
> 
> Given those models and a conversion set, the converter signature would be like:
> 
> ```python
> def convert(src: EventModel, dest: EventModelClass, conversion: Set[Tuple]) -> EventModel:
>     """Converts src model instance to a dest model instance given a conversion set.
> 
>     Returns instantiated model class with converted data.
>     """
> ```
> 
> WDYT?

I agree that to validate the xAPI events we would need to use pydantic models. But I don't understand why we need them for conversion ?
This PR is not at the `converter signature` step, for now only the `conversion: Set[Tuple]` is defined (but I use classes instead ex. ServerEventToXapi) and how the `conversion: Set[Tuple]` `class` is used to convert from one format to another.

jmaupetit:
> I agree that to validate the xAPI events we would need to use pydantic models. But I don't understand why we need them for conversion ?

We need them because we will do both conversion and validation at the same time by instantiating an xAPI event model with data extracted from the edx event model instance thanks to the conversion table.

I think it's important at this stage that we agree on the methodology.

src/ralph/models/xapi/page_viewed.py
```diff
@@ -0,0 +1,119 @@
...
+    """Represents the `actor.account` xAPI field."""
+
+    name: constr(min_length=1) = "student"
+    homePage: constr(min_length=1)
+
+    @validator("name", pre=True, always=True)
+    def set_name(cls, name):  # pylint:disable=no-self-argument,no-self-use
+        """Sets the name to student when it's None or empty."""
+
+        return name or "student"
```
SergioSim, line 40: If I understood it right, this would be conversion and validation at the same time ?
We pass the edx `context.user_id` to the pydantic `name` field without change. Before checking for `constr(min_length=1)`  pydantic runs the validator `set_name` which checks/prepares the `name` field.

src/ralph/models/edx/converters/xapi/server.py
```diff
@@ -0,0 +1,75 @@
...
+                    XAPI_EXTENSION_IP: GetFromField("ip"),
+                    XAPI_EXTENSION_ORG_ID: GetFromField("context__org_id"),
+                    XAPI_EXTENSION_PATH: GetFromField("context__path"),
+                    XAPI_EXTENSION_REQUEST: GetFromField("event"),
+                },
+            },
+            "object": {
+                "id": GetFromField(
+                    "event_type", lambda event_type: self.platform + event_type
+                )
```
SergioSim, line 72: I tried to replace `GetFromField` with `LazyModelField` without success for now.
`platform` is not in the edx event but provided with an cli argument.
In pydantic it's possible to define [private model attributes](https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes) but it seems there is no way to access them from a validator function ? Any Ideas ?

quitterie-lcs: Just to understand better when using `LazyModelField`, you don't apply the transformation `x : x+ self.platform` but the [path parameter](https://github.com/openfun/ralph/blob/master/src/ralph/models/selector.py#L24) is directly `x + self.platform` ?

src/ralph/models/xapi/page_viewed.py
quitterie-lcs: Is this validator useful? You can assign the default value to student directly, isnt'it what you are already doing? 
Or maybe another possibility would be to assign the type str and as you only have two attributes of str type you could use the config class.
The more I get to use pydantic, and the more I think validator are not always a good solution

```
class ActorAccountField(BaseModelWithConfig):
    """Represents the `actor.account` xAPI field."""

    name: str = "student"
    homePage: str

        class Config:
            min_anystr_length = 1
```

src/ralph/models/edx/converters/xapi/server.py
SergioSim: Yes, `LazyModelField` don't do any transformations it helps just to retrieve a value from the event.

`GetFromField`on the other hand  retrieves the value from the event and applies a transformation.

If I understood it right, we want to pass the value (of `event_type`) "as it is" in the source event to the pydantic xAPI model and then let the model do the transformation part (x + self.platform).

But I don't find a way to pass `self.platform` to the pydantic xAPI model and access it's value form the validator to do the transformation (Without using global or static variables).

src/ralph/models/xapi/page_viewed.py
SergioSim: I thing this is a confusing part of doing validation and conversion at the same time.
The validator actually does the "transfomation" part.
If we omit for example the homePage attribute: 
For `ActorAccountField(name="")`, `ActorAccountField(name=None)`, `ActorAccountField()`, `ActorAccountField(name={})`, etc. we would get `ActorAccountField(name="student")` and not raise exceptions.

src/ralph/models/converter.py
```diff
@@ -0,0 +1,77 @@
...
+        """Returns a conversion set used for conversion."""
+
+    def convert(self, event: dict, event_str: str):
+        """Converts and serializes the event.
+
+        Args:
+            event (dict): the event to convert and serialize.
+            event_str (str): the event original string value.
+        Returns:
+            (string): converted and JSON serialized event.
```
jmaupetit, line 50: From an API perspective, I think we should not returned a serialized object but rather a (converted) pydantic model instance.

jmaupetit (commented):
Looks promising! I'll give it a new round once isolated from #84

tests/models/edx/converters/xapi/test_base.py
```diff
@@ -0,0 +1,19 @@
+"""Tests for the base xAPI Converter"""
+
+from ralph.models.edx.converters.xapi.base import BaseXapiConverter
+
+
+class DummyBaseXapiConverter(BaseXapiConverter):
```
quitterie-lcs, line 6: I would integrate this  class in the `test_models_edx_converters_xapi_base_base_xapi_converter_successful_initialization` function

tests/models/test_converter.py
```diff
@@ -0,0 +1,116 @@
...
+
+class DummyBaseConverter(BaseConverter):
+    """Dummy implementation of abstract BaseConverter."""
+
+    __model__ = BaseModel
+
+    def get_conversion_set(self):  # pylint: disable=no-self-use
+        """Returns a conversion set used for conversion."""
+
+        return set()
```
quitterie-lcs, line 30: Idem, I would define this class in each test function it is necessary, so that we keep the test file clean: we would only have test functions in it. WDYT?

src/ralph/models/converter.py
SergioSim: Good Idea) Thanks) updated)

src/ralph/models/converter.py
```diff
@@ -0,0 +1,77 @@
...
+        self.conversion_dict = {}
+        self.field_array = []
+        for field in self.get_conversion_set():
+            destination_path = field[0].split(MODEL_PATH_SEPARATOR)
+            value = field[1]() if callable(field[1]) else None
+            set_dict_value_from_path(self.conversion_dict, destination_path, value)
+            if not value:
+                transformer = field[2] if len(field) == 3 else lambda x: x
+                source_path = field[1].split(MODEL_PATH_SEPARATOR)
+                self.field_array.append((destination_path, source_path, transformer))
```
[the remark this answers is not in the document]
SergioSim: Good Idea) thanks) extracted this part in `prepare_converter` function)

tests/models/edx/converters/xapi/test_server.py
```diff
@@ -0,0 +1,146 @@
...
+from ralph.models import converter as converter_module
+from ralph.models.edx.converters.xapi.server import ServerEventToXapi
+from ralph.models.edx.server import ServerEvent, ServerEventField
+
+
+@pytest.fixture(autouse=True)
+def setup_edx_uuid_namespace(monkeypatch):
+    """Monkeypatches the required EDX_UUID_NAMESPACE configuration."""
+
+    uuid = "51f5c14e-0150-4766-88d2-6f7991ad6bfe"
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks) Moving `uuid` as an object memeber simplified the testing and removed this setup function)

tests/models/edx/converters/xapi/test_server.py
```diff
@@ -0,0 +1,146 @@
...
+    st.builds(
+        ServerEvent, event=st.builds(ServerEventField), referer=provisional.urls()
+    )
+)
+def test_models_edx_converters_xapi_server_server_event_to_xapi_convert_constant_uuid(
+    event,
+):
+    """Tests that ServerEventToXapi.convert returns a JSON string with a constant UUID."""
+
+    home_page = "https://www.fun-mooc.fr"
```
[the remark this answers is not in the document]
SergioSim: Sounds good) thanks) using hypothesis for the `home_page` now)

tests/models/edx/converters/xapi/test_server.py
```diff
@@ -0,0 +1,146 @@
...
+    event,
+):
+    """Tests that ServerEventToXapi.convert returns the expected converter xAPI statement."""
+
+    home_page = "https://www.fun-mooc.fr"
+    event_str = event.json()
+    event = json.loads(event_str)
+    event["context"]["user_id"] = "1"
+    event["event_type"] = "/main/blog"
+    event["context"]["path"] = "/main/blog"
```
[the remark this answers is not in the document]
SergioSim: Thanks) delegated partly this part to hypothesis)

tests/models/edx/converters/xapi/test_base.py
SergioSim: Good Idea) Thanks) Moved the class definitions inside the test methods)

tests/models/test_converter.py
SergioSim: Thanks) Updated)

src/ralph/models/converter.py
```diff
@@ -0,0 +1,82 @@
...
+    def get_conversion_set(self):
+        """Returns a conversion set used for conversion.
+
+        Each element in the conversion set should be a tuple of length 2 or 3.
+        - The first element of each tuple is a destination path string.
+        - The second element can be a source path string or a function with no arguments.
+        - The optional third element defines a transformer function with one argument.
+        """
+
+    def convert(self, event: dict, event_str: str):
```
jmaupetit, line 38: Maybe we have a design issue here: providing both the model instance and its source string is not that elegant. It's only used to calculate a UUID given a string, so maybe we should delegate this elsewhere (not in the converter), and provide a UUID to the `convert` method instead? Or should we embed source string to our event models?

src/ralph/models/edx/converters/xapi/base.py
```diff
@@ -0,0 +1,27 @@
...
+    def parse_course_id(course_id: str):
+        """Returns a dictionary with `course` and `module` of edX event's `context.course_id`."""
+
+        course_and_module = {"course": None, "module": None}
+        if not course_id:
+            return course_and_module
+        match = re.match(r"^course-v1:.+\+(.+)\+(.+)$", course_id)
+        if not match:
+            return course_and_module
+        return {"course": match.group(1), "module": match.group(2)}
```
quitterie-lcs, line 27: Maybe it can be made in a more compact way. 
```suggestion
        match = re.match(r"^course-v1:.+\+(.+)\+(.+)$", course_id)
        if match:
            return {"course": match.group(1), "module": match.group(2)}
        else:
            return {"course": None, "module": None}
```

src/ralph/models/converter.py
```diff
@@ -0,0 +1,82 @@
...
+
+from ralph.defaults import MODEL_PATH_SEPARATOR
+from ralph.exceptions import ConversionException
+from ralph.utils import get_dict_value_from_path, set_dict_value_from_path
+
+
+class BaseConverter(ABC):
+    """Converter Base Class.
+
+    Converters define a conversion set to convert from one event type to another.
```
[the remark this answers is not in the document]
SergioSim: Nice catch) Thanks)

src/ralph/models/edx/converters/xapi/server.py
```diff
@@ -0,0 +1,55 @@
...
+    WARNING: The converter does not include the following edX fields as they may
+    contain sensitive data: `username`, `referer`, `event`, `event_source`, `ip`,
+    `agent`, `accept_language:`, `context.course_user_tags`
+
+    WARNING: The converter may not include the following edX fields:
+        - context.org_id: When `org_id` is an empty string.
+        - context.course_id: When `course_id` is an empty string.
+    """
+
+    __model__ = PageViewed
```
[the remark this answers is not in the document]
SergioSim: This might be confusing, indeed. The model of the converter is the destination model. Maybe it's worth renaming it to `__destination__` or `__result__` ?

src/ralph/models/converter.py
SergioSim: You are right) Thanks) Removed the `uuid_namespace` object field and replaced the convert method argument `event_str` with `uuid`)

src/ralph/models/edx/converters/xapi/base.py
SergioSim: Good Idea) Thanks) Indeed, course_id is never `None`, removed the check)

src/ralph/models/converter.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    __model__: BaseModel
+
+    def __init__(self, uuid_namespace: UUID):
+        """Initializes BaseConverter."""
+
+        self.uuid_namespace = uuid_namespace
+        self.field_array = []
+        self.prepare_converter(self.field_array, self.get_conversion_set())
```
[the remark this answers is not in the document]
SergioSim: I'm hesitating about this, the preparation step is used to split the source and destination paths into lists once the object is initialized. (We could keep the same object to convert multiple events in a row).
I could move this part to the convert function (but then it will be re-executed for each event). WDYT ?

src/ralph/models/converter.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @abstractmethod
+    def get_conversion_set(self):
+        """Returns a conversion set used for conversion.
+
+        Each element in the conversion set should be a tuple of length 2 or 3.
+        - The first element of each tuple is a destination path string.
+        - The second element can be a source path string or a function with no arguments.
+        - The optional third element defines a transformer function with one argument.
```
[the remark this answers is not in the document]
SergioSim: Good Idea) I propose to use `lambda _: _` in place of `None` - this avoids to treat it as a special case) WDYT ?
Regarding, list of (transformation) functions (composition): I don't see in which case we would need multiple functions in one tuple ? Do you mean to combine multiple sources for a single destination ?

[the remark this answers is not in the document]
jmaupetit: > I propose to use lambda _: _ in place of None - this avoids to treat it as a special case) 

From a developer/API perspective wouldn't it be better to use a named tuple with defaults? _E.g._:

```python
from collections import namedtuple

ConversionItem = namedtuple(
    "ConversionItem",
    "dest src transformation",
    default=[None, lambda _: _]
)

example_item = ConversionItem(
    dest="actor__account__homePage", 
    transformation=lambda:"example.org"
)
```

>  Regarding, list of (transformation) functions (composition): I don't see in which case we would need multiple functions in one tuple ? Do you mean to combine multiple sources for a single destination ?

I mean combining multiple simple transformations to a source, _e.g._:

```python
transformed = base64(lower(src))
```

That can be listed in this tuple as:

```python
transformations = [lower, base64]
```

src/ralph/models/converter.py
```diff
@@ -0,0 +1,82 @@
...
+        Args:
+            event (dict): the event to convert and serialize.
+            event_str (str): the event original string value.
+        Returns:
+            (model): converted pydantic model.
+        """
+
+        converted_event = {}
+        for destination_path, source_path, transformer in self.field_array:
+            field = get_dict_value_from_path(event, source_path)
```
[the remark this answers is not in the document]
SergioSim: This was confusing, sorry, the source_path switched with transformer in case it were callable in the prepare method. Now, with constant tuple length of 3 this is not  the case anymore)

src/ralph/models/converter.py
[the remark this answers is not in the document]
jmaupetit: I think it will only have a minor performance impact but you are right. You can leave it as is for now.

src/ralph/models/edx/converters/xapi/server.py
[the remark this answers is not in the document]
jmaupetit: What is confusing is that your converter is called `ServerEventToXapi` for a destination model that is a `PageViewed` event.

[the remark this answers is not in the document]
SergioSim: Good point) Thanks) I agree) will rename the class to `ServerEventToPageViewed`)

src/ralph/models/converter.py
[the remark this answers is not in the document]
SergioSim: This is a great Idea) Thanks) As discussed, updated this part with dataclasses (similar to named tuples)

src/ralph/models/converter.py
```diff
@@ -1,65 +1,82 @@
...
-            field_array.append((destination_path, source_path, transformer))
+    def _get_conversion_items(self) -> set[ConversionItem]:
+        """Returns a set of ConversionItems used for conversion."""
+
+    def __iter__(self):
+        return iter(self._conversion_items)
+
+
+def convert(event: dict, conversion_set: BaseConversionSet, uuid: str) -> BaseModel:
+    """Converts the event using the provided converter and UUID."""
```
jmaupetit, line 62:
```suggestion
    """Converts the event using the provided converter and a UUID."""
```

src/ralph/models/converter.py
```diff
@@ -1,65 +1,82 @@
...
+
+    converted_event = {}
+    for conversion_item in conversion_set:
+        transformed_field = conversion_item.get_value(event)
```
jmaupetit, line 66:
```suggestion
        value = conversion_item.get_value(event)
```

src/ralph/models/converter.py
```diff
@@ -1,65 +1,82 @@
...
+        if transformed_field not in [None, "", {}]:
+            set_dict_value_from_path(
+                converted_event, conversion_item.dest, transformed_field
+            )
+    converted_event["id"] = uuid
```
jmaupetit: So we assume that whatever the destination format/model, an ID field is defined? Maybe we should not consider that it has not already been set?

```suggestion
    if "id" not in converted_event:
        converted_event["id"] = uuid
```

src/ralph/models/converter.py
```diff
@@ -1,65 +1,82 @@
...
+    return conversion_set.__dest__(**converted_event)
+
+
+def convert_raw(
+    raw_event: str, converter: BaseConversionSet, uuid_namespace: UUID
+) -> BaseModel:
+    """Converts the raw event string using the provided converter."""
+
+    event = json.loads(raw_event)
```
jmaupetit, line 79: Maybe we should secure this and raise a custom exception if the raw event is poorly formatted?

src/ralph/models/converter.py
SergioSim: Good Idea) thanks) updated)

src/ralph/models/converter.py
SergioSim: Nice catch) thanks) updated with:
```
def convert_dict_event(
    event: dict, conversion_set: BaseConversionSet, id_: UUID5
) -> BaseModel:
    """Converts the event dict using the provided conversion_set and a UUID5 `id` field.
```

src/ralph/models/converter.py
SergioSim: Good point) Yes, indeed) Now the `id` is set at the start and can be overwritten if it's set in the `conversion_set`.
```converted_event = {"id": id_}```

File: src/ralph/models/converter.py (hunk 20 above)
Comment (SergioSim) [reply src/ralph/models/converter.py]: Good Idea) Thanks) updated) Now it's raising a `BadFormatException`

File: CHANGELOG.md [hunk 21]
```diff
@@ -10,6 +10,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- EdX server event to xAPI page `viewed` and edX `page_close` to xAPI
+  page `terminated` converters
```
```suggestion
- EdX to xAPI converters for page `viewed` and`page_close` events
- Implement core event format converter
```

src/ralph/models/conversion_set.py
```diff
@@ -0,0 +1,56 @@
+"""Base conversion set definition"""
```
jmaupetit, line 1: I would move this module content to `src/ralph/models/converter.py`. It's not that big for now.

src/ralph/models/conversion_set.py
```diff
@@ -0,0 +1,56 @@
...
+    dest: tuple[str]
+    src: Union[tuple[str], None]
+    transformers: tuple[Callable[[Any], Any]]
+
+    def __init__(self, dest: str, src=None, transformers=lambda _: _):
+        object.__setattr__(self, "dest", tuple(dest.split(MODEL_PATH_SEPARATOR)))
+        src = tuple(src.split(MODEL_PATH_SEPARATOR)) if src else None
+        object.__setattr__(self, "src", src)
+        transformers = tuple([transformers]) if callable(transformers) else transformers
+        object.__setattr__(self, "transformers", transformers)
```
jmaupetit, line 26: We have to admit that this frozen dataclass API is ugly :sweat_smile:

src/ralph/models/conversion_set.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def get_value(self, event: dict):
+        """Calls transformer functions and returns the transformed value."""
```
jmaupetit, line 29:
```suggestion
        """Returns fetched source value after having applied transformers to it."""
```

src/ralph/models/converter.py
jmaupetit: It seems confusing to me.

src/ralph/models/converter.py
```diff
@@ -0,0 +1,71 @@
...
+        ValidationError: When the final converted event is invalid.
+    """
+
+    converted_event = {"id": id_}
+    for conversion_item in conversion_set:
+        try:
+            value = conversion_item.get_value(event)
+        except Exception as err:
+            msg = "Failed to get the transformed value for field: %s"
+            raise ConversionException(msg % conversion_item.src) from err
```
jmaupetit, line 39:
```suggestion
            msg = f"Failed to get the transformed value for field: {conversion_item.src}"
            raise ConversionException(msg) from err
```

src/ralph/models/conversion_set.py
SergioSim: Yes Indeed :smile: we could use regular dataclasses if we would put them in a `conversion_list` instead of a `conversion_ set`. WDYT ?

src/ralph/models/conversion_set.py
SergioSim: Good Idea) Thanks) Updated)

src/ralph/models/conversion_set.py
SergioSim: Sounds good) Thanks) Updated)

src/ralph/models/converter.py
SergioSim: Do you prefer the way it was before where the `conversion_set` could not overwrite the id ?
I assume in the convert method that it's up to the caller to set the right ID (the caller could generate a new one or extract an existing one)

src/ralph/models/conversion_set.py
jmaupetit: The question is: using a set is smart but do we need advanced set features (union, etc.) or a list would be sufficient?

SergioSim: For now we only use `union` to combine the `base` conversion_set with `server_event` or `page_closed` conversion_set, which is almost the same as appending to a list.
But, Indeed it could be useful in future to easily construct new conversion_sets by adding/removing/intersecting etc.)

src/ralph/models/converter.py
```diff
@@ -24,18 +21,50 @@ class ConversionItem:
...
+        """Initializes ConversionItem.
+
+        Args:
+            dest (str): The destination path where to place the converted value.
+            src (str or None): The source from where the value to convert is fetched.
+                - When `src` is a path (ex. `context__user_id`) - the value is the item
+                    of the event at the path.
+                - When `src` is `*` - the value is the whole event.
+                - When `src` is `**` - the value is the original event string.
+                - When `src` is `None` - the value is `None`.
```
jmaupetit: I would suggest to add a `raw_input` boolean argument instead (which defaults to `False`) and add a `raw_input` attribute that will be used in the `get_value` method (see other comments). WDYT?

src/ralph/models/converter.py
```diff
@@ -24,18 +21,50 @@ class ConversionItem:
...
-        src = tuple(src.split(MODEL_PATH_SEPARATOR)) if src else None
+        special = [None, "*", "**"]
+        src = src if src in special else tuple(src.split(MODEL_PATH_SEPARATOR))
         object.__setattr__(self, "src", src)
         transformers = tuple([transformers]) if callable(transformers) else transformers
         object.__setattr__(self, "transformers", transformers)
 
-    def get_value(self, event: dict):
-        """Returns fetched source value after having applied transformers to it."""
+    def get_value(self, event: dict, event_str: str):
```
jmaupetit, line 45:
```suggestion
    def get_value(self, data: Union[dict, str]):
```

src/ralph/models/converter.py
```diff
@@ -24,18 +21,50 @@ class ConversionItem:
...
 
-        value = get_dict_value_from_path(event, self.src)
-        for transformer in self.transformers:
-            value = transformer(value)
+        if self.src == "*":
+            value = event
+        elif self.src == "**":
+            value = event_str
+        else:
+            value = get_dict_value_from_path(event, self.src)
```
jmaupetit, line 61:
```suggestion
        if self.raw_input:
            value = data
        else:
            value = get_dict_value_from_path(data, self.src)
```

src/ralph/models/converter.py
```diff
@@ -79,44 +108,33 @@ def convert_dict_event(
...
+    converted_event = {}
     for conversion_item in conversion_set:
-        try:
-            value = conversion_item.get_value(event)
-        except Exception as err:
-            msg = (
-                f"Failed to get the transformed value for field: {conversion_item.src}"
-            )
-            raise ConversionException(msg) from err
+        value = conversion_item.get_value(event, event_str)
```
jmaupetit, line 113:
```suggestion
        data = event
        if conversion_item.raw_input:
            data = event_str
        value = conversion_item.get_value(data)
```

jmaupetit (commented):
Great! 

I have a suggestion to avoid having `None`, `*` or `**` special cases.

src/ralph/models/converter.py
SergioSim: Good Idea! Thanks) Updated this part to reduce the special cases:
Now, when `src` is None the whole event is passed else it's a path and we fetch a specific field.

src/ralph/models/converter.py
SergioSim: Thanks) This simplifies the `get_value` method a lot) The only thing that bothers me here is that now the caller has to check the `conversion_item` `raw_input` field before calling `get_value` to pass the right value (ether dict or string).

src/ralph/models/converter.py
```diff
@@ -19,53 +19,49 @@ class ConversionItem:
...
             src (str or None): The source from where the value to convert is fetched.
                 - When `src` is a path (ex. `context__user_id`) - the value is the item
                     of the event at the path.
-                - When `src` is `*` - the value is the whole event dictionary.
-                - When `src` is `**` - the value is the original event string.
-                - When `src` is `None` - the value is `None`.
+                - When `src` is `None` - the value is the whole event.
             transformers (function or tuple of functions): The function(s) to apply on the source
                 value.
+            raw_input (bool): Flag indicating wether `get_value` will receive a raw event string or
```
jmaupetit, line 35:
```suggestion
            raw_input (bool): Flag indicating whether `get_value` will receive a raw event string or
```

src/ralph/models/converter.py
SergioSim: Thanks) updated this part, but now it's the `src` field deciding if we use the whole event or a specific field.

src/ralph/models/converter.py
SergioSim: Nice catch) Thanks)

src/ralph/models/converter.py
SergioSim: I would prefer to delegate this check to the `get_value` method. WDYT ?

src/ralph/models/converter.py
```diff
@@ -0,0 +1,137 @@
...
+    """ConversionSet Base Class.
+
+    ConversionSets are used to convert from one event format to another.
+    """
+
+    __src__: BaseModel
+    __dest__: BaseModel
+
+    def __init__(self):
+        """Initializes BaseConversionSet"""
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks)

tests/models/test_converter.py
```diff
@@ -0,0 +1,180 @@
...
+        (ConversionItem(""), {"foo": "bar", "baz": {"qux": "quux"}}),
+        (ConversionItem("", None, lambda event: event["foo"]), "bar"),
+    ],
+)
+def test_models_converter_conversion_item_get_value_with_successful_transformers(
+    conversion_item, expected
+):
+    """Tests that the get_value method returns the expected value."""
+
+    event = {"foo": "bar", "baz": {"qux": "quux"}}
```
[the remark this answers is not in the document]
SergioSim: Yes, sure) Thanks)

jmaupetit (approved):
LGTM, I only have a suggestion for far too long test names :sweat_smile:

tests/models/edx/converters/xapi/test_base.py
```diff
@@ -0,0 +1,61 @@
...
+    [
+        ("", {"course": None, "module": None}),
+        ("course-v1:+course+not_empty", {"course": None, "module": None}),
+        ("course-v1:org", {"course": None, "module": None}),
+        ("course-v1:org+course", {"course": None, "module": None}),
+        ("course-v1:org+course+", {"course": None, "module": None}),
+        ("course-v1:org+course+module", {"course": "course", "module": "module"}),
+    ],
+)
+def test_models_edx_converters_xapi_base_base_xapi_converter_parse_course_id(
```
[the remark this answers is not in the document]
SergioSim: Good point) Thanks) Indeed the test names get quite long. Updated them to follow the pattern `test_` `file_name_` `class/method_` `test_name`

PR #81, opened 2021-03-26
⚠️ [DO NOT MERGE] genson JSON Schema generator

SergioSim:
## Purpose

Generating JSON schemas from existing data could help us define relevant pydantic models.

## Proposal

Maybe we could run this version of ralph on one log file (in test environment?) to get a first look at the edX event schemas ?

I have run it on my logs in local and got the following results:  https://pastebin.com/SFrc4RNU
By [parsing the nested JSONs](https://github.com/openfun/ralph/pull/81/commits/d62f7ca40c591ae1530951f5aec874051fadde60) in the events it's possible to get their schema too: https://pastebin.com/Cpb2jBd0 
The `datamodel-codegen` generated pydantic models are [here](https://github.com/SergioSim/ralph/commit/b66add20d5060971ecafcf734be5f11256359563).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ install_requires =
     click==7.1.2
     click-option-group==0.5.2
     elasticsearch==7.11.0
+    genson==1.2.2 
     ovh==0.5.0
     pydantic==1.8.1
     python-keystoneclient==4.2.0
--- a/src/ralph/cli.py
+++ b/src/ralph/cli.py
@@ -19,6 +19,7 @@ from ralph.defaults import (
 )
 from ralph.exceptions import UnsupportedBackendException
 from ralph.logger import configure_logging
+from ralph.models.generator import generate_json_schemas
 from ralph.models.selector import ModelSelector
 from ralph.models.validator import Validator
 from ralph.utils import (
@@ -163,6 +164,26 @@ def extract(parser):
 
     for event in parser.parse(sys.stdin):
         click.echo(event)
+
+
+@cli.command()
+@click.option(
+    "-f",
+    "--format",
+    "format_",
+    type=click.Choice(["edx"]),
+    required=True,
+    help="Input events format to validate",
+)
+def genson(format_):
+    """Generates JSON schemas from input events that are not covered by validate."""
+
+    logger.info("Generating %s JSON schemas using genson", format_)
+
+    model_selector = ModelSelector(f"ralph.models.{format_}")
+
+    for schema in generate_json_schemas(sys.stdin, model_selector):
+        click.echo(schema)
 
 
 @cli.command()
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -26,6 +26,7 @@ install_requires =
     click==7.1.2
     click-option-group==0.5.2
     elasticsearch==7.11.0
+    genson==1.2.2
```
jmaupetit, line 29: Why preferring `genson` over [`datamodel-code-generator`](https://github.com/koxudaxi/datamodel-code-generator/) as the latest supports JSON input data?

SergioSim: `genson` is able to update it's JSON schema while receiving additional events while `datamodel-code-generator` only takes one `dict` or JSON schema and generates the corresponding `pydantic` model. They work in pair. We feed `genson` 100 events of source `browser` / type `seq_next` and `genson` generates a JSON schema that would match all these events. Then we use `datamodel-code-generator` to generate the corresponding `pydantic` model.

src/ralph/cli.py
```diff
@@ -165,6 +166,26 @@ def extract(parser):
...
+@click.option(
+    "-f",
+    "--format",
+    "format_",
+    type=click.Choice(["edx"]),
+    required=True,
+    help="Input events format to validate",
+)
+def genson(format_):
+    """Generates JSON schemas from input events that are not covered by validate."""
```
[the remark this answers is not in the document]
SergioSim: I think we don't need to generate schemas we already have ? But the actual reason is because this helped to remove the `simple server event`s from the scope as they can't be separated by `event_source`/`event_type`.

setup.cfg
SergioSim: Apropos, `genson` can be extended to include additional information in the JSON schemas (minimum/maximum values for integers, min-max length of strings, discover categorical values in strings using regex etc... But I'm hesitating to use these features at this stage as this could include some user data in the JSON schemas which should be avoided.

jmaupetit: Okay, the combination of the two tools makes sense, thanks for the clarification.

src/ralph/cli.py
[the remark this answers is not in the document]
jmaupetit: > I think we don't need to generate schemas we already have?

It will constitute a validation test of the proposed approach :wink:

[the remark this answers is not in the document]
SergioSim: Good point) You are right) I will update this part to include all events)

setup.cfg
jmaupetit: So maybe we should add `datamodel-code-generator` to dependencies, so that the Docker image bundles tools required to generate models from JSON-serialized data?

SergioSim: Good Idea) Thanks)

SergioSim:
As discussed, moved this part to a separate project ([gen4ralph](https://github.com/SergioSim/gen4ralph)), closing this PR.

PR #82, opened 2021-03-29
Improve event pydantic models structure

quitterie-lcs:
## Purpose

Current pydantic models are defined with a semantic including 'model'. It is confusing as some of models are for events and other for fields. To clarify the definition, event models have been renamed only with the edx event name and field with field suffix.

## Proposal

- [x] remove the Model suffix from models
- [x] explicit base model naming that cannot be used directly as AbstractXXX
- [x] rewrite tests with `hypothesis`


jmaupetit (commented):
Great! You can mention that we've renamed our models in the CHANGELOG (and that we now use hypothesis in our tests).

tests/models/edx/test_base.py
```diff
@@ -1,84 +1,24 @@
...
-def test_models_edx_base_event_with_invalid_content(kwargs, error):
-    """Tests that an invalid base event raises a ValidationError."""
+def test_models_edx_base_event_with_valid_content(event):
+    """Tests that a valid base event does not raise a ValidationError."""
 
-    with pytest.raises(ValidationError, match=error):
-        BaseEventFactory(**kwargs)
+    assert len(event.username) == 0 or (len(event.username) in range(2, 31, 1))
+    assert (
+        re.match(r"^course-v1:.+\+.+\+.+$", event.context.course_id)
```
SergioSim, line 22: To test a regex I would prefer to see some example values it's matching and others for which it's not matching. WDYT ?

tests/models/edx/test_browser.py
```diff
@@ -1,77 +1,16 @@
...
-    ],
-)
-def test_models_edx_browser_page_close_browser_event_with_invalid_content(
-    kwargs, error
-):
-    """Tests that an invalid page_close browser event raises a ValidationError."""
-
-    with pytest.raises(ValidationError, match=error):
-        PageCloseBrowserEventFactory(**kwargs)
+    assert re.match(r"^[a-f0-9]{32}$", event.session) or event.session == ""
```
SergioSim, line 16: Same question about regex tests

tests/models/test_selector.py
```diff
@@ -163,14 +158,10 @@ def test_models_selector_model_selector_model_rules(model_rules, rules):
...
-    [
-        (ServerEventFactory(), ServerEventModel),
-        (PageCloseBrowserEventFactory(), PageCloseBrowserEventModel),
-    ],
-)
-def test_models_selector_model_selector_get_model_with_valid_event(event, model):
+@pytest.mark.parametrize("model", [PageClose])
+@settings(max_examples=1)
+@given(st.builds(PageClose, referer=provisional.urls(), page=provisional.urls()))
+def test_models_selector_model_selector_get_model_with_valid_event(model, event):
```
SergioSim, line 164: Is it possible to include the `ServerEvent` case in this test too ? If not, maybe we could split this test in two tests - one for `ServerEvent` and one for `PageClose` ?

SergioSim (commented):
Thank you for improving our models structure and tests !

tests/models/test_selector.py
quitterie-lcs: Why do you want to test this mechanism with two examples? Once it has been tested with one event, I think it is not necessary to test with PageClose? WDYT?

SergioSim: The Idea is to use this test for all event models we define. It assures that for a given event the `get_model` method returns the corresponding model.

quitterie-lcs: I think that is a point we need to discuss IRL. I would not implement this test for two reasons: 
1. we test the selector mechanism here and this mechanism is not event-specific
2. maintain the tests for all the event would be very huge and maybe too hard to maintain when having all events implemented
WDYT ?

SergioSim: Ok, let's discuss this IRL) Indeed this test might get huge with a lot of events, you are right. But for me this mechanism is event-specific as each event implements the `__selector__` field which is responsible for the model retrieval in `get_model`.

jmaupetit: As discussed with @quitterie-lcs, model selectors can be tested in the model's test suite, not here. Only the model selector logic may be tested here. WDYT?

quitterie-lcs: @SergioSim I got the point of testing `get_model` for each event. Hence, we can move both tests `test_models_selector_model_selector_get_model_with_valid_event` and `test_models_selector_model_selector_get_model_with_invalid_event` in the corresponding model test file?

SergioSim: Good point) Yes, this would be a solution) If I understood it right, we would write for each model a test case with the ModelSelector in the corresponding model's test suite ? I could propose to automatize this test in a single one:

1. for each `ModelSelector` supported model:
2. generate `test_event` with hypothesis
3. check that `ModelSelector.get_model(test_event)` returns the corresponding model.

WDYT ?

jmaupetit: It's a security check that can be automated indeed, but I'm not sure it's really valuable. I would like to battle test our selectors on a model basis in various conditions.

tests/models/edx/test_base.py
quitterie-lcs: You're right, it is something to our model. Coming in my next push

src/ralph/models/edx/base.py
```diff
@@ -56,30 +55,24 @@ class BaseContextModel(BaseModelWithConfig):
...
-        course_id = values.get("course_id")
-        regex = f"course-v1:{org_id}\\+.+\\+.+"
-        if (org_id or course_id) and not re.match(regex, course_id):
-            raise ValueError(f"course_id must match regex `{regex}`")
-        return values
+class AbstractBaseEventField(BaseModelWithConfig):
+    """Represents the base event field model inherited by all event `event` fields.
+
+    The base model does not have any attributes as event field does not have common sub-fields.
+    """
```
jmaupetit, line 69: Except for the semantic, I don't get why we need this empty  abstract model.

jmaupetit (approved):
I only have minor comments. LGTM :muscle:

tests/models/edx/test_browser.py
```diff
@@ -1,77 +1,64 @@
...
+@given(st.builds(PageClose, referer=provisional.urls(), page=provisional.urls()))
+def test_models_selector_model_selector_get_model_with_valid_event(model, event):
+    """Tests given a page_close event the get_model method should return PageClose model."""
 
-    try:
-        PageCloseBrowserEventFactory()
-    except ValidationError:
-        pytest.fail("Valid page_close browser event should not raise exceptions")
+    event = json.loads(event.json())
+    assert ModelSelector(module="ralph.models.edx").get_model(event) is model
```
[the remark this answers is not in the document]
quitterie-lcs: true, that's better

PR #85, opened 2021-03-31
🔒(project) add system package upgrades in docker image builds

quitterie-lcs:
## Purpose

In Docker images, the upgrades of system packages is not run when building. It represents a security failure, as we let failures of
previous packages versions in the images. 

## Proposal

We have added the `apt-get upgrade` command to ensure that packages are up-to-date with their latest version and security failures have been removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Install security updates in project Docker images
 - Model selector to retrieve associated pydantic model of a given event
 - `validate` command to lint edx events using pydantic models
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -4,6 +4,11 @@ FROM python:3.9-slim as base
 # Upgrade pip to its latest release to speed up dependencies installation
 RUN pip install --upgrade pip
 
+RUN apt-get update && \
+    apt-get -y dist-upgrade && \
```
jmaupetit, line 8:
```suggestion
    apt-get -y upgrade && \
```

jmaupetit (commented):
Thanks! I would remove your commit subject to `project` :wink:

jmaupetit:
Also: can you track this change in the changelog?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Upgrade of system packages to install security updates when building Docker 
+images
```
```suggestion
- Install security updates in project Docker images
```

jmaupetit (approved):
I only have a minor comment.

PR #94, opened 2021-05-04
⚡️(tray) adjust probes configuration

jmaupetit:
## Purpose

A readiness probe for a pod that is not a server is useless. ~Moreover, ralph CLI's boot is too greedy, we need to reduce the liveness probe frequency to reduce useless CPU usage.~

Edit: as stated by @madmatah the same remark applies to the liveness probe.

## Proposal

- [x] remove deploy `readinessProbe`
- [x] ~increase `livenessProbe` `periodSeconds` to 1 minute~
- [x] remove deploy `livenessProbe`

Edit: we've decided to completely remove probes as they seems useless for a CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Migrate model tests from factories to hypothesis strategies
 - Upgrade `python-swiftclient` to `3.11.1`
 - Tray: switch from openshift to k8s (BC)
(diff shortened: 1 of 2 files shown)


madmatah (commented):
Beside that, I'd add the following question: what is the value added by the livenessProbe declared here?

jmaupetit:
None. Good point. I'll completely remove probes. :pray:

madmatah (approved):
You saved a bunch of kittens with this PR!

PR #98, opened 2021-05-21
🎨(models) restructure navigation xapi models code structure

quitterie-lcs:
## Purpose

xAPI events models are gathered on module. To get better tracking and
management of models and their subfields, code is restructured.

## Proposal

- [x] gather navigation event files on a specific folder
- [x] move specific navigation event fields to the corresponding folder


src/ralph/models/xapi/navigation/events.py
```diff
@@ -0,0 +1,33 @@
+"""Navigation xAPI event definitions"""
```
jmaupetit, line 1: For semantic consistency, this file should rather be called `statements.py`, right?

jmaupetit (approved):

PR #103, opened 2021-06-08
ORA save_submission converter

SergioSim:
## Purpose

The edX `openassessmentblock.save_submission` event occurs when a user saves his current response for an open-ended question.
We want to convert this event to xAPI format.

## Proposal

`ORASaveSubmission` validates the edX event
`ORASaveSubmissionToQuestionSaved` specifies the conversion set
`QuestionSaved` validates the converted xAPI statement

### DIFF
--- a/src/ralph/models/edx/base.py
+++ b/src/ralph/models/edx/base.py
@@ -5,7 +5,7 @@ from ipaddress import IPv4Address
 from pathlib import Path
 from typing import Dict, Literal, Optional, Union
 
-from pydantic import BaseModel, HttpUrl, constr
+from pydantic import AnyHttpUrl, BaseModel, constr
 
 
 class BaseModelWithConfig(BaseModel):
@@ -149,7 +149,7 @@ class BaseEdxModel(BaseModelWithConfig):
     ip: Union[IPv4Address, Literal[""]]
     agent: str
     host: str
-    referer: Union[HttpUrl, Literal[""]]
+    referer: Union[AnyHttpUrl, Literal[""]]
     accept_language: str
     context: BaseContextField
     time: datetime
--- a/src/ralph/models/xapi/fields/verbs.py
+++ b/src/ralph/models/xapi/fields/verbs.py
@@ -3,6 +3,8 @@
 from ..config import BaseModelWithConfig
 from ..constants import (
     LANG_EN_US_DISPLAY,
+    VERB_SAVED_DISPLAY,
+    VERB_SAVED_ID,
     VERB_TERMINATED_DISPLAY,
     VERB_TERMINATED_ID,
     VERB_VIEWED_DISPLAY,
@@ -35,4 +37,18 @@ class TerminatedVerbField(BaseModelWithConfig):
     id: VERB_TERMINATED_ID = VERB_TERMINATED_ID.__args__[0]
     display: dict[LANG_EN_US_DISPLAY, VERB_TERMINATED_DISPLAY] = {
         LANG_EN_US_DISPLAY.__args__[0]: VERB_TERMINATED_DISPLAY.__args__[0]
+    }
+
+
+class SavedVerbField(BaseModelWithConfig):
+    """Represents the `verb` xAPI Field for the action `saved`.
+
+    Attributes:
+       id (str): Consists of the value `http://activitystrea.ms/schema/1.0/save`.
+       display (dict): Consists of the dictionary `{"en-US": "saved"}`.
+    """
+
+    id: VERB_SAVED_ID = VERB_SAVED_ID.__args__[0]
+    display: dict[LANG_EN_US_DISPLAY, VERB_SAVED_DISPLAY] = {
+        LANG_EN_US_DISPLAY.__args__[0]: VERB_SAVED_DISPLAY.__args__[0]
     }
--- a/src/ralph/models/xapi/fields/verbs.py
+++ b/src/ralph/models/xapi/fields/verbs.py
@@ -2,7 +2,7 @@
 
 from ..config import BaseModelWithConfig
 from ..constants import (
-    LANG_EN_DISPLAY,
+    LANG_EN_US_DISPLAY,
     VERB_TERMINATED_DISPLAY,
     VERB_TERMINATED_ID,
     VERB_VIEWED_DISPLAY,
@@ -11,28 +11,28 @@ from ..constants import (
 
 
 class ViewedVerbField(BaseModelWithConfig):
-    """Represents the `verb` xAPI Field for page viewed xAPI statement.
+    """Represents the `verb` xAPI Field for the action `viewed`.
 
     Attributes:
        id (str): Consists of the value `http://id.tincanapi.com/verb/viewed`.
-       display (dict): Consists of the dictionary `{"en": "viewed"}`.
+       display (dict): Consists of the dictionary `{"en-US": "viewed"}`.
     """
 
     id: VERB_VIEWED_ID = VERB_VIEWED_ID.__args__[0]
-    display: dict[LANG_EN_DISPLAY, VERB_VIEWED_DISPLAY] = {
-        LANG_EN_DISPLAY.__args__[0]: VERB_VIEWED_DISPLAY.__args__[0]
+    display: dict[LANG_EN_US_DISPLAY, VERB_VIEWED_DISPLAY] = {
+        LANG_EN_US_DISPLAY.__args__[0]: VERB_VIEWED_DISPLAY.__args__[0]
     }
 
 
 class TerminatedVerbField(BaseModelWithConfig):
-    """Represents the `verb` xAPI Field for page terminated xAPI statement.
+    """Represents the `verb` xAPI Field for the action `terminated`.
 
     Attributes:
        id (str): Consists of the value `http://adlnet.gov/expapi/verbs/terminated`.
-       display (dict): Consists of the dictionary `{"en": "terminated"}`.
+       display (dict): Consists of the dictionary `{"en-US": "terminated"}`.
     """
 
     id: VERB_TERMINATED_ID = VERB_TERMINATED_ID.__args__[0]
-    display: dict[LANG_EN_DISPLAY, VERB_TERMINATED_DISPLAY] = {
-        LANG_EN_DISPLAY.__args__[0]: VERB_TERMINATED_DISPLAY.__args__[0]
+    display: dict[LANG_EN_US_DISPLAY, VERB_TERMINATED_DISPLAY] = {
+        LANG_EN_US_DISPLAY.__args__[0]: VERB_TERMINATED_DISPLAY.__args__[0]
     }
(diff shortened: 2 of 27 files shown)


src/ralph/models/xapi/fields/verbs.py
```diff
@@ -11,28 +15,42 @@
...
 
     Attributes:
        id (str): Consists of the value `http://id.tincanapi.com/verb/viewed`.
        display (dict): Consists of the dictionary `{"en": "viewed"}`.
     """
 
-    id: VERB_VIEWED_ID = VERB_VIEWED_ID.__args__[0]
-    display: dict[LANG_EN_DISPLAY, VERB_VIEWED_DISPLAY] = {
-        LANG_EN_DISPLAY.__args__[0]: VERB_VIEWED_DISPLAY.__args__[0]
+    id: Literal[VERB_VIEWED_ID] = VERB_VIEWED_ID
```
SergioSim, line 25: I was wrong, it's possible to have variables in Literals. Although `pylance` complains about it and marks everything in red - but it works.

src/ralph/models/xapi/fields/verbs.py
```diff
@@ -11,28 +15,42 @@
...
-    id: VERB_TERMINATED_ID = VERB_TERMINATED_ID.__args__[0]
-    display: dict[LANG_EN_DISPLAY, VERB_TERMINATED_DISPLAY] = {
-        LANG_EN_DISPLAY.__args__[0]: VERB_TERMINATED_DISPLAY.__args__[0]
+    id: Literal[VERB_TERMINATED_ID] = VERB_TERMINATED_ID
+    display: dict[Literal[LANG_EN_US_DISPLAY], Literal[VERB_TERMINATED_DISPLAY]] = {
+        LANG_EN_US_DISPLAY: VERB_TERMINATED_DISPLAY
+    }
+
+
+class SavedVerbField(BaseModelWithConfig):
```
SergioSim, line 45: xAPI verb definitions are quite similar, only the default `id` and `display` values are different, may be we could use a factory here ? Example: [verbs.py](https://github.com/SergioSim/ralph/blob/build-verb-field/src/ralph/models/xapi/fields/verbs.py)

src/ralph/models/edx/base.py
```diff
@@ -125,7 +125,7 @@ class BaseEvent(BaseModelWithConfig):
     ip: Union[IPv4Address, Literal[""]]
     agent: str
     host: str
-    referer: Union[HttpUrl, Literal[""]]
+    referer: Union[AnyHttpUrl, Literal[""]]
```
SergioSim, line 150: As we are not using the `referer` field in the xAPI conversion, may be we could make the validation less strict for this field ? Just checking if it's a string  or any value ? This could simplify the tests and speed up the validation. WDYT?

tests/models/edx/converters/xapi/test_open_response_assessment.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    xapi_event_dict = json.loads(xapi_event.json(exclude_none=True, by_alias=True))
+    assert xapi_event_dict == {
+        "id": str(uuid5(UUID(uuid_namespace), event_str)),
+        "actor": {
+            "account": {"homePage": home_page, "name": "1"},
+        },
+        "object": {
+            "definition": {
+                "name": {"en-US": "question"},
```
SergioSim, line 62: Question regarding the `object.definition.name` field:
We could pass here the edx `context.module.display_name` field which holds the title of the ORA question but this title is not always in English and I don't see an easy solution to guess the used language. Do you prefer the static `question` value or the latter?

src/ralph/models/edx/base.py
jmaupetit: Even if we are not using it (for now), we have to be pydantic on this :wink:

src/ralph/models/xapi/fields/verbs.py
jmaupetit: I prefer being declarative and thus repetitive for our models. Having factories will end up with an unmaintainable spaghetti code.

src/ralph/models/xapi/open_response_assessment/fields/objects.py
```diff
@@ -0,0 +1,42 @@
...
+    LANG_EN_US_DISPLAY,
+)
+from ...fields.objects import ObjectDefinitionExtensionsField, ObjectField
+
+
+class QuestionObjectDefinitionField(BaseModelWithConfig):
+    """Represents the `object.definition` xAPI field for question saved xAPI statement.
+
+    WARNING: It doesn't include the recommended `description` field nor the optional `moreInfo`,
+    `Interaction properties` and `extensions` fields.
```
jmaupetit, line 18: Shouldn't we declare them as optional?

src/ralph/models/xapi/open_response_assessment/fields/objects.py
```diff
@@ -0,0 +1,42 @@
...
+        LANG_EN_US_DISPLAY: ACTIVITY_QUESTION_DISPLAY
+    }
+    type: Literal[ACTIVITY_QUESTION_ID] = ACTIVITY_QUESTION_ID
+    extensions: Optional[ObjectDefinitionExtensionsField]
+
+
+class QuestionObjectField(ObjectField):
+    """Represents the `object` xAPI field for question saved xAPI statement.
+
+    WARNING: It doesn't include the optional `objectType` field.
```
jmaupetit, line 36: ditto

tests/models/edx/converters/xapi/test_open_response_assessment.py
```diff
@@ -0,0 +1,77 @@
...
+            saved_response=st.builds(
+                ORASaveSubmissionEventSavedResponseField,
+                parts=st.just([{"text": "first"}, {"text": "second"}]),
+            ),
+        ),
+    ),
+    provisional.urls(),
+)
+@pytest.mark.parametrize("uuid_namespace", ["ee241f8b-174f-5bdb-bae9-c09de5fe017f"])
+def test_open_response_assessment_ora_save_submission_to_quesiont_saved_conversion(
```
jmaupetit, line 44:
```suggestion
def test_open_response_assessment_ora_save_submission_to_question_saved_conversion(
```

tests/models/edx/converters/xapi/test_open_response_assessment.py
jmaupetit: It would be nice to have the original question! Should we assign a default language that could be overridden via a Ralph setting or a CLI option?

SergioSim: Good Idea) If we could filter out events that are related to courses that are in a specific language prior the convert command and then pass the right language option - this could work.
We could automatize this by querying the mongodb `modulestore.structures` and `modulestore.active_versions` collections to get a dictionary with all course_keys (org+course_id+session_id) associated with their language.

tests/models/edx/converters/xapi/test_open_response_assessment.py
SergioSim: Nice catch) Thanks)

src/ralph/models/edx/base.py
SergioSim: Ok) Thanks) Then just downgrading it to `AnyHttpUrl` to accept `localhost` as the referer too.

src/ralph/models/xapi/fields/verbs.py
SergioSim: Thanks) Indeed, this would be harder to maintain)

src/ralph/models/xapi/open_response_assessment/fields/objects.py
SergioSim: Good Idea) Updated this part) Now the base `ObjectField` model defines all optional fields and its children redefine some of the fields with additional constraints.

quitterie-lcs (commented):
If it is not too annoying for you, could you split the `converter` commit into two commits, one for the models, and one for the converter ?

src/ralph/models/xapi/video/fields/objects.py
```diff
@@ -1,39 +1,37 @@
...
 
-    name: Optional[dict[LANG_EN_US_DISPLAY, str]]
     type: VIDEO_OBJECT_DEFINITION_TYPE = VIDEO_OBJECT_DEFINITION_TYPE.__args__[0]
     extensions: Optional[ObjectDefinitionExtensionsField]
 
 
 class VideoObjectField(ObjectField):
     """Represents the `object` xAPI field for video statements.
 
+    WARNING: Contains an additional name property, this is not a violation of
```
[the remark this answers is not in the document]
SergioSim: Thanks)

tests/models/xapi/test_open_response_assessment.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@settings(max_examples=1)
+@given(
+    st.builds(
+        QuestionSaved,
+        actor=st.builds(
+            ActorField,
+            account=st.builds(
+                ActorAccountField, name=st.just("username"), homePage=provisional.urls()
```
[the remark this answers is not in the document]
SergioSim: Nice catch! I didn't consider this case but on some occasions, it is indeed possible to interact with openassessment questions without being logged in.  (probably used for third-party integrations).
Will update this.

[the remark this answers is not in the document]
SergioSim: Oh, actually the problem is different, hypothesis by default uses an empty string in this case which is invalid as xAPI statements should not have empty fields, and consequently, the model raises a ValidationError.
I propose to keep the default username as a workaround.

tests/models/xapi/test_open_response_assessment.py
```diff
@@ -0,0 +1,56 @@
...
+            extensions=st.builds(
+                QuestionSavedResultExtensionsField,
+                **{EXTENSION_RESPONSES_ID: st.just(["first", "second"])}
+            ),
+        ),
+    )
+)
+def test_models_xapi_ora_question_saved_selector_with_valid_statement(statement):
+    """Tests given a QuestionSaved statement, the get_model method should return
+    a QuestionSaved model."""
```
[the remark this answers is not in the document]
SergioSim: Thanks)

tests/models/xapi/fields/test_verbs.py
```diff
@@ -10,7 +10,7 @@ def test_models_xapi_fields_verb_page_viewed_verb_field():
 
     assert json.loads(ViewedVerbField().json()) == {
```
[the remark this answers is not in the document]
SergioSim: Well spotted! Thanks) updated this part)

tests/models/xapi/fields/test_verbs.py
```diff
@@ -19,5 +19,5 @@ def test_models_xapi_fields_verb_page_terminated_verb_field():
 
     assert json.loads(TerminatedVerbField().json()) == {
```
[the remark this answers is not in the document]
SergioSim: Thanks) updated)

SergioSim:
>  If it is not too annoying for you, could you split the `converter` commit into two commits, one for the models, and one for the converter ?

Good Idea) Thanks) updated the commit)

SergioSim:
Closed in favor of #226

PR #113, opened 2021-07-06
🐛(fixture) fix context management for gelf logger fixture

quitterie-lcs:
## Purpose

With the upgrade of pylint package, the stream handler requires to use a context.

## Proposal

It was ignored in the pull request #112 concerning the upgrade, and fix here, additionnally with yielding the logger.


jmaupetit (commented):
Wunderbar! Can you split your changes into two distinct commits?

jmaupetit (approved):

PR #114, opened 2021-07-09
✨(backends) add custom op_type support for es backend

jmaupetit:
## Purpose

Elasticsearch supports multiple operation types (aka `op_type`s) for bulk action requests: create, index, delete and update.

Using Elasticsearch new index templates requires that you perform "create" operations instead of the default "index" operation. We need more flexibility in Ralph.

## Proposal

We now support all operation types that can be specified as a backend option in the CLI commands that allow the usage of the elasticsearch backend:

```
--es-op-type [create|index|update|delete]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to
 - Install security updates in project Docker images
 - Model selector to retrieve associated pydantic model of a given event
 - `validate` command to lint edx events using pydantic models
+- Support all available bulk operation types for the elasticsearch backend
+  (create, index, update, delete) using the `--es-op-type` option
 
 ### Changed
 
--- a/src/ralph/backends/database/es.py
+++ b/src/ralph/backends/database/es.py
@@ -3,13 +3,28 @@
 import json
 import logging
 import sys
+from enum import Enum
 
 from elasticsearch import Elasticsearch
 from elasticsearch.helpers import scan, streaming_bulk
 
+from ralph.exceptions import BackendParameterException
+
 from .base import BaseDatabase
 
 logger = logging.getLogger(__name__)
+
+
+class OpType(Enum):
+    """Elasticsearch operation types"""
+
+    INDEX = "index"
+    CREATE = "create"
+    DELETE = "delete"
+    UPDATE = "update"
+
+
+DEFAULT_OP_TYPE = OpType.INDEX.value
 
 
 class ESDatabase(BaseDatabase):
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ and this project adheres to
 - Install security updates in project Docker images
 - Model selector to retrieve associated pydantic model of a given event
 - `validate` command to lint edx events using pydantic models
+- The elasticsearch backend now supports all available bulk operation types
```
quitterie-lcs, line 29: I would prefer an infinitive formulation in the Changelog. WDYT?

tests/backends/database/test_es.py
```diff
@@ -72,13 +86,48 @@ def test_backends_database_es_to_documents_method(es):
...
+
+def test_backends_database_es_to_documents_method_with_create_op_type(es):
+    """Tests to_documents method using the create op_type."""
+    # pylint: disable=invalid-name,unused-argument
+
+    # Create stream data
+    stream = StringIO("\n".join([json.dumps({"id": idx}) for idx in range(10)]))
+    stream.seek(0)
+
+    database = ESDatabase(
```
quitterie-lcs, line 114:
```suggestion
    database = ESDatabase(hosts=ES_TEST_HOSTS, index=ES_TEST_INDEX)
```

quitterie-lcs (approved):

src/ralph/backends/database/es.py
```diff
@@ -41,16 +69,30 @@ def get(self, chunk_size=500):
                 bytes(json.dumps(document.get("_source")) + "\n", encoding="utf-8")
             )
 
-    def to_documents(self, stream, get_id):
+    def to_documents(self, stream, get_id, op_type):
```
SergioSim, line 72:
```suggestion
    def to_documents(self, stream, get_id):
```
Maybe we could use `self.opt_type` ?

tests/backends/database/test_es.py
jmaupetit: Black forces me to do so :shrug:

CHANGELOG.md
jmaupetit: Better?

SergioSim (approved):
Looks good) Only one optional suggestion
image

src/ralph/backends/database/es.py
jmaupetit: You are right as we already depend on the class instance `index` attribute.

quitterie-lcs (approved):

PR #118, opened 2021-07-15
Fix elasticsearch datastream support

jmaupetit:
## Purpose

Pushing statements to an elasticsearch data stream fails due to index mapping inconsistency.

## Proposal

Removing the `"_type": "document"` to bulk actions seems to fix the issue.

Note that statements send to an elasticsearch data stream requires to name your timestamp field `@timestamp`.

### DIFF
--- a/tests/backends/database/test_es.py
+++ b/tests/backends/database/test_es.py
@@ -4,6 +4,7 @@ import json
 import random
 import sys
 from collections.abc import Iterable
+from datetime import datetime
 from io import BytesIO, StringIO
 from pathlib import Path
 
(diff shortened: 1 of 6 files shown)


tests/backends/database/test_es.py
```diff
@@ -314,3 +313,33 @@ def test_backends_database_es_put_with_badly_formatted_data_in_force_mode(
     assert sorted([hit["_source"]["id"] for hit in hits]) == [
         i for i in range(10) if i != 2
     ]
+
+
+def test_backends_database_es_put_with_datastream(es_data_stream, fs, monkeypatch):
+    """Test ES put method when using a configured data stream"""
```
SergioSim, line 319:
```suggestion
    """Tests ES put method when using a configured data stream."""
```

SergioSim (approved):
Looks good for me :+1: 
image

PR #121, opened 2021-07-20
Add websocket backend

jmaupetit:
## Purpose

Some logging tools provide websockets for real-time processing. This is a great added-value for Ralph to support consuming such streams.

## Proposal

- [x] add `ws` stream backend
- [x] install `jq` in ralph Docker image
- [x] allow to run a custom command in app deployment
- [x] add tests for the `ws` stream backend

### DIFF
--- a/src/ralph/backends/stream/ws.py
+++ b/src/ralph/backends/stream/ws.py
@@ -1,0 +1,38 @@
+"""Websocket stream backend for Ralph"""
+
+import asyncio
+import logging
+import sys
+
+import websockets
+
+from .base import BaseStream
+
+logger = logging.getLogger(__name__)
+
+
+class WSStream(BaseStream):
+    """Websocket stream backend."""
+
+    name = "ws"
+
+    def __init__(self, uri: str):
+        """Instantiate the websocket client.
+
+        uri: the URI to connect to
+        """
+
+        self.uri = uri
+
+    def stream(self):
+        """Stream websocket content to stdout."""
+        # pylint: disable=no-member
+
+        logger.debug("Streaming from websocket uri: %s", self.uri)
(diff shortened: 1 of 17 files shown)


src/ralph/backends/stream/ws.py
```diff
@@ -0,0 +1,37 @@
...
+logger = logging.getLogger(__name__)
+
+
+class WSStream(BaseStream):
+    """Websocket stream backend"""
+
+    name = "ws"
+
+    def __init__(self, uri: str):
+        """Instanciate the websocket client.
```
quitterie-lcs, line 20:
```suggestion
        """Instantiate the websocket client.
```

tests/backends/stream/test_base.py
```diff
@@ -0,0 +1,19 @@
...
+
+    class GoodStream(BaseStream):
+        """Correct implementation with required abstract methods."""
+
+        name = "good"
+
+        def stream(self):
+            """Fakes the stream method."""
+
+    GoodStream()
```
SergioSim, line 17: Is this to test that we don't get exceptions during instantiation ?
```suggestion
    try:
        GoodStream()
    except Exception:
        pytest.fail("A correct implementation of BaseStream should not raise exceptions")
```

quitterie-lcs (approved):

SergioSim (approved):
Looks good for me :+1:  Only some optional minor suggestions)
image

tests/backends/stream/test_base.py
jmaupetit: Yup. IMO it's overkill as it will fail with an explicit message mentioning that the abstract method should be implemented.

PR #125, opened 2021-08-03
Restructure edx models code

quitterie-lcs:
## Purpose

Based on the structure of xapi models implementation, the structure of edx models for already implemented events has been improved.

## Proposal

- [x] navigational events models
- [x] enrollment events models

### DIFF
--- a/tests/models/edx/test_navigational.py
+++ b/tests/models/edx/test_navigational.py
@@ -9,8 +9,8 @@ from hypothesis import strategies as st
 from pydantic.error_wrappers import ValidationError
 
 from ralph.models.edx.base import BaseContextField
-from ralph.models.edx.navigational import (
-    NavigationalEventField,
+from ralph.models.edx.navigational.fields.events import NavigationalEventField
+from ralph.models.edx.navigational.statements import (
     UIPageClose,
     UISeqGoto,
     UISeqNext,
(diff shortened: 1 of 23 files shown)


tests/models/edx/test_navigational.py
```diff
@@ -94,11 +123,33 @@ def test_models_edx_ui_seq_goto_selector_with_valid_event(event):
...
-    """Tests given a seq_next event the get_model method should return UISeqNext model."""
+def test_models_edx_ui_seq_next_with_valid_statement(statement):
+    """Tests that a `seq_next` statement has the expected `event_type`, `name`, `old` and `new`
+    from `event` field."""
 
-    event = json.loads(event.json())
-    assert ModelSelector(module="ralph.models.edx").get_model(event) is UISeqNext
+    assert statement.event_type == "seq_next"
+    assert statement.name == "seq_next"
+    assert statement.event.new == statement.event.old + 1
```
SergioSim, line 132:
```suggestion
```
As these values are preset in the test we don't need to test them ?

tests/models/edx/test_navigational.py
```diff
@@ -140,11 +191,33 @@ def test_models_edx_ui_seq_next_event_with_invalid_content(old, new, event):
...
 )
-def test_models_edx_ui_seq_prev_selector_with_valid_event(event):
-    """Tests given a seq_prev event the get_model method should return UISeqPrev model."""
+def test_models_edx_ui_seq_prev_with_valid_statement(statement):
+    """Tests that a `seq_prev` statement has the expected `event_type`, `name`, `old` and `new`
+    from `event` field."""
+
+    assert statement.event_type == "seq_prev"
+    assert statement.name == "seq_prev"
+    assert statement.event.new == statement.event.old - 1
```
SergioSim, line 200:
```suggestion
```
here too ?

SergioSim (approved):
LGTM) I like this new structure)
Have only a couple of suggestions and one minor concern - now in the edx module we use `event` and `statement` interchangeably which might be a bit confusing.
(ex. `edx.enrollment.statements.EdxCourseEnrollmentActivated(BaseServerEvent)`)
But I guess as the meaning is similar this should be OK)
image

quitterie-lcs:
@SergioSim Thanks for your review. I have updated with your proposition. 
About the `statement/server` issue, I cleaned up as much as I could or when I thought it was relevant to use `statement` rather than `event` (that is what we did for `xapi` ones). In the last commit, I have removed `event` suffix in model naming and just let the name of the model or with `model` suffix for `base` models (always to be coherent with `xapi` modelization). WDYTOI?

SergioSim:
> @SergioSim Thanks for your review. I have updated with your proposition.
> About the `statement/server` issue, I cleaned up as much as I could or when I thought it was relevant to use `statement` rather than `event` (that is what we did for `xapi` ones). In the last commit, I have removed `event` suffix in model naming and just let the name of the model or with `model` suffix for `base` models (always to be coherent with `xapi` modelization). WDYTOI?

@quitterie-lcs This makes it more consistent, thank you for going the extra mile !

PR #144, opened 2021-09-21
🔧(project) add 'run' Makefile rule

jmaupetit:
## Purpose

All our project have a default `make run` rule to start the development stack. This project seems to be an exception.

## Proposal

Elasticsearch is used by Ralph itself, but also grafana in the potsie project using the potsie network. This service should be targetted by the default `run` Makefile rule.


quitterie-lcs (approved):
Good idea!

PR #166, opened 2022-01-03
Bootstrap the fast API server

mbenadda:
## Purpose

Setup a CLI command `ralph runserver` that runs a development version of our FastAPI based LRS API. Implement HTTP Basic auth through a hardcoded list of users and scopes.

## Proposal

- [x] Set up the fastAPI server with a shell command
- [x] Add authentification through a middleware based on a user credentials file
- [x] Make the user credentials file into a command line parameter
- [x] Add unit tests for authentication & authorization

### Credential files (updated)

User credentials file in `.ralph/auth.json` (gitignored directory):

```json
[
  {
    "username": "mehdi",
    "hash": "$2b$12$yBXrzIuRIk6yaft5KUgVFOIPv0PskCCh9PXmF2t7pno.qUZ5LK0D2",
    "scopes": ["mehdi_scope"]
  }
]

```

### Authentication errors

Right now there is no HTTP Basic Auth prompt. I could probably [tweak authentication middleware error responses](https://www.starlette.io/authentication/#custom-authentication-error-responses) to add it but since the API server is expected to be called programmatically I'm not sure if we want to add them.

Also all authentication errors are 400 errors right now (as it's the default behavior for the middleware). I can use the same autentication middleware customization to return a different error code but I'm not sure it's possible to discriminate between several different HTTP error codes.


src/ralph/api/auth.py
```diff
@@ -0,0 +1,78 @@
...
+            scheme, credentials = auth.split()
+            if scheme.lower() != "basic":
+                raise AuthenticationError("Missing authentication credentials.")
+            decoded = base64.b64decode(credentials).decode("ascii")
+        except (ValueError, UnicodeDecodeError, binascii.Error) as exc:
+            raise AuthenticationError("Invalid authentication credentials.")
+
+        # Split username/password and check them against our stored user credentials
+        username, _, password = decoded.partition(":")
+        user_info = get_stored_credentials().get(username, {})
```
jmaupetit, line 64: AUTH_FILE should be defined in `ralph.defaults` using the `config` utility:

```python
# ralph/defaults.py
AUTH_FILE = Path(config("RALPH_AUTH_FILE", APP_DIR / "auth.json"))
```

And I would prefer the auth file to be a list of user entries such as:

```json
[
  {"username": "foo"},
  {"username": "bar"}
]
```

```suggestion
        user_info = filter(
            lambda u: u.get("username") == username, 
            get_stored_credentials(AUTH_FILE)
        )
```

.gitignore
```diff
@@ -36,6 +36,9 @@ ENV/
 env.bak/
 venv.bak/
 
+# Development credentials
+credentials/
```
jmaupetit, line 40: I would prefer this file to stand in the application directory (see comments below).

src/ralph/cli.py
```diff
@@ -393,3 +397,52 @@ def list_(details, new, backend, **options):
...
+    help=f"Serve the LRS-compatible API on this port (default {DEFAULT_RUNSERVER_PORT})",
+)
+@click.option(
+    "-f",
+    "--auth-file",
+    type=str,
+    default="credentials/users.json",
+    help="File containing authentification & authorization information for API calls",
+)
+def runserver(host, port, auth_file):
```
jmaupetit, line 427: IMO making this path configurable using the environment or a config file with a relevant default is sufficient (using the `config` utility).

```suggestion
def runserver(host, port):
```

src/ralph/cli.py
```diff
@@ -393,3 +397,52 @@ def list_(details, new, backend, **options):
...
+    logger.info(
+        (
+            "Do not use runserver in production - start production servers "
+            "through a process manager such as gunicorn/supervisor/circus."
+        )
+    )
+    try:
+        # NB: we must pass the live app in the same thread to uvicorn.run so it can access
+        # the current click context and make use of CLI parameters.
+        # This also prevents us from enabling live-reloading.
```
jmaupetit, line 442: This should no longer be the case, right?

mbenadda: Yes, this is not necessary if we do not need to access CLI params from inside the server. For now there was only `auth_file` so we can use an import string dans enable auto reloading 👍

bin/ralph
```diff
@@ -9,4 +9,4 @@ if [ "$1" == "--no-deps" ]; then
     shift
 fi
 
-DOCKER_USER=${DOCKER_USER} docker-compose run --rm -T ${NO_DEPS} app ralph ${@}
+DOCKER_USER=${DOCKER_USER} docker-compose run --service-ports --rm -T ${NO_DEPS} app ralph ${@}
```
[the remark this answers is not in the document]
jmaupetit: You are totally right. I should invest why.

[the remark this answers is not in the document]
mbenadda: I think it's the expected behavior (see [the docs](https://docs.docker.com/compose/reference/run/))

src/ralph/defaults.py
```diff
@@ -107,6 +107,8 @@ def load_config(config_file_path):
 
 APP_DIR = Path(environ.get("RALPH_APP_DIR", get_app_dir("ralph")))
 DEFAULT_ENCODING = getattr(io, "LOCALE_ENCODING", "utf8")
+DEFAULT_RUNSERVER_HOST = "0.0.0.0"
+DEFAULT_RUNSERVER_PORT = 8100
```
quitterie-lcs, line 111: Why not removing the default value in `click` and add a config file to those variable? Hence, they can also be set in the `env` file.

mbenadda: Removing the default value? You mean removing the options from click and just using the environment for this?

jmaupetit (commented):
Awesome! I only have minor comments and two requests:

1. Can you update the PR description concerning expected credentials file format?
2. Can you start a documentation for this server (e.g. `docs/api.md`)? For now, it would be a getting started for server usage (at least credentials generation and performing requests using a CLI such as HTTPie or Curl).

jmaupetit:
And I think you can explicitly ignore the bandit linter warning as it is expected!

src/ralph/api/auth.py
```diff
@@ -0,0 +1,102 @@
...
+    async def authenticate(self, conn: Request):
+        """
+        Get the basic auth parameters from the Authorization header, and check them
+        against our own list of hashed credentials.
+        """
+        if "Authorization" not in conn.headers:
+            raise AuthenticationError("Missing authentication credentials.")
+
+        # Extract basic auth credentials from the relevant header
+        auth = conn.headers["Authorization"]
```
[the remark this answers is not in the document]
mbenadda: In this instance I was complying with `pylint`, as the argument is named `conn` in the base class. I agree with you that request is way better and I'll just overrule `pylint` 👍

mbenadda:
I added the docs page and moved the host/port params to be config options instead of CLI flags.

I have a problem with `lint-git` that warns of a `print()` statement in the docs. I think we can just merge it anyway as it should not be in the diff anymore and therefore not cause issues with downstream commits.

docs/api.md
```diff
@@ -0,0 +1,63 @@
...
+
+```bash
+$ bin ralph runserver
+```
+
+Before you can start your API server and make requests against it, you need to set up your credentials.
+
+### Creating a credentials file
+
+The credentials file is expected to be a valid JSON file. Its location is specified by the `RALPH_AUTH_FILE` configuration value. By default, `ralph` will look for it at `.ralph/auth.json`.
```
jmaupetit, line 17:
```suggestion
The credentials file is expected to be a valid JSON file. Its location is specified by the `RALPH_AUTH_FILE` configuration value. By default, `ralph` will look for the `auth.json` file in the application directory (see [click documentation for details](https://click.palletsprojects.com/en/8.0.x/api/#click.get_app_dir)).
```

jmaupetit (commented):
Yay, thanks!

Yup we will ignore the linter warning due to the `print` statement in the documentation.

docs/api.md
```diff
@@ -0,0 +1,63 @@
...
+
+The expected format is a list of entries (JSON objects) each containing the username, `bcrypt` hashed+salted password and scopes the relevant user can access:
```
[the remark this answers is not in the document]
quitterie-lcs: `Username` is correct I think

docs/api.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+# Install bcrypt
+$ python3 -m pip install bcrypt
+# Generate the hash and print it to console
+$ python3 -c 'import bcrypt; print(bcrypt.hashpw(b"PASSWORD", bcrypt.gensalt()).decode("ascii"))'
+```
+
+### Making a GET request
+
+The first request that can be answered by the ralph API server is a `whoami` request, which checks the user is authenticated and returns their username and permission scopes.
```
quitterie-lcs, line 47:
```suggestion
The first request that can be answered by the ralph API server is a `whoami` request, which checks if the user is authenticated and returns their username and permission scopes.
```

docs/api.md
```diff
@@ -0,0 +1,63 @@
...
+< {"error":"Missing authentication credentials."}
+```
+
+Send your username and password to the API server through HTTP Basic Auth:
+
+```bash
+$ curl --user john.doe@example.com:PASSWORD http://localhost:8100/whoami
+< HTTP/1.1 200 OK
+< {"username":"john.doe@example.com","scopes":["authenticated","example_scope"]}
+```
```
jmaupetit, line 63: I am sorry I forgot one last thing: you need to refer to this documentation in the `mkdocs.yml` file (add an item in the `nav` list).

jmaupetit (approved):

PR #171, opened 2022-01-24
♻️(api) rework authentication to use a fastapi dependency

mbenadda:
## Purpose

As we built authentication into our Ralph API, we looked for auth middleware and found built-in, compatible middleware from Starlette that was explicitly compatible with FastAPI.

We did not know at that time that there are incompatibilities with this specific middleware that prevent us from doing any sort of authorization work with FastAPI, like this middleware does if used directly with Starlette.

See https://github.com/tiangolo/fastapi/issues/104

## Proposal

We moved on to refactor our authentication to use a FastAPI dependency instead of a middleware.

As we start actually using authorization scopes, we can build another layer of FastAPI dependency on top of our `authenticated_user` to require certain scopes to access a given resource.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -26,35 +26,38 @@ STORED_CREDENTIALS = json.dumps(
 
 def test_get_whoami_no_credentials():
     """
-    whoami route returns a 400 error when no credentials are sent.
+    whoami route returns a 401 error when no credentials are sent.
     """
     response = client.get("/whoami")
-    assert response.status_code == 400
-    assert response.json() == {"error": "Missing authentication credentials."}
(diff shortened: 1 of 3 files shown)


tests/test_auth.py
```diff
@@ -29,26 +29,29 @@ def test_get_whoami_no_credentials():
     whoami route returns a 400 error when no credentials are sent.
```
quitterie-lcs, line 29:
```suggestion
    whoami route returns a 401 error when no credentials are sent.
```
And the next docstrings for which the test assertion has been updated with a `401` code error

jmaupetit (approved):

PR #175, opened 2022-02-14
Implement the LRS GET Statement API endpoint

mbenadda:
## Purpose

Implement the LRS GET Statement API endpoint


mbenadda:
Archiving this attempt at implementing the LRS spec as we chose to go with a quicker, partial implementation first with #176.

PR #176, opened 2022-02-15
✨(elasticsearch) build a quick partial implementation of LRS GET

mbenadda:
## Purpose

We're quickly building a partial implementation of the LRS specification to experiment with customers. With the additional
constraint of not being able to build our own index for search purposes, we're just doing what we can do with the basic existing shape of the documents.

We're documenting all parameters for GET, specifying which ones are actually supported or not in the current version.

## Proposal

- [x] implement a quick and partial version of LRS GET
- [x] document all query parameters (specifying which ones are unsupported)


mbenadda:
@jmaupetit I updated the PR with proper serialization of results and consistent-over-requests pagination

src/ralph/api/routers/statements.py
```diff
@@ -0,0 +1,244 @@
...
+        path = request.url.components.path
+        query = request.url.components.query
+        response.update(
+            {
+                "more": (
+                    f"{path}{query + '&' if query else '?'}pit_id={pit_id}"
+                    f"&search_after={'|'.join(last_hit_sort)}"
+                )
+            }
+        )
```
jmaupetit, line 237: I don't see pagination management in the response when the number of statements exceeds the limit?

src/ralph/api/routers/statements.py
```diff
@@ -0,0 +1,244 @@
...
+        size=limit,
+    )
+
+    # Prepare the link to get the next page of requests, while preserving the
+    # consistency of search results.
+    # NB: There is an unhandled edge case where the total number of results is
+    # exactly a multiple of the "limit", in which case we'll offer an extra page
+    # with 0 results.
+    response = {}
+    if len(es_response["hits"]["hits"]) == limit:
```
jmaupetit, line 225: The number of hits is in the ES response right?

```suggestion
    if es_response["hits"]["total"]["value"] == limit:
```

jmaupetit (commented):
Amazing start, thanks!

For more efficiency, in a near future, I think we should support the async ES client and stream the response instead.

src/ralph/api/routers/statements.py
mbenadda: The number of statements cannot exceed limit by definition.

Either it is the same if there are more results than the `limit` number (because limit meant we only got `limit` results, eg `1000` results out of `10000000`), or it is smaller if there  are fewer than `limit` results (eg `limit` is `1000` and there are `300` results).

src/ralph/api/routers/statements.py
mbenadda: So `es_response["hits"]["total"]["value"]` is not very useful in my opinion: it (pretends to) represents the total number of hits, not the number of hits contained in the current response.

I say `pretends to` because it is capped at an arbitrary number defined by settings (which is 10000 by default and is not recommended to be increased to a large number).


If we want to better manage pagination, IMO, the way is to use a custom aggregation that simply counts the total number of matching documents. I did not pursue this as the current implementation (only having a minor issue in an edge case) seemed good enough to me in the context of this POC.

src/ralph/api/routers/statements.py
```diff
@@ -0,0 +1,180 @@
+"""API routes related to statements."""
+
+from datetime import datetime
+from typing import Literal, Optional
+from uuid import UUID
+
+from elasticsearch import Elasticsearch  # NB: should be AsyncElasticsearch
```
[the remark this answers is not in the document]
mbenadda: I added this comment when I did not manage to import `AsyncElasticsearch`. I later found out we just need to add an optional dependency for this.

src/ralph/api/routers/statements.py
```diff
@@ -0,0 +1,244 @@
...
+    if since:
+        es_query_filters += [{"range": {"timestamp": {"gt": since}}}]
+
+    if until:
+        es_query_filters += [{"range": {"timestamp": {"lte": until}}}]
+
+    if len(es_query_filters) > 0:
+        es_query = {"query": {"bool": {"filter": es_query_filters}}}
+    else:
+        es_query = {"query": {"match_all": {}}}
```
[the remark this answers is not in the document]
mbenadda: I can make this change, but is there a particular reason for it? It seems less clear to me this way.

src/ralph/api/routers/statements.py
```diff
@@ -0,0 +1,244 @@
...
+        description=(
+            "Only Statements stored at or "
+            "before the specified Timestamp are returned"
+        ),
+    ),
+    limit: Optional[int] = Query(
+        ES_MAX_SEARCH_HITS_COUNT,
+        description=(
+            "Maximum number of Statements to return. "
+            "0 indicates return the maximum the server will allow"
```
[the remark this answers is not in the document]
mbenadda: This is from the LRS spec: `0` is just a conventional value for "no limit", which in practice will just use the maximum value allowed by the server.

src/ralph/api/routers/statements.py
mbenadda: That's also how I use `limit`. Therefore a page can only ever contain `limit` (eg. 1000) statements or less, never more.

Then I use this to decide if I should include `"more"` in the response:  if the number of results I'm getting is less than `limit`, that means I've reached the last page of results.

If there are 6450 results, the first 6 pages will have 1000 results, and the last one 450: when I get 450 results, I know I've reached the end of my pagination — and when I have a page with 1000 results, I'm assuming I get 1000 because I have more than 1000 results remaining. Of course, if there are exactly 6000 results this breaks down and I'll get 6 pages with 1000 results and one page with zero result.

jmaupetit: Oh, thanks for the clear explanation: I finally see the light!

I would have the opposite approach that adds the `more` field by default unless if `len(statements) < limit`, I find it easier to understand. WDYT?

jmaupetit (approved):

quitterie-lcs (commented):
Only have one question: why not using `hypothesis` to generate statements in the tests?

mbenadda:
So I have looked into it and I'm not sure how appropriate `hypothesis` is for API tests (see https://hypothesis.readthedocs.io/en/latest/examples.html#fuzzing-an-http-api), especially as we need tight control of the mock data to test our search and sorting parameters.

In any case, I'd rather merge this one so we can move on, and we can always look into testing the API with `hypothesis` later on.

PR #181, opened 2022-03-15
✨(elasticsearch) build an implementation of the LRS POST endpoint

mbenadda:
## Purpose

As we built the GET statements endpoint to access statements in Elasticsearch, we need to build the POST endpoint to provide a way to put statements there using the LRS-compliant API.

Fow now we only support POSTing lists of statements to add, and not the other use cases for POST statements, such as secondary PUT (one statement) and secondary GET (passing GET parameters in the request body).


tests/api/test_statements_post.py
```diff
@@ -0,0 +1,229 @@
...
+
+    response = client.post(
+        "/xAPI/statements/",
+        headers={"Authorization": f"Basic {credentials}"},
+        json=statements,
+    )
+
+    assert response.status_code == 200
+    assert response.json()[0] == statements[0]["id"]
+    assert isinstance(response.json()[1], str)
```
quitterie-lcs, line 145: Do we know if it is only a random string ? Or is it an UUID?

quitterie-lcs (commented):
Nice job ! 
Only have one minor question...

tests/api/test_statements_post.py
mbenadda: It should be a UUID in string format. Maybe you're suggesting we check that with a regex?

jmaupetit: Yup, it should be!

src/ralph/api/routers/statements.py
```diff
@@ -254,3 +255,77 @@ async def get(
...
+    LRS Specification:
+    https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Communication.md#212-post-statements
+    """
+
+    # The statements dict has multiple functions:
+    # - generate IDs for statements that are missing them;
+    # - use the list of keys to perform validations and as a final return value;
+    # - provide an iterable containing both the statements and generated IDs for bulk.
+    statements_dict = {
+        str(statement.id) if statement.id else str(uuid4()): statement
```
jmaupetit, line 291: Shouldn't we reject statements without IDs instead?

src/ralph/api/routers/statements.py
```diff
@@ -254,3 +255,77 @@ async def get(
...
+        [
+            {
+                "_index": "statements",
+                "_id": statement_id,
+                "_op_type": "index",
+                "_source": statement.json(),
+            }
+            for statement_id, statement in statements_dict.items()
+        ],
+    )
```
jmaupetit, line 329: Maybe we should provide more details to ease debug, WDYT?

```suggestion
    try:
        actions, errors = bulk(
            ES_CLIENT,
            [
                {
                    "_index": "statements",
                    "_id": statement_id,
                    "_op_type": "index",
                    "_source": statement.json(),
                }
                for statement_id, statement in statements_dict.items()
            ],
        )
    except BulkIndexError:
        logger.error("Failed to index submitted statements")
        logger.debug("Error was: %s", str(errors))
        raise HTTPException(status_code=500, detail="Statements bulk indexation failed")
        
    logger.info("Indexed %d statements with success", actions)
```

tests/api/test_statements_post.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+@pytest.fixture(autouse=True)
+def teardown_es_index():
+    """
+    Clean up Elasticsearch after each test.
+    """
+    ES_INDICES_CLIENT.create(ES_TEST_INDEX)
+    yield
+    ES_INDICES_CLIENT.delete(ES_TEST_INDEX)
```
jmaupetit, line 61: Why not using the `es` fixture [1] here?

[1] https://github.com/openfun/ralph/blob/master/tests/fixtures/backends.py#L49

tests/api/test_statements_post.py
```diff
@@ -0,0 +1,229 @@
...
+                {
+                    "username": "ralph",
+                    "hash": bcrypt.hashpw(b"admin", bcrypt.gensalt()).decode("UTF-8"),
+                    "scopes": ["ralph_test_scope"],
+                }
+            ]
+        ),
+    )
+
+    return credentials
```
jmaupetit, line 51: As it is used in the GET statements endpoint tests, we can make a fixture instead of duplicating this function?

jmaupetit (commented):
Great POC! :pray:

src/ralph/api/routers/statements.py
jmaupetit: Well spotted, thanks!

src/ralph/api/models.py
```diff
@@ -0,0 +1,61 @@
...
+
+class LaxVerbField(BaseModelWithLaxConfig):
+    """
+    Lightest definition of a verb field compliant to the specification.
+    """
+
+    id: AnyUrl
+
+
+class LaxStatement(BaseModelWithLaxConfig):
```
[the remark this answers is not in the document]
jmaupetit: Thank you for this @SergioSim :pray:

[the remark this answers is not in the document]
mbenadda: It should be possible, with the caveat that AFAIK we may not want to perform complete validation of all incoming statements as that might be a performance hit when adding large quantities of statements.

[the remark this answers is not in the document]
jmaupetit: Indeed, we need to assess the impact on using `BaseXapiModel` vs `LaxStatement`! We can also work on togglable validators (via a global setting).

SergioSim (commented):
Awesome!
Looks good to me)
Just regarding the failing ci-build, could you try to add these changes: [update mkdocs](https://github.com/openfun/ralph/pull/182/commits/663f15d1c685f5bc9cd246e46f0ca5745c2a55a6) and this [line in the Dockerfile](https://github.com/openfun/ralph/pull/182/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R5) to check whether this makes the build pass?

jmaupetit:
I've fixed the CI and various dependency breaks in my latest PR, see #183 We should be fine now.

Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM python:3.9-slim as base
 
 # Upgrade pip to its latest release to speed up dependencies installation
-RUN pip install --upgrade pip
+RUN pip install --upgrade pip setuptools
```
jmaupetit, line 5: I think you could safely remove this commit once rebased onto main.

src/ralph/api/routers/statements.py
```diff
@@ -254,3 +259,93 @@ async def get(
...
+    statements_ids = list(statements_dict.keys())
+    if len(statements) != len(statements_ids):
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Duplicate statement IDs in the list of statements",
+        )
+
+    # pylint: disable=unexpected-keyword-arg
+    es_response = ES_CLIENT.search(
+        index=ES_INDEX, body={"query": {"terms": {"_id": statements_ids}}}
```
jmaupetit, line 312: Does the following avoid having to disable the `unexpected-keyword-arg` pylint warning?
```suggestion
        index=ES_INDEX, query={"terms": {"_id": statements_ids}}
```

tests/api/test_statements_post.py
```diff
@@ -0,0 +1,224 @@
...
+    response = client.post(
+        "/xAPI/statements/",
+        headers={"Authorization": f"Basic {auth_credentials}"},
+        json=statement,
+    )
+
+    assert response.status_code == 200
+    assert response.json() == [statement["id"]]
+    ES_INDICES_CLIENT.refresh()
+    hits = ES_CLIENT.search(index=ES_TEST_INDEX)["hits"]["hits"]
```
jmaupetit, line 52: The `es` fixture is an elasticsearch client instance, so you can use it as follow:

```suggestion
    es.indices.refresh(index=ES_TEST_INDEX)
    hits = es.search(index=ES_TEST_INDEX)["hits"]["hits"]
```

tests/api/test_statements_post.py
```diff
@@ -0,0 +1,224 @@
...
+from elasticsearch import Elasticsearch
+from elasticsearch.client import IndicesClient
+from fastapi.testclient import TestClient
+
+from ralph.api import app
+
+from tests.fixtures.backends import ES_TEST_HOSTS, ES_TEST_INDEX
+
+ES_CLIENT = Elasticsearch(ES_TEST_HOSTS)
+ES_INDICES_CLIENT = IndicesClient(ES_CLIENT)
```
jmaupetit, line 17: You can safely remove those.

tests/api/test_statements_post.py
jmaupetit: The same pattern applies for subsequent tests.

jmaupetit (commented):
I only have minor comments. After a final rebase, we should be good yo go! :pray:

src/ralph/api/routers/statements.py
mbenadda: `pylint` also complains about the `query` argument. I'm not sure if its a problem with `pylint` or `python-elasticsearch` but some arguments are just not getting picked up by `pylint`.

jmaupetit (approved):

PR #184, opened 2022-04-15
🎨(backends) make backends yield results instead of writing to stdout

SergioSim:
## Purpose

We want to make our backends more usable as a library.

## Proposal

Make backends yield results instead of writing to stdout.
Except for the websocked backend, which now expects a file-like object as an argument, which is then used to write the results.

- [x] make backends yield results
- [ ] add a query method for database backends
- [ ] ease database client instantiation / configuration


src/ralph/backends/database/base.py
```diff
@@ -1,17 +1,18 @@
...
 
     @abstractmethod
     def get(self, chunk_size=10):
-        """Read chunk_size records and stream them to stdout"""
+        """Reads `chunk_size` records from the database and yields them."""
 
     @abstractmethod
-    def put(self, chunk_size=10, ignore_errors=False):
-        """Write chunk_size records from stdin"""
+    def put(self, stream: Iterable, chunk_size=10, ignore_errors=False):
```
jmaupetit, line 17: Shouldn't we be more specific?

```suggestion
    def put(
        self,
        stream: Union[BinaryIO, TextIO],
        chunk_size: int = 10,
        ignore_errors: bool = False,
    ):
```

src/ralph/backends/database/base.py
```diff
@@ -1,17 +1,18 @@
...
 
 
 class BaseDatabase(ABC):
-    """Base database backend interface"""
+    """Base database backend interface."""
 
     name = "base"
 
     @abstractmethod
     def get(self, chunk_size=10):
```
jmaupetit, line 13:
```suggestion
    def get(self, chunk_size: int = 10):
```

src/ralph/backends/database/es.py
```diff
@@ -80,21 +77,13 @@ def to_documents(self, stream, get_id):
...
-                    {
-                        "_source": item,
-                    }
-                )
+                action.update({"_source": item})
             yield action
 
-    def put(self, chunk_size=500, ignore_errors=False):
-        """Write documents streamed from the standard input to the instance index"""
+    def put(self, stream, chunk_size=500, ignore_errors=False):
```
jmaupetit, line 122: And thus :point_down: 

```suggestion
    def put(
        self,
        stream: TextIO,
        chunk_size: int = 500,
        ignore_errors: bool = False,
    ):
```

src/ralph/backends/database/es.py
```diff
@@ -62,15 +61,13 @@ def __init__(
         self.op_type = op_type
 
     def get(self, chunk_size=500):
```
jmaupetit, line 69:
```suggestion
    def get(self, chunk_size: int = 500):
```

src/ralph/backends/database/es.py
```diff
@@ -62,15 +61,13 @@ def __init__(
...
-        """Get index documents and stream them"""
+        """Gets index documents and yields them."""
 
         for document in scan(self.client, index=self.index, size=chunk_size):
-            sys.stdout.buffer.write(
-                bytes(json.dumps(document.get("_source")) + "\n", encoding="utf-8")
-            )
+            yield document
 
     def to_documents(self, stream, get_id):
```
jmaupetit, line 106: Should we also type this function signature?

jmaupetit: And also the type of yielded objects?

src/ralph/cli.py
```diff
@@ -38,12 +38,11 @@
...
-DATABASE_BACKENDS = (lambda: [backend.value for backend in DatabaseBackends])()
-PARSERS = (lambda: [parser.value for parser in Parsers])()
-STORAGE_BACKENDS = (lambda: [backend.value for backend in StorageBackends])()
-STREAM_BACKENDS = (lambda: [backend.value for backend in StreamBackends])()
-BACKENDS = (lambda: DATABASE_BACKENDS + STORAGE_BACKENDS + STREAM_BACKENDS)()
+DATABASE_BACKENDS = [backend.value for backend in DatabaseBackends]
+PARSERS = [parser.value for parser in Parsers]
+STORAGE_BACKENDS = [backend.value for backend in StorageBackends]
+STREAM_BACKENDS = [backend.value for backend in StreamBackends]
+BACKENDS = DATABASE_BACKENDS + STORAGE_BACKENDS + STREAM_BACKENDS
```
jmaupetit, line 45: Why removing lazy evaluation?

src/ralph/cli.py
```diff
@@ -310,11 +309,14 @@ def fetch(backend, archive, chunk_size, **options):
     backend_type = get_backend_type(backend_class)
 
     if backend_type == BackendTypes.STORAGE:
-        backend.read(archive, chunk_size=chunk_size)
+        for data in backend.read(archive, chunk_size=chunk_size):
+            click.echo(data, nl=False)
     elif backend_type == BackendTypes.DATABASE:
-        backend.get(chunk_size=chunk_size)
+        for document in backend.get(chunk_size):
+            source = json.dumps(document.get("_source"))
```
jmaupetit, line 316: This will only work for the ES backend right?

tests/backends/storage/test_ldp.py
```diff
@@ -449,9 +439,7 @@ class MockStdout:
         }
     ]
 
-    mock_stdout.buffer.seek(0)
-    with gzip.open(mock_stdout.buffer, "rb") as output:
-        assert json.loads(output.read()) == archive_content
+    assert json.loads(gzip_decode(result)) == archive_content
```
jmaupetit, line 442: :1st_place_medal:

jmaupetit (commented):
Wunderbar!

src/ralph/backends/database/base.py
SergioSim: Good point) Thanks)

src/ralph/backends/database/base.py
SergioSim: Thanks)

src/ralph/backends/database/es.py
SergioSim: Good Idea) Thanks) Added the type hints)

src/ralph/cli.py
SergioSim: I'm not very sure, but I would like to question whether these are lazy evaluations.
Aren't the `()` at the end eagerly evaluating the lambda functions?

jmaupetit: Outch. Yes. I have a better suggestion, WDYT?

```suggestion
DATABASE_BACKENDS = map(lambda b: b.value, DatabaseBackends)
PARSERS = map(lambda p: p.value, Parsers)
STORAGE_BACKENDS = map(lambda b: b.value, StorageBackends)
STREAM_BACKENDS = map(lambda b: b.value, StreamBackends)
BACKENDS = DATABASE_BACKENDS + STORAGE_BACKENDS + STREAM_BACKENDS
```

SergioSim: I like the Idea to use `map` instead) Thanks)
Just wrapped it inside a lambda to be able to iterate over it multiple times.
However, pylint now complains about assigning lambda functions to variables.
Updated this part [in the last commit](https://github.com/openfun/ralph/pull/184/commits/4896806bd4aaddf541178eca94e74e73e169ad5b).

src/ralph/cli.py
jmaupetit: :thinking: After thought: I would prefer streaming hits as is and use a dedicated parser with the `extract` command instead. I have the feeling that we will end up with unmaintainable spaghetti code if we use this pattern. WDYT?

src/ralph/cli.py
```diff
@@ -38,12 +39,13 @@
...
-# Lazy evaluations
-DATABASE_BACKENDS = (lambda: [backend.value for backend in DatabaseBackends])()
-PARSERS = (lambda: [parser.value for parser in Parsers])()
-STORAGE_BACKENDS = (lambda: [backend.value for backend in StorageBackends])()
-STREAM_BACKENDS = (lambda: [backend.value for backend in StreamBackends])()
-BACKENDS = (lambda: DATABASE_BACKENDS + STORAGE_BACKENDS + STREAM_BACKENDS)()
+DATABASE_BACKENDS = lambda: map(lambda b: b.value, DatabaseBackends)  # noqa: E731
+PARSERS = lambda: map(lambda p: p.value, Parsers)  # noqa: E731
+STORAGE_BACKENDS = lambda: map(lambda b: b.value, StorageBackends)  # noqa: E731
+STREAM_BACKENDS = lambda: map(lambda b: b.value, StreamBackends)  # noqa: E731
```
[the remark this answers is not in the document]
SergioSim: Yes, however, maps are exhausted once iterated over.
For example:
```python
>>> a = map(lambda x: x, [1, 2, 3])
>>> for i in a:
...     print(i)
... 
1
2
3
>>> for i in a:
...     print(i)
... 
>>> 
```
One solution is to wrap `map` it inside a lambda to get a new generator for each iteration.

src/ralph/cli.py
SergioSim: Do we replace:
```python
source = json.dumps(document.get("_source"))
```
With:
```python
source = json.dumps(document)
```
And create an `ElasticSearchParser` class to do the `.get("_source")` part ?
Or do we change the `ESDatabase.get` method to return strings instead of objects?
Or maybe we could add a switch argument like `to_string=True` in the `ESDatabase.get` method to benefit both formats (string and object)?
Sorry, I have more questions than answers :sweat_smile:

src/ralph/cli.py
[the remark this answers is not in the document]
jmaupetit: You are right, I missed that point. But using lambda in here looks like anti-pattern. Should we define our lazy list object here?

src/ralph/cli.py
jmaupetit: IMO a backend's `get` method should return raw records, and we should use parser to get core (xAPI) events.

SergioSim: Ok, updated this part to make the `fetch` command return the raw ES records and added an `ElasticsearchParser` to extract the `_source` field.

src/ralph/backends/storage/base.py
```diff
@@ -1,25 +1,26 @@
...
-        """List files in the storage backend"""
+        """Lists files in the storage backend."""
 
     @abstractmethod
     def url(self, name):
-        """Get `name` file absolute URL"""
+        """Gets `name` file absolute URL."""
 
     @abstractmethod
     def read(self, name, chunk_size=4096):
```
quitterie-lcs, line 21:
```suggestion
    def read(self, name, chunk_size: int = 4096):
```

src/ralph/backends/storage/fs.py
```diff
@@ -69,18 +68,18 @@ def list(self, details=False, new=False):
             yield self._details(archive) if details else archive
 
     def url(self, name):
-        """Get `name` file absolute URL"""
+        """Gets `name` file absolute URL."""
 
         return str(self._get_filepath(name).resolve(strict=True))
 
     def read(self, name, chunk_size=4096):
```
quitterie-lcs, line 75:
```suggestion
    def read(self, name, chunk_size: int = 4096):
```

src/ralph/backends/stream/ws.py
```diff
@@ -17,22 +17,23 @@ class WSStream(BaseStream):
...
+        Args:
+            uri (str): The URI to connect to.
         """
 
         self.uri = uri
 
-    def stream(self):
-        """Stream websocket content to stdout."""
+    def stream(self, target: BinaryIO):
+        """Stream websocket content to target."""
```
quitterie-lcs, line 29:
```suggestion
        """Streams websocket content to target."""
```

quitterie-lcs (commented):
Good job! :muscle:

src/ralph/backends/storage/base.py
SergioSim: Well spotted) Thanks)

src/ralph/backends/stream/ws.py
SergioSim: Thanks)

src/ralph/parsers.py
```diff
@@ -40,29 +39,55 @@ def parse(self, input_file):
         """Parses GELF formatted logs (one JSON string event per row).
 
         Args:
-            input_file (string): Path to the log file to parse (could be gunzipped).
+            input_file (file-like): The log file to parse.
 
         Yields:
-            event: events raw short_message string
-
+            event: events raw short_message string.
```
quitterie-lcs, line 45:
```suggestion
            event: Events raw short_message string.
```

src/ralph/parsers.py
```diff
@@ -40,29 +39,55 @@ def parse(self, input_file):
...
                 logger.debug("Raised error was: %s", err)
             except KeyError as err:
-                logger.error(
-                    (
-                        "Input event '%s' doesn't comply with GELF format! It will be "
-                        "ignored."
-                    ),
-                    event,
+                msg = (
+                    "Input event '%s' doesn't comply with GELF format! "
```
quitterie-lcs, line 59:
```suggestion
                    "Input event '%s' doesn't comply with GELF format!\n"
```

src/ralph/parsers.py
```diff
@@ -40,29 +39,55 @@ def parse(self, input_file):
...
 
         for event in input_file:
             try:
                 yield json.loads(event)["short_message"]
             except (json.JSONDecodeError, TypeError) as err:
-                logger.error(
-                    "Input event '%s' is not a valid JSON string! It will be ignored.",
-                    event,
-                )
+                msg = "Input event '%s' is not a valid JSON string! It will be ignored."
```
quitterie-lcs, line 54:
```suggestion
                msg = (
                    "Input event '%s' is not a valid JSON string!\n"
                    "It will be ignored."
                )
```

src/ralph/parsers.py
SergioSim: Well spotted) Thanks)

src/ralph/api/routers/statements.py
```diff
@@ -323,29 +300,13 @@ async def post(statements: Union[LaxStatement, List[LaxStatement]]):
...
-        )
+        ES_CLIENT.put(statements_dict.values(), ignore_errors=False)
     except BulkIndexError as exc:
         logger.error("Failed to index submitted statements")
         raise HTTPException(
             status_code=500, detail="Statements bulk indexation failed"
         ) from exc
 
-    logger.info("Indexed %d statements with success", actions)
+    logger.info("Indexed %d statements with success", len(statements_dict))
```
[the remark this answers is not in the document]
SergioSim: Good idea) Thanks) Updated this part)

tests/backends/database/test_es.py
```diff
@@ -263,11 +265,12 @@ def test_backends_database_es_put_with_badly_formatted_data_raises_a_bulkindexer
     )
 
     # By default, we should raise an error and stop the importation
-    with pytest.raises(BulkIndexError):
+    with pytest.raises(BulkIndexError) as exception_info:
         database.put(sys.stdin, chunk_size=2)
     es.indices.refresh(index=ES_TEST_INDEX)
     hits = es.search(index=ES_TEST_INDEX)["hits"]["hits"]
     assert len(hits) == 5
+    assert exception_info.value.args[-1] == 5
```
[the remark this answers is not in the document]
SergioSim: To be honest, I neither) Good Idea to test that) the expected message matches. However, there is a slight change: the full message becomes: "('1 document(s) failed to index.', 5)" instead of ''1 document(s) failed to index.".

jmaupetit (approved):

PR #186, opened 2022-04-25
🐛(tray) do not create a cronjobs list when no cronjob has been defined

jmaupetit:
## Purpose

An empty ralph_cronjobs list generates an empty CronJobList template leading to a deployment failure, k8s refusing submitted manifest.

## Proposal

- [x] only create the CronJobList object when the ralph_cronjobs list contains items


SergioSim (approved):
Looks good)
image

PR #187, opened 2022-04-29
🔥(models) remove ModelRules constraint

SergioSim:
## Purpose

The model rules constraint limited the validation & conversion logic to work only with disjoint rule sets, thus avoiding the case where a matching ruleset could lead to multiple results.
In the future, we intend to support xAPI profile validation.
However, xAPI profile template selection constraints are not always disjoint rulesets.

## Proposal

- [x] remove model rules
- [x] update validator & convertor to support the case where multiple models match an event.

### DIFF
--- a/src/ralph/models/converter.py
+++ b/src/ralph/models/converter.py
@@ -229,13 +229,21 @@ class Converter:
             ValidationError: When the final converted event is invalid.
         """
 
+        error = None
         event = json.loads(event_str)
-        model = self.model_selector.get_model(event)
-        conversion_set = self.src_conversion_set.get(model, None)
-        if not conversion_set:
-            message = "No conversion set found for input event"
-            raise MissingConversionSetException(message)
-        return convert_dict_event(event, event_str, conversion_set)
+        for model in self.model_selector.get_models(event):
+            conversion_set = self.src_conversion_set.get(model, None)
+            if not conversion_set:
+                message = "No conversion set found for input event"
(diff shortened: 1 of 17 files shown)


src/ralph/models/converter.py
```diff
@@ -229,13 +229,23 @@ def _convert_event(self, event_str: str):
...
+                error = MissingConversionSetException(message)
+                continue
+            try:
+                result = convert_dict_event(event, event_str, conversion_set)
+                break
+            except (ConversionException, ValidationError) as err:
+                error = err
+        if error:
+            raise error
+        return result
```
quitterie-lcs, line 248: The docstring should be updated concerning the `returns` field :wink:

src/ralph/models/selector.py
```diff
@@ -158,10 +162,10 @@ def get_decision_tree(self, model_rules):
                 continue
             true_subtree[model] = list(filter(lambda rule: rule != root_rule, rules))
 
-        if next(iter(true_subtree.values())):
+        if sorted(true_subtree.values(), key=lambda x: -len(x))[0]:
```
quitterie-lcs, line 165: So much more beautiful than a `next(iter))` :heart_eyes:

quitterie-lcs (commented):
One global remark: fall all selector tests, could you replace `get_model` with `get_first_model` in the docstring?

quitterie-lcs (approved):
LGTM

PR #194, opened 2022-06-02
Add MongoDB backend

jmaupetit:
## Purpose

MongoDB is a classical database for LRSs such as LearningLocker or Apereo Learning Analytics Initiative. Supporting this backend seems a must-have for Ralph.

## Proposal

- [x] add LearningLocker parser
- [x] add MongoDB backend
- [x] ~add learning locker parser~ **edit: postponed to a new PR**

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -40,6 +40,15 @@ RALPH_APP_DIR=/app/.ralph
 # RALPH_ES_TEST_HOSTS=http://elasticsearch:9200
 # RALPH_ES_TEST_INDEX=test-index
 
+# MONGO database backend
+
+# RALPH_MONGO_COLLECTION=foo
+# RALPH_MONGO_DATABASE=statements
+# RALPH_MONGO_URI=mongodb://localhost:27017/
+# RALPH_MONGO_TEST_COLLECTION=foo
+# RALPH_MONGO_TEST_DATABASE=statements
+# RALPH_MONGO_TEST_URI=mongodb://mongodb:27017/
+
 # Sentry
 
 # RALPH_SENTRY_DSN=
--- a/Makefile
+++ b/Makefile
@@ -106,14 +106,16 @@ logs: ## display app logs (follow mode)
 	@$(COMPOSE) logs -f app
 .PHONY: logs
 
-run: ## alias for run-es
+run: ## alias for run-es run-mongo
 run: \
-	run-es
+	run-es \
+	run-mongo
 .PHONY: run
 
 run-all: ## start all supported local backends
 run-all: \
 	run-es \
+	run-mongo \
 	run-swift
 .PHONY: run-all
 
@@ -122,6 +124,12 @@ run-es: ## start elasticsearch backend
 	@echo "Waiting for elasticsearch to be up and running..."
 	@$(COMPOSE_RUN) dockerize -wait tcp://elasticsearch:9200 -timeout 60s
 .PHONY: run-es
+
+run-mongo: ## start mongodb backend
+	@$(COMPOSE) up -d mongo
+	@echo "Waiting for mongo to be up and running..."
+	@$(COMPOSE_RUN) dockerize -wait tcp://mongo:27017 -timeout 60s
+.PHONY: run-mongo
 
 run-swift: ## start swift backend
 	@$(COMPOSE) up -d swift
@@ -138,7 +146,7 @@ stop: ## stops backend servers
 .PHONY: stop
 
 test: ## run back-end tests
-test: run-es
+test: run
 	bin/pytest
 .PHONY: test
 
--- a/src/ralph/backends/database/mongo.py
+++ b/src/ralph/backends/database/mongo.py
@@ -1,0 +1,135 @@
+"""MongoDB database backend for Ralph"""
+
+import hashlib
+import json
+import logging
+from typing import Generator, TextIO, Union
+
+from bson.objectid import ObjectId
+from pymongo import MongoClient
+from pymongo.errors import BulkWriteError
+
+from ralph.defaults import MONGO_COLLECTION, MONGO_DATABASE, MONGO_URI
+from ralph.exceptions import BadFormatException
+
+from .base import BaseDatabase
+
+logger = logging.getLogger(__name__)
+
+
+class MongoDatabase(BaseDatabase):
+    """Mongo database backend."""
+
+    name = "mongo"
+
+    def __init__(
+        self,
+        connection_uri: str = MONGO_URI,
+        database: str = MONGO_DATABASE,
+        collection: str = MONGO_COLLECTION,
+        client_options: dict = None,
+    ):
+        """Instantiates the Mongo client.
+
+        Args:
+            connection_uri (str): MongoDB connection URI.
+            database (str): MongoDB database to connect to.
+            collection (str): MongoDB database collection to get objects from.
+            client_options (dict): A dictionary of valid options for the MongoClient
+                class initialization.
+        """
+        if client_options is None:
+            client_options = {}
+
+        self.client = MongoClient(connection_uri, **client_options)
+        self.database = getattr(self.client, database)
+        self.collection = getattr(self.database, collection)
+
+    def get(self, chunk_size: int = 500):
+        """Gets collection documents and yields them."""
+
+        for document in self.collection.find(batch_size=chunk_size):
+            # Make the document json-serializable
+            document.update({"_id": str(document.get("_id"))})
+            yield document
+
+    @staticmethod
+    def to_documents(
+        stream: Union[TextIO, list], ignore_errors: bool = False
+    ) -> Generator[dict, None, None]:
+        """Converts `stream` lines (one statement per line) to Mongo documents.
+
+        We expect statements to have at least an `id` field that will be used
+        to compute a unique MongoDB Object ID. This ensures that we will not
+        duplicate statements in our database.
+        """
+
+        for line in stream:
+            statement = json.loads(line) if isinstance(line, str) else line
+            if "id" not in statement:
+                msg = f"statement {statement} has no 'id' field"
+                if ignore_errors:
+                    logger.warning(msg)
+                    continue
+                raise BadFormatException(msg)
+            document = {
+                "_id": ObjectId(
+                    hashlib.sha256(bytes(statement["id"], "utf-8")).hexdigest()[:24]
+                ),
+                "_source": statement,
+            }
+
+            yield document
+
+    def bulk_import(self, batch: list, ignore_errors: bool = False):
+        """Inserts a batch of documents into the selected database collection."""
+
+        try:
+            new_documents = self.collection.insert_many(batch)
+        except BulkWriteError as error:
+            if not ignore_errors:
+                raise error
+            logger.warning(
+                "Bulk importation failed for current documents chunk but you choose "
+                "to ignore it.",
+            )
+            return error.details["nInserted"]
+
+        inserted_count = len(new_documents.inserted_ids)
+        logger.debug("Inserted %d documents chunk with success", inserted_count)
+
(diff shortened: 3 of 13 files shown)


src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,120 @@
...
+        """Converts `stream` lines to Mongo documents."""
+
+        for line in stream:
+            statement = json.loads(line) if isinstance(line, str) else line
+            document = {
+                "_id": ObjectId(
+                    hashlib.sha256(
+                        bytes(statement.get("id", None), "utf-8")
+                    ).hexdigest()[:24]
+                ),
```
jmaupetit, line 66: This section requires at least a comment to explain why we need to enforce the ID.

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,120 @@
...
+        for document in self.to_documents(stream):
+
+            batch.append(document)
+            if len(batch) < chunk_size:
+                continue
+
+            success += self.bulk_import(batch, ignore_errors=ignore_errors)
+            batch = []
+
+        # Edge case: if the total number of documents is lower that the chunk size
```
jmaupetit, line 114:
```suggestion
        # Edge case: if the total number of documents is lower than the chunk size
```

jmaupetit (commented):
Ooops, LearningLocker parser tests are missing. Will do this asap.

jmaupetit:
That said, `ElasticSearchParser` are also missing @SergioSim :wink:

SergioSim:
> That said, `ElasticSearchParser` are also missing @SergioSim wink

The parser was covered in the [cli test](https://github.com/openfun/ralph/blob/master/tests/test_cli.py#L148) but I agree that it is always a good idea to test it separately) Well spotted) I will add the test)

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,128 @@
...
+    name = "mongo"
+
+    def __init__(
+        self,
+        connection_uri: str = MONGO_URI,
+        database: str = MONGO_DATABASE,
+        collection: str = MONGO_COLLECTION,
+        client_options: dict = None,
+    ):
+        """Instanciates the Mongo client.
```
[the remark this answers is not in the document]
jmaupetit: Once day, I'll remember this. One day. :grimacing:

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,128 @@
...
+        except BulkWriteError as error:
+            if not ignore_errors:
+                raise error
+            logger.warning(
+                "Bulk importation failed for current documents chunk but you choose "
+                "to ignore it.",
+            )
+            return 0
+
+        success = len(new_documents.inserted_ids)
```
quitterie-lcs, line 91: Shouldn't the name be more explicit?

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,128 @@
...
+
+        for line in stream:
+            statement = json.loads(line) if isinstance(line, str) else line
+            document = {
+                "_id": ObjectId(
+                    hashlib.sha256(
+                        bytes(statement.get("id", None), "utf-8")
+                    ).hexdigest()[:24]
+                ),
+                "statement": statement,
```
SergioSim, line 72: Why do we put the `statement` inside a "statement" key? Could we write the input data directly at the root level?

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            yield document
+
+    def bulk_import(self, batch: list, ignore_errors: bool = False):
+        """Bulk import a batch of documents to selected database collection."""
```
SergioSim, line 78:
```suggestion
        """Inserts a batch of documents into the selected database collection."""
```

SergioSim (commented):
This looks awesome! Have only a few minor suggestions & questions

src/ralph/backends/database/mongo.py
jmaupetit: I mimicked learning locker's schema: https://docs.learninglocker.net/http-statements/

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,128 @@
...
+        to compute a unique MongoDB Object ID. This ensures that we will not
+        duplicate statements in our database.
+        """
+
+        for line in stream:
+            statement = json.loads(line) if isinstance(line, str) else line
+            document = {
+                "_id": ObjectId(
+                    hashlib.sha256(
+                        bytes(statement.get("id", None), "utf-8")
```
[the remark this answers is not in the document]
jmaupetit: I think it's not the backend's responsibility to create such identifier for a statement that has been externally generated. I would prefer considering that the LRS has generated the statement identifier. But you are right: we should raise an error if the statement does not container an identifier at this stage.

.env.dist
```diff
@@ -40,6 +40,12 @@ RALPH_APP_DIR=/app/.ralph
 # RALPH_ES_TEST_HOSTS=http://elasticsearch:9200
 # RALPH_ES_TEST_INDEX=test-index
 
+# MONGO database backend
+
+# RALPH_MONGO_COLLECTION=marsha
```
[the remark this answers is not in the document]
jmaupetit: This is a commented example, but we can choose a more relevant / generic name. Do you have a suggestion?

src/ralph/backends/database/mongo.py
jmaupetit: `success`?

Makefile
```diff
@@ -108,12 +108,14 @@ logs: ## display app logs (follow mode)
 
 run: ## alias for run-es
```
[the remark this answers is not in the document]
quitterie-lcs: Shouldn't we delete this one and just keep the `run-es` and `run-mongo` commands ?

.env.dist
```diff
@@ -40,6 +40,12 @@ RALPH_APP_DIR=/app/.ralph
...
+# RALPH_MONGO_DATABASE=statements
+# RALPH_MONGO_URI=mongodb://localhost:27017/
+
```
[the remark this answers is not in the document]
jmaupetit: I don't get it. Do you suggest to suppress the empty line before `RALPH_MONGO` vars definition?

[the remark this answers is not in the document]
SergioSim: It's a suggestion to add additional "TEST" environment variables which are used in the tests)

jmaupetit:
I've postponed the learning locker events parser to a new PR as this work is starting to grow and would preferably be reviewed separately.

src/ralph/backends/database/mongo.py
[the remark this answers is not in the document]
SergioSim: Sounds good) Could we adapt the error message to be more explicit about what went wrong? Maybe wrap it in a try-catch or access the id field directly (`statement["id"]`) to get a `KeyError: 'id'` in that case? WDYT?

.env.dist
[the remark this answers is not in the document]
jmaupetit: I think I need more coffee :sweat_smile:

.env.dist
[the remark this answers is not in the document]
jmaupetit: Deal!

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,128 @@
...
+        self.database = getattr(self.client, database)
+        self.collection = getattr(self.database, collection)
+
+    def get(self, chunk_size: int = 500):
+        """Gets collection documents and yields them."""
+
+        for document in self.collection.find(batch_size=chunk_size):
+            # Make the document json-serializable
+            document.update({"_id": str(document.get("_id"))})
+            yield document
```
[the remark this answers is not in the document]
jmaupetit: > This is true when the xAPI statement is stored at the root of the document, however, LearningLocker, as you mentioned, stores it inside a "statement" field which has next to it a "person" field containing another MongoDB _id field.

I think you are making a point that needs to be discussed furthermore about the "schema" we use to store statements in the MongoDB backend. Let's do this IRL!

> Alternatively, we could select only the "statement" field in the find query to only work with xAPI data in the case of LearningLocker, however, this would make the Mongo backend limited to LRSes following the LearningLocker schema.

I think we should stay agnostic. The `learninglocker` parser is responsible to extract statements from a leaninglocker Mongo database.

src/ralph/backends/database/mongo.py
jmaupetit: Deal (`inserted_count`).

src/ralph/backends/database/mongo.py
SergioSim: Could the MongoDB backend be more generic in this regard? For example, [lxHive LRS](https://github.com/g3i/lxHive) seems to store xAPI statements at the root level of the MongoDB document.

src/ralph/backends/database/mongo.py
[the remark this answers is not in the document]
jmaupetit: Done.

src/ralph/backends/database/mongo.py
jmaupetit: That's interesting! We should had this to our future IRL discussion.

src/ralph/backends/database/mongo.py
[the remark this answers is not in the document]
jmaupetit: As mention in another comment, if we allow custom queries to get documents in database backends, it is no longer required to write dedicated parsers to support various LRS's "schema".

Concerning binary data, @SergioSim will declare an issue to keep in mind considering `python-bsonjs` if we encounter such use-case.

jmaupetit:
All comments have been addressed now. I would be pleased to have a new review to merge this work @quitterie-lcs & @SergioSim :pray:

tests/backends/database/test_mongo.py
```diff
@@ -0,0 +1,206 @@
...
+    assert hasattr(backend.client, MONGO_TEST_DATABASE)
+    database = getattr(backend.client, MONGO_TEST_DATABASE)
+    assert hasattr(database, MONGO_TEST_COLLECTION)
+
+
+def test_backends_database_mongo_get_method(mongo):
+    """Test the mongo backend get method."""
+
+    # Create records
+    documents = MongoDatabase.to_documents([{"id": "foo"}, {"id": "bar"}])
```
quitterie-lcs, line 39: What do you think of using the `fixed_dictionaries` strategy from hypothesis to generate statements?

jmaupetit: Maybe it's too convoluted for those tests? Also it might be less efficient.

quitterie-lcs: you're right, let's keep the KISS principle in mind

src/ralph/backends/database/mongo.py
```diff
@@ -0,0 +1,135 @@
...
+        try:
+            new_documents = self.collection.insert_many(batch)
+        except BulkWriteError as error:
+            if not ignore_errors:
+                raise error
+            logger.warning(
+                "Bulk importation failed for current documents chunk but you choose "
+                "to ignore it.",
+            )
+            return 0
```
SergioSim, line 96:
```suggestion
            return error.details["nInserted"]
```
It seems pymongo inserts any documents prior to the first failure and omits all subsequent insertions.

SergioSim (approved):
Looks good to me) Have only one minor suggestion)

Image

src/ralph/backends/database/mongo.py
SergioSim: Apropos, could it be helpful to include the number of inserted/skipped documents in the warning log message? WDYT?

PR #204, opened 2022-07-13
🎨(project) use pydantic for settings management

SergioSim:
## Purpose

We choose to use pydantic settings management as it offers a standard
approach for managing settings using environment variables, CLI arguments,
and other sources.
It is well supported by the fastAPI library and provides additional
features for example, "secret variables" which may become useful in
future.

## Proposal

- [X] Use pydantic settings management for global configuration variables
- [X] Add tests
- [X] Use (static) pydantic settings management for previously dynamic configuration variables (obtained through inspection of backend classes)
- [x] Update documentation

## Breaking changes

- Configuration settings are retrived using `from ralph.conf import settings` instead of `from ralph.defaults import {field_name}`
- All backends constructor parameters now have default values obtained from the configuration
- Replaced YAML configuration support with dotenv configuration.
- Backend and Parser parameter environment variables have a new namespace:
  - previously, we used `RALPH_{BACKEND_NAME}_{BACKEND_PARAMETER}`
  - now, we use `RALPH_BACKENDS__{BACKEND_TYPE}__{BACKEND_NAME}__{BACKEND_PARAMETER}`
  - and for parsers: `RALPH_PARSERS__{PARSER_NAME}`
- The backend name is now retrieved from the configuration key name instead of the backend class `name` attribute


jmaupetit (commented):
Wow, looks promising!

General remarks:

1. As the `defaults.py` file no longer stores defaults but configuration, I would prefer moving it to `conf.py`
2. I like the Django pattern for settings management, for our project this will look like the following, WDYT?

```python
# ralph/conf.py
from pydantic import BaseSettings

class Settings(BaseSettings):
    FOO: str = "bar"

settings = Settings()

# ralph/foo.py
from ralph.conf import settings

foo = settings.FOO
```
3. I would prefer separating concerns between the application settings management and CLI-related business, IMO it should not be mixed in the settings.

Let's talk about this IRL!

src/ralph/backends/storage/swift.py
```diff
@@ -32,16 +27,20 @@ class SwiftStorage(
...
-        os_identity_api_version=SWIFT_OS_IDENTITY_API_VERSION,
+        os_tenant_id: str = settings.BACKENDS.STORAGE.SWIFT.OS_TENANT_ID,
+        os_tenant_name: str = settings.BACKENDS.STORAGE.SWIFT.OS_TENANT_NAME,
+        os_username: str = settings.BACKENDS.STORAGE.SWIFT.OS_USERNAME,
+        os_password: str = settings.BACKENDS.STORAGE.SWIFT.OS_PASSWORD,
+        os_region_name: str = settings.BACKENDS.STORAGE.SWIFT.OS_REGION_NAME,
+        os_storage_url: str = settings.BACKENDS.STORAGE.SWIFT.OS_STORAGE_URL,
+        os_user_domain_name: str = settings.BACKENDS.STORAGE.SWIFT.OS_USER_DOMAIN_NAME,
+        os_project_domain_name: str = (
+            settings.BACKENDS.STORAGE.SWIFT.OS_PROJECT_DOMAIN_NAME
```
SergioSim, line 38:
```suggestion
            swift_settings.OS_PROJECT_DOMAIN_NAME
```
It's quite verbose here, could we use an alias? or do we keep the same settings pattern everywhere?

src/ralph/defaults.py
```diff
@@ -1,141 +1,9 @@
...
-            "level": "INFO",
-        },
-        "swiftclient": {
-            "handlers": ["console"],
-            "level": "ERROR",
-        },
-    },
-}
-
 APP_DIR = Path(environ.get("RALPH_APP_DIR", get_app_dir("ralph")))
```
SergioSim, line 8: Do we keep `APP_DIR` and `MODEL_PATH_SEPARATOR` in defaults.py, or do we move them to conf.py too?

src/ralph/backends/storage/swift.py
jmaupetit: By an alias you mean something like the following?

```python
# swift.py
from ralph.conf import settings

swift_settings = settings.BACKENDS.STORAGE.SWIFT
```

If yes, I'm totally fine with it.

SergioSim: Thanks) updated this)

src/ralph/defaults.py
SergioSim: Good)

tests/api/test_statements_get.py
```diff
@@ -269,12 +265,16 @@ def test_get_statements_until_timestamp(auth_credentials):
...
-def test_get_statements_with_pagination(auth_credentials):
+def test_get_statements_with_pagination(monkeypatch, auth_credentials):
     """
     When the first page does not contain all possible results, it includes
     a "more" property with a link to get the next page of results.
     """
+
+    monkeypatch.setattr(
+        "ralph.api.routers.statements.settings.RUNSERVER_MAX_SEARCH_HITS_COUNT", 2
+    )
```
jmaupetit, line 276: Thank you for this. I hate `mock` :grimacing:

tests/backends/stream/test_ws.py
```diff
@@ -17,10 +16,6 @@ def test_backends_stream_ws_stream_instantiation(ws):
 
     assert WSStream.name == "ws"
 
-    error = "missing 1 required positional argument: 'uri'"
-    with pytest.raises(TypeError, match=error):
-        WSStream()  # pylint: disable=no-value-for-parameter
```
jmaupetit, line 22: Why removing this test?

src/ralph/conf.py
```diff
@@ -0,0 +1,235 @@
...
+from os import environ
+from pathlib import Path
+from typing import Literal, Union
+
+from click import get_app_dir
+from pydantic import BaseModel, BaseSettings
+
+from ralph.utils import import_string
+
+APP_DIR = Path(environ.get("RALPH_APP_DIR", get_app_dir("ralph")))
```
jmaupetit, line 13: Can we define it in the main Settings class and self reference this field in the rest of the configuration?

CHANGELOG.md
```diff
@@ -33,6 +34,12 @@ and this project adheres to
 - Upgrade `uvicorn` to `0.18.2`
 - Upgrade `websockets` to `10.3`
 - Make backends yield results instead of writing to standard streams
+- Use pydantic settings management instead of global variables in defaults.py
+- Rename backend and parser parameter environment variables
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks) updated this part)

tests/backends/stream/test_ws.py
SergioSim: Good catch) Thanks) The WSStream constructor signature was updated with a default value. Updated this test)

src/ralph/conf.py
SergioSim: I don't find an easy way to achieve this.

Tried something like this:
```
PATH: str = Field(
    default_factory=lambda: str(Settings.__fields__["APP_DIR"].default / "archives")
)
```
However, this approach only uses the static default value and doesn't get updated if the default is changed via an environment variable (which happens at instantiation) and it doesn't work in the nested Config class :disappointed:.

Another solution might be defining a second settings class (something like `CoreSettings`), which defines APP_DIR and is instantiated prior to defining Settings. WDYT?

jmaupetit: The following snippet just works :tm::

```python
from os import environ
from pathlib import Path

from pydantic import BaseSettings


class Settings(BaseSettings):

    ROOT_DIR: Path = Path(environ.get("HOME", Path.cwd())) / ".config/myapp"
    HISTORY_FILE_PATH: Path = ROOT_DIR / "history.json"


settings = Settings()
print(settings)
```

The output:

```
ROOT_DIR=PosixPath('/home/maupetit/.config/myapp')
HISTORY_FILE_PATH=PosixPath('/home/maupetit/.config/myapp/history.json')                                                                                                                                                                               
```

WDYT?

SergioSim: That's Awesome! I wasn't aware of this) Thanks) However, I think two issues remain: 

We use the `APP_DIR` in the nested Config metaclass:

```python
class Settings(BaseSettings):
    class Config:
        env_file = APP_DIR / ".env"
```

and in `FSStorageBackendSettings`, a separate model that is inside Settings (has to be defined before).

jmaupetit: Ok, then. I think your first idea to define separated core settings solves all issues you raised. Go!

```python
from os import environ
from pathlib import Path

from pydantic import BaseSettings


class CoreSettings(BaseSettings):

    ROOT_DIR: Path = Path(environ.get("HOME", Path.cwd())) / ".config/myapp"


class Settings(BaseSettings):

    _CORE: BaseSettings = CoreSettings()
    HISTORY_FILE_PATH: Path = _CORE.ROOT_DIR / "history.json"

    class Config:
        env_file: str = CoreSettings().ROOT_DIR / ".env"


settings = Settings()
print(settings)
```

SergioSim: Thank you for your help :pray: Updated this part)

src/ralph/conf.py
```diff
@@ -181,22 +200,19 @@ class ParserSettings(BaseModel):
...
-    class Config:
+    class Config(BaseSettingsConfig):
         """Pydantic Configuration."""
 
-        case_sensitive = True
-        env_nested_delimiter = "__"
-        env_prefix = "RALPH_"
-        env_file = APP_DIR / ".env"
-        env_file_encoding = getattr(io, "LOCALE_ENCODING", "utf8")
+        env_file = core_settings.APP_DIR / ".env"
```
jmaupetit, line 206: The default path to the dot env file is expected to be `.env` by convention, _i.e._ a dot file in the current directory. Dot env files are supposed to be set on a per-project basis, not globally for the application.

jmaupetit (commented):
One tiny change, a rebase and we will be good to go! :tada: :tada: :tada:

src/ralph/conf.py
SergioSim: Thanks) I wasn't aware of the convention; good to know) Updated this part in this [fixup commit](https://github.com/openfun/ralph/commit/aab3122288f5c5c25962fa62c0a165b979ade7df))

jmaupetit (approved):

PR #206, opened 2022-07-19
🚧(project) use pydantic for settings management (static version)

SergioSim:
A version of the previous pull request #204 but using only static configuration values.

We use a test instead to assure that static configuration values align with backend signatures.


SergioSim:
We moved the changes of this pull request to the previously created  #204  pull request

PR #212, opened 2022-09-05
🐛(backends) fix history mixin logger

jmaupetit:
## Purpose

Mixins logger is not used for the history mixin, droping all messages to the root logger.

## Proposal

- [x] target the appropriate logger


SergioSim (approved):
Well spotted! :muscle:

image

quitterie-lcs (approved):

Issue #154, opened 2021-11-10
Integrate pydocstyle linter


jmaupetit:
## Purpose

For now, we do not lint our docstrings, and Black is not opinionated about this. I think we should enforce our numpy docstring style.

## Proposal

Integrate [pydocstyles](https://github.com/PyCQA/pydocstyle) using the `numpy` convention.

PR #214, opened 2022-09-13
✨(project) addition of pydocstyle as a docstring linter

VictorVerbeke:
## Purpose

This PR aims to implement the issue #154, which proposes to implement pydocstyle as a docstring linter.

I added numerous docstring error codes in ignore so we can fix them one at a time until the whole documentation is valid.

Some errors should be ignored:
- D105 can be ignored as it concern magic methods, and they oftendo not require documentation.
- D106 can also be ignored as it concerns nested classes, which are rarely documented. But it can be corrected, as there are not many D106 errors.
- D205 is problematic as some docstring titles are two lines long. I would consider ignoring it.
- D400 is linked to D205: if the title is two lines long, the first line does not end with a period. I would also ignore this error.

## Contributions

- [x] Add pydocstyle in the Makefile, in the setup.cfg and in the .circleci config.
- [x] Fix of errors D100, D401
- [ ] Fix of errors D406 and D407 (underlined sections for numpy convention)
- [ ] Fix of error D202 (Removal of unnecessary line after docstrings)
- [ ] Fix of error D104 (Public packages docstring, mostly in __init__.py files)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,8 @@ lint: \
   lint-black \
   lint-flake8 \
   lint-pylint \
-  lint-bandit
+  lint-bandit \
+  lint-pydocstyle
 .PHONY: lint
 
 lint-black: ## lint back-end python sources with black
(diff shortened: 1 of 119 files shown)


jmaupetit (commented):
Wow, thanks!

Can you fix the CI `lint-git` job? :wink:

Makefile
```diff
@@ -102,6 +103,10 @@ lint-bandit: ## lint back-end python sources with bandit
 	@$(COMPOSE_TEST_RUN_APP) bandit -qr src/ralph
 .PHONY: lint-bandit
 
+lint-pydocstyle: ## Lint Python docstrings with pydocstyle
```
quitterie-lcs, line 106:
```suggestion
lint-pydocstyle: ## lint Python docstrings with pydocstyle
```

quitterie-lcs (commented):
Great job! Only one very tiny remark in my review :wink: :

PR #217, opened 2022-09-30
Tray: add LRS server

jmaupetit:
## Purpose

Add support to run Ralph's LRS server on k8s.

## Proposal

- [x] add database status checking
- [x] add health router
- [x] start the LRS server as the default command
- [x] add LRS server support
- [x] add k8s-related helpers
- [x] use k8s-related helpers in the CI

PR #218, opened 2022-10-11
📦️(project) improve dependencies management

jmaupetit:
## Purpose

Ralph can be used as a CLI, as a server, or as a library. To handle this complexity, we need to split dependencies.

## Proposal

We decided to split dependencies to core and extra requirements:

- the core package only contains dependencies to use Ralph as a library (mostly to use models in a python project),
- the `cli` extra installs Click-related dependencies,
- the `backends` extra installs all supported backends; it might also correspond to a library usage,
- the `lrs` extra installs FastAPI-related dependencies for the LRS server implementation.

A full installation command would look like:

```
pip install ralph-malph[backends,cli,lrs]
```

While a library installation could be:

```
pip install ralph-malph
```

or

```
pip install ralph-malph[backends]
```

Depends on #217 for the Tray testing in the CI.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -39,14 +39,26 @@ Ralph package can be installed from PyPI using the `pip` tool:
 $ python3.9 -m venv venv
 $ source venv/bin/activate
 
-# Install the package (in a virtualenv)
-(venv) $ pip install ralph-malph
+# Install the full package (in a virtualenv)
+(venv) $ pip install ralph-malph[backend-es,backend-ldp,backend-mongo,backend-swift,backend-ws,cli,lrs]
+
+# Install only the core package with the Elasticsearch backend and the LRS (in a virtualenv)
+(venv) $ pip install ralph-malph[backend-es,lrs]
 
 # Test the ralph command (it should be in your PATH)
 (venv) $ ralph --help
 ```
 
-Alternatively, Docker users can pull the latest ralph image and start playing
+If you only need to integrate `ralph` models in your project, you don't need to
+install the `backends`, `cli` or `lrs` extra dependencies, the core package is
+what you need:
+
+```sh
+# Install the core library (in a virtualenv)
+(venv) $ pip install ralph-malph
+```
+
+Alternatively, Docker users can pull the latest `ralph` image and start playing
 with it:
 
 ```sh
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,32 +23,11 @@ classifiers =
 [options]
 include_package_data = True
 install_requires =
-    bcrypt==3.2.2
-    click==8.1.3
-    click-option-group==0.5.3
-    elasticsearch==8.3.3
-    fastapi==0.79.1
-    gunicorn==20.1.0
-    ; We temporary pin `h11` to avoid pip downloading the latest version to solve a
-    ; dependency conflict caused by `httpx` which requires httpcore>=0.15.0,<0.16.0 and
-    ; `httpcore` depends on h11>=0.11,<0.13.
-    ; See: https://github.com/encode/httpx/issues/2244
-    h11==0.12.0
-    httpx==0.23.0
-    Jinja2==3.1.2
+    ; By default, we only consider core dependencies required to use Ralph as a
+    ; library (mostly models).
     langcodes==3.3.0
-    ovh==1.0.0
     pydantic[dotenv,email]==1.9.2
-    pymongo[srv]==4.2.0
-    pyparsing==3.0.9
-    python-keystoneclient==5.0.0
-    python-swiftclient==4.0.1
-    requests==2.28.1
     rfc3987==1.3.8
-    sentry_sdk==1.9.5
-    uvicorn[standard]==0.18.2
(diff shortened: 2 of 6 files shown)


quitterie-lcs (commented):
LGTM!
However when renovate bot updates dependencies every week, we have to be aware not to forget that backend, lrs and cli dependencies have to be updated in the Changelog :wink:

jmaupetit:
Ready for a new review! :pray:

docs/index.md
```diff
@@ -39,14 +39,23 @@ Ralph package can be installed from PyPI using the `pip` tool:
...
+If you only need to integrate Ralph's models in your project, you don't need to
+install the `backends`, `cli` or `lrs` extra dependencies, the core package is
+what you need:
+
+```sh
+# Install the core library (in a virtualenv)
+(venv) $ pip install ralph-malph
+```
+
+Alternatively, Docker users can pull the latest Ralph's image and start playing
```
sampaccoud, line 58:
```suggestion
Alternatively, Docker users can pull the latest `ralph` image and start playing
```

setup.cfg
```diff
@@ -23,39 +23,29 @@ classifiers =
...
 
 [options.extras_require]
+backends =
+    elasticsearch==8.3.3
+    ovh==1.0.0
+    pymongo[srv]==4.2.0
+    python-keystoneclient==5.0.0
+    python-swiftclient==4.0.1
+    requests==2.28.1
+    websockets==10.3
```
sampaccoud, line 45: There might be another pattern more adapted for this to avoid installing all backend libraries: entry points.
This is what pylint is doing for example where you add functionalities by installing separate backages. This also allows extension by external projects...

src/ralph/conf.py
```diff
@@ -4,10 +4,18 @@
 from pathlib import Path
 from typing import Literal, Union
 
-from click import get_app_dir
+try:
+    from click import get_app_dir
+except ImportError:
+    # If we use Ralph as a library and Click is not installed, we consider the
+    # application directory to be the current directory. For non-CLI usage, it
+    # as no consequences.
```
sampaccoud, line 12:
```suggestion
    # has no consequences.
```

src/ralph/conf.py
```diff
@@ -4,10 +4,18 @@
...
+    from unittest.mock import Mock
+
+    get_app_dir = Mock(return_value=".")
```
sampaccoud, line 15: I'm not confortable with using mocks for anything else than tests. But maybe it's close minded. 
Is it better to write 2 classes and return either one depending on the case? Or set APP_DIR in the `__new__` method?

setup.cfg
jmaupetit: If I understand well, what you describe is a dynamic plugin system and I love this idea! I think it's far beyond the scope of this PR, but I'll declare an issue for this. Thanks for this suggestion ❤️

src/ralph/conf.py
jmaupetit: > I'm not confortable with using mocks for anything else than tests. But maybe it's close minded.

jmaupetit: > I'm not confortable with using mocks for anything else than tests. But maybe it's close minded.

I was also surprised to read this, but it seems to be a common pattern: https://realpython.com/python-import/#handle-missing-packages-use-a-mock-instead

setup.cfg
sampaccoud: So in this PR will you split in several install options?

docs/index.md
```diff
@@ -39,14 +39,23 @@ Ralph package can be installed from PyPI using the `pip` tool:
...
-(venv) $ pip install ralph-malph
+# Install the full package (in a virtualenv)
+(venv) $ pip install ralph-malph[backends,cli,lrs]
 
 # Test the ralph command (it should be in your PATH)
 (venv) $ ralph --help
 ```
 
-Alternatively, Docker users can pull the latest ralph image and start playing
+If you only need to integrate Ralph's models in your project, you don't need to
```
quitterie-lcs, line 49:
```suggestion
If you only need to integrate `ralph` models in your project, you don't need to
```

quitterie-lcs (commented):
LGTM! One tiny remark

setup.cfg
jmaupetit: What you are suggesting is to provide one extra-option per backend?

```
# Install only the Elasticsearch backend
$ pip install ralph-malph[backend-es]

# Install Mongo and Swift backends, plus the cli
$ pip install ralph-malph[backend-mongo,backend-swift,cli]
```

If yes, we can do that. Installing the full package, will be a bit verbose, but we can.

sampaccoud: Up to you to decide. My understanding was that it would always be one of the backends and not several backends at the same time (except maybe for developpers?).

jmaupetit: I've fixed it in https://github.com/openfun/ralph/pull/218/commits/918206271e7a45849632657bdb1bc6478455b124

sampaccoud (approved):

PR #222, opened 2022-10-20
Add pydocstyle linter

quitterie-lcs:
## Purpose

`pydocstyle` is a docstring linter that checks compliance for docstrings with Python conventions.

## Proposal

- Configure pydocstyle 
- Fix errors


setup.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python # noqa: D100
```
jmaupetit, line 1: Who's complaining here? About What?

setup.cfg
```diff
@@ -111,6 +112,11 @@ exclude =
     node_modules,
     */migrations/*
 
+[pydocstyle]
+convention = google
+match = .*\.py
+match_dir = ^(?!(tests)).*
```
jmaupetit, line 118: Can you instead restrict the analysis to `src/ralph`?

setup.py
quitterie-lcs: Not necessary anymore

jmaupetit (approved):

PR #224, opened 2022-10-21
Improve dependencies management

jmaupetit:
## Purpose

Current project dependencies definition needs more love:

1. if we want the project to be used as a library, we need to increase the primary dependencies version contraints, and,
2. when Ralph is installed as a library, the `ralph` command is installed and available in the PATH even if Click is not installed leading to unobvious error message when executing the aforementioned command. More than that, if installed as a CLI, the `runserver` command will fail if the the LRS flavor is not installed.

## Proposal

- [x] relax dependencies version contraints
- [x] improve installation flavors consistency by raising an explicit error message
- [x] test dependencies import error

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,9 +25,9 @@ include_package_data = True
 install_requires =
     ; By default, we only consider core dependencies required to use Ralph as a
     ; library (mostly models).
-    langcodes==3.3.0
-    pydantic[dotenv,email]==1.10.2
-    rfc3987==1.3.8
+    langcodes>=3.2.0
+    pydantic[dotenv,email]>=1.10.0
+    rfc3987>=1.3.0
 package_dir =
     =src
 packages = find:
@@ -36,20 +36,21 @@ python_requires = >= 3.9
 
 [options.extras_require]
 backend-es =
-    elasticsearch==8.4.3
+    elasticsearch>=8.0.0
 backend-ldp =
-    ovh==1.0.0
-    requests==2.28.1
+    ovh>=1.0.0
+    requests>=2.0.0
 backend-mongo =
-    pymongo[srv]==4.3.2
+    pymongo[srv]>=4.0.0
 backend-swift =
-    python-keystoneclient==5.0.1
-    python-swiftclient==4.1.0
+    python-keystoneclient>=5.0.0
+    python-swiftclient>=4.0.0
 backend-ws =
-    websockets==10.3
+    websockets>=10.3
 cli =
-    click==8.1.3
-    click-option-group==0.5.5
+    click>=8.1.0
+    click-option-group>=0.5.0
+    sentry-sdk>=1.9.0
 dev =
     bandit==1.7.4
     black==22.10.0
--- a/tests/test_dependencies.py
+++ b/tests/test_dependencies.py
@@ -1,0 +1,47 @@
+"""Tests for Ralph flavors dependencies"""
+
+import importlib
+import re
+import sys
+
+import pytest
+from click.testing import CliRunner
+
+
+def test_dependencies_ralph_command_requires_click(monkeypatch):
+    """Tests Click module installation while executing the ralph command."""
+
+    monkeypatch.setitem(sys.modules, "click", None)
+
+    # Force ralph.cli reload now that click is considered as missing
+    if "ralph.cli" in sys.modules:
+        del sys.modules["ralph.cli"]
+
+    with pytest.raises(
+        ModuleNotFoundError,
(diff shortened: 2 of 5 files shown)


jmaupetit:
Waiting for #225 to be merged first.

setup.cfg
```diff
@@ -25,9 +25,9 @@ include_package_data = True
 install_requires =
     ; By default, we only consider core dependencies required to use Ralph as a
     ; library (mostly models).
-    langcodes==3.3.0
-    pydantic[dotenv,email]==1.10.2
-    rfc3987==1.3.8
+    langcodes>=3.2.0
+    pydantic[dotenv,email]>=1.9.0
```
SergioSim, line 29: Could we require `pydantic>=1.10`? It includes a lot of fixes and improvements (and also a fix for JSON types for hypothesis that we made)

setup.cfg
```diff
@@ -36,20 +36,21 @@ python_requires = >= 3.9
...
+    python-swiftclient>=4.0.0
 backend-ws =
-    websockets==10.3
+    websockets>=10.3
 cli =
-    click==8.1.3
-    click-option-group==0.5.5
+    click>=8.1.0
+    click-option-group>=0.5.0
+    sentry-sdk>=1.9.0
```
SergioSim, line 53: Are `lrs` dependencies in the scope of this PR?
Could it cause problems having `sentry_sdk==1.10.1` in `cli` and `sentry-sdk>=1.9.0` in `lrs` ?

SergioSim (commented):
This looks great) Have only minor questions)

setup.cfg
jmaupetit: As long as version constrains are not conflicting I would say no, but we need to be careful on this.

jmaupetit:
Ready for a new review @SergioSim & @quitterie-lcs :pray:

tests/test_dependencies.py
```diff
@@ -0,0 +1,47 @@
...
+        del sys.modules["ralph.cli"]
+
+    with pytest.raises(
+        ModuleNotFoundError,
+        match=re.escape(
+            "You need to install 'cli' optional dependencies to use the ralph "
+            "command: pip install ralph-malph[cli]"
+        ),
+    ):
+        importlib.import_module("ralph.cli")
```
SergioSim, line 27: We could avoid the if condition by importing `from ralph import cli` and changing

```suggestion
    with pytest.raises(
        ModuleNotFoundError,
        match=re.escape(
            "You need to install 'cli' optional dependencies to use the ralph "
            "command: pip install ralph-malph[cli]"
        ),
    ):
        importlib.reload(cli)
```
WDYT?

tests/test_dependencies.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_dependencies_runserver_subcommand_requires_click(monkeypatch):
+    """Tests Uvicorn module installation while executing the runserver sub command."""
+
+    monkeypatch.setitem(sys.modules, "uvicorn", None)
+
+    # Force ralph.cli reload now that uvicorn is considered as missing
+    if "ralph.cli" in sys.modules:
+        del sys.modules["ralph.cli"]
+    cli = importlib.import_module("ralph.cli")
```
SergioSim, line 38:
```suggestion
    monkeypatch.delattr(cli, "uvicorn")
```

tests/test_dependencies.py
```diff
@@ -0,0 +1,47 @@
...
+        ModuleNotFoundError,
+        match=re.escape(
+            "You need to install 'cli' optional dependencies to use the ralph "
+            "command: pip install ralph-malph[cli]"
+        ),
+    ):
+        importlib.import_module("ralph.cli")
+
+
+def test_dependencies_runserver_subcommand_requires_click(monkeypatch):
```
SergioSim, line 30:
```suggestion
def test_dependencies_runserver_subcommand_requires_uvicorn(monkeypatch):
```

SergioSim (approved):
LGTM) I have only a minor suggestion regarding the tests
image

tests/test_dependencies.py
jmaupetit: At the end, your suggestion produces the same effect, but I found it less explicit.

tests/test_dependencies.py
jmaupetit: Once again, I found it less explicit.

tests/test_dependencies.py
jmaupetit: Well spotted, thanks!

PR #227, opened 2022-10-28
Add helm chart

waammar:
## Purpose

Add Helm chart for Ralph

## Description

Created the chart under `src/helm` I tried to migrate the logic from the `.j2` tray files.
The `version` label was replaced as the helm convention for `app.kubernetes.io/version` 
The checksum from the secret name was deleted. The annotation `checksum/config:` was used instead on the secret, the checksum is automatically generated from `vault.yaml` it help to update the secret object every time a value is updated. 
Note that the `.Chart.AppVersion` is better used to track the version deployed and it should reflect the `image.tag` 
There is only one vars files `values.yaml` the vault is under specific file `vault.yaml` used only for generating the secrets.

## Questions

- Why do you have two service with the same config? the only difference is the label `static` 
- What is `deployment_stamp` ? 
- What is the purpose of `prefix` ?
- How would you like to manage env.? 

## ToDo
- [ ] Add README.md

You can use `helm template . ` under `./src/helm/ralph` to check the output of the chart, it will show you the final manifest that will be deployed.


jmaupetit:
> Why do you have two service with the same config? the only difference is the label static

To make a long story short: to be able to point to a service that has a fixed name (e.g. without blue-green suffix).
Context: https://github.com/openfun/arnold/pull/427

> What is deployment_stamp ?

Our way to distinguish stacks (previous/current/next) in a blue-green deployment strategy. Our playbooks use this selector.

> What is the purpose of prefix ?

Prefix for ingresses?

> How would you like to manage env.?

We want to use the most relevant/recommended community tools. What do you suggest?

jmaupetit:
> > > Why do you have two service with the same config? the only difference is the label static
> > 
> > 
> > To make a long story short: to be able to point to a service that has a fixed name (e.g. without blue-green suffix). Context: [openfun/arnold#427](https://github.com/openfun/arnold/pull/427)
> 
> Do you want to keep this strategy? you can still have one service, (one for each blue and green version) and add two host in the ingress for the last working (static) version with a hostname without the suffix blue/green

No we don't want to keep Arnold-related strategy in this new standard implementation. From what we understand, flux/flagger can handle the blue-green deployment strategy without having to manage this as we did. As a consequence, only one version of each service is required: the static-one.

> > > What is deployment_stamp ?
> > 
> > 
> > Our way to distinguish stacks (previous/current/next) in a blue-green deployment strategy. Our playbooks use this selector.
> 
> I assume you want to preserve this mechanism? so I need to implement this label on the chart

Nope.

> > > What is the purpose of prefix ?
> > 
> > 
> > Prefix for ingresses?
> 
> Yes, I saw it on [ingress](https://github.com/openfun/ralph/blob/master/src/tray/templates/services/app/ingress.yml.j2#L9) and [services](https://github.com/openfun/ralph/blob/master/src/tray/templates/services/app/static-svc.yml.j2#L7) I assume now that the value will be either blue or green?

This is a URL prefix, _e.g._ `previous.my-service.com` or `next.my-service.com`

> > > How would you like to manage env.?
> > 
> > 
> > We want to use the most relevant/recommended community tools. What do you suggest?
> 
> The convention is to start with one `values.yaml` file that contains all the default values and config, and add one values file by env. that only change the related configuration. You must then add the correct values file to your deployment argument, in next iteration we may think of using [Helmfile](https://github.com/roboll/helmfile) tool, that may improve a bit the deploy process to make it more simple, in the long run, it is better to automate the deploy with CD tool like [ArgoCD](https://argo-cd.readthedocs.io/en/stable/)

Duly noted. ArgoCD is in the scope, but for now you don't have to handle this complexity in the current implementation.

src/helm/ralph/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
jmaupetit, line 24:
```suggestion
appVersion: "3.0.0"
```

src/helm/ralph/templates/_helpers.tpl
```diff
@@ -0,0 +1,54 @@
...
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
```
jmaupetit, line 24: Can we indent this block to improve readability?

```suggestion
{{- define "ralph.fullname" -}}
  {{- if .Values.fullnameOverride }}
    {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
  {{- else }}
    {{- $name := default .Chart.Name .Values.nameOverride }}
    {{- if contains $name .Release.Name }}
      {{- .Release.Name | trunc 63 | trimSuffix "-" }}
    {{- else }}
      {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
    {{- end }}
  {{- end }}
{{- end }}
```

waammar: no we can't, for two reason 

1. this is a helper generated by helm 
2. no indentation is by design, the templating will add unintended space to the values and thus create helm parsing error

waammar:
@jmaupetit Can you please make a full review of this PR? I made the final clean up.

src/helm/ralph/templates/_helpers.tpl
jmaupetit: Okay then.

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
+# Ralph Helm Chart
+
+This is the helm chart used to deploy ralph application.
+
+All default values are in `values.yaml`
```
jmaupetit, line 5:
```suggestion
All default values are in `values.yaml`.

```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Review manisfest
+To generate and review your manifest, under `./src/helm`
```
jmaupetit, line 7:
```suggestion
## Review manifest

To generate and review your manifest, under `./src/helm` run the following command:

```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+helm template .
```
jmaupetit, line 9: We usually add the prompt so that users are aware that these are shell commands:

```suggestion
$ helm template .
```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Deploy chart
+### Requirements
+* Helm
+* Needed Kubernetes context selected
+
+
+### Environments
+Please note that with Helm, you can extend the values files. There is no need to copy/paste all the default values and you can replace a value by setting it in your env file.
```
jmaupetit, line 19:
```suggestion

Please note that with Helm, you can extend the values files. There is no need to copy/paste all the default values and you can replace a value by setting it in your env file.
```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+You can add an environments values files under the root of the chart, eg. `dev-values.yaml` and set only the needed customization.
```
jmaupetit, line 21:
```suggestion
You can add an environment values file under the root of the chart, _e.g._ `dev-values.yaml` and set only needed customizations.
```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+This chart use the file `vaul.yaml` to set the mandatory secrets for the application
```
jmaupetit, line 24:
```suggestion
This chart uses the `vault.yaml` file to set the mandatory secrets for the application.
```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### How to
+
+Under `./src/helm`
+
+```
+helm upgrade --install RELEASE_NAME ralph/. --values ralph/dev-values.yaml
```
jmaupetit, line 31:
```suggestion
$ helm upgrade --install RELEASE_NAME ralph/. --values ralph/dev-values.yaml
```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+Tips:
```
jmaupetit, line 34:
```suggestion
Tips:

```

src/helm/README.md
```diff
@@ -0,0 +1,37 @@
...
+* use `--values` to pass an env values files to extend and/or replace the default values
```
jmaupetit, line 35:
```suggestion
* use `--values` to pass an env values file to extend and/or replace the default values
```

src/helm/ralph/templates/_helpers.tpl
```diff
@@ -0,0 +1,53 @@
...
+
+{{/*
+Selector labels
+*/}}
+{{- define "ralph.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "ralph.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+app: ralph
+service: app
+{{- end }}
```
jmaupetit, line 53: A carriage return is missing.

src/helm/ralph/templates/cronjob.yaml
```diff
@@ -0,0 +1,71 @@
...
+              persistentVolumeClaim:
+                claimName: ralph-pvc-history
+            {{- if .Values.ralph.elastic.mount_ca_secret }}
+            - name: es-ca-certificate
+              secret:
+                secretName: {{ .Values.ralph.elastic.ca_secret_name }}
+            {{- end }}
+{{- end }}
+{{- end }}
+{{- end }}
```
jmaupetit, line 71: ditto

src/helm/ralph/templates/secrets.yaml
```diff
@@ -0,0 +1,59 @@
...
+    checksum/config: {{ (tpl (.Files.Glob "vault.yaml").AsSecrets . ) | sha256sum }}
+type: Opaque
+data:
+{{- $v := $.Files.Get "vault.yaml" | fromYaml }}
+{{- range $key, $val := $v }}
+    {{ if eq "ES_CA_CERTIFICATE" $key }}
+        {{ "es-cluster.pem:" | nindent 2 }}: {{ $val | b64enc }}
+    {{- end }}
+{{- end }}
+{{- end }}
```
jmaupetit, line 59: ditto

jmaupetit (commented):
Great! I'll test it by the end of the day and give you more feedback.

src/helm/ralph/templates/secrets.yaml
```diff
@@ -0,0 +1,59 @@
...
+  annotations:
+    checksum/config: {{ (tpl (.Files.Glob "vault.yaml").AsSecrets . ) | sha256sum }}
+type: Opaque
+data:
+{{- $v := $.Files.Get "vault.yaml" | fromYaml }}
+{{- range $key, $val := $v }}
+    {{ if eq "LRS_AUTH" $key }}
+        {{ "auth.json:" | nindent 2 }}: {{ $val | toJson | b64enc }}
+    {{- end }}
+{{- end }}
```
jmaupetit, line 37: Can we do something like the following?

```suggestion
{{ "auth.json:" | nindent 2 }}: {{ get $v "LRS_AUTH" | toJson | b64enc }}
```

src/helm/ralph/templates/secrets.yaml
```diff
@@ -0,0 +1,59 @@
...
+  annotations:
+    checksum/config: {{ (tpl (.Files.Glob "vault.yaml").AsSecrets . ) | sha256sum }}
+type: Opaque
+data:
+{{- $v := $.Files.Get "vault.yaml" | fromYaml }}
+{{- range $key, $val := $v }}
+    {{ if eq "ES_CA_CERTIFICATE" $key }}
+        {{ "es-cluster.pem:" | nindent 2 }}: {{ $val | b64enc }}
+    {{- end }}
+{{- end }}
```
jmaupetit, line 58: ditto

src/helm/README.md
waammar: Fixed

src/helm/ralph/templates/secrets.yaml
waammar: I'd love to, but this is required because of the logic that you added in the tray, the vault file with secret value, I tried to implement the same logic here. Yes, this is a bit over designed in Helm and we may fix it if we use a normal var file. We can even prevent to commit it and add it as requirement to deploy, and in `values.yaml` we add only empty values.

src/helm/ralph/templates/secrets.yaml
waammar: same answer as before, I'll resolve this comment.

waammar:
I fixed some issues, but I still have one related to the `auth.json` of LRS, the yaml to json conversion is failing, it is also related to the complex logic here, maybe we should discuss the design of this solution

src/helm/ralph/templates/secrets.yaml
waammar: fixed

src/helm/ralph/templates/deployment.yaml
```diff
@@ -0,0 +1,98 @@
...
+                  value: "lrs.{{ .Values.namespace }}.{{ .Values.ralph_lrs.domain_name }}"
+            initialDelaySeconds: 5
+            periodSeconds: 5
+          env:
+            - name: RALPH_APP_DIR
+              value: "/app/.ralph"
+            - name: RALPH_AUTH_FILE
+              value: "/var/run/ralph/auth.json"
+            - name: RALPH_BACKENDS__DATABASE__ES__CLIENT_OPTIONS__ca_certs
+              value: "/usr/local/share/ca-certificates/es-cluster.pem"
```
jmaupetit, line 62: As mentioned yesterday, this can be improved thanks to a `ConfigMap` (looping over a `values.yml` file configuration) injected in the environment.

jmaupetit (approved):
Thank you for this contribution @waammar!

PR #228, opened 2022-11-03
🏗️(backends) unify data backends fs

SergioSim:
## Purpose

## Proposal

- [x] add Data and LRS Backend interfaces
- [x] add FileSystemDataBackend


tests/backends/data/test_fs.py
```diff
@@ -0,0 +1,301 @@
...
+        default_chunk_size=1,
+    )
+    assert os.path.exists(deep_path)
+    assert backend.default_directory == Path(deep_path)
+    assert backend.default_directory.is_dir()
+    assert backend.default_query_string == "foo.txt"
+    assert backend.default_chunk_size == 1
+
+    # Check that a storage with the same path doesn't throw an exception
+    FileSystemDataBackend(deep_path)
```
quitterie-lcs, line 46: Could you test it in a try except block to handle the exception(s) ?

quitterie-lcs (commented):
Could we add a bullet point in the CHANGELOG? WDYT?

tests/backends/data/test_fs.py
SergioSim: Good Idea) Thanks) Will update that)

SergioSim:
> Could we add a bullet point in the CHANGELOG? WDYT?

That's a good point) for now, we avoid updating the CHANGELOG on purpose, to keep it simple to rebase the `unify-data-backends` with `master`.  Thus we would like to postpone the CHANGELOG update once the `unify-data-backends` will be ready to be merged with master, if possible)

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+            with path.open("rb") as file:
+                reader = self._read_raw if raw_output else self._read_dict
+                for chunk in reader(file, chunk_size):
+                    yield chunk
+
+            # The file has been read, add a new entry to the history.
+            self.append_to_history(
+                {
+                    "backend": self.name,
+                    "command": "fetch",
```
jmaupetit, line 170: Should we switch the history to a backend logic instead of a CLI driven approach?

```suggestion
                    "action": "read",
```

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+                    # WARNING: previously only the file name was used as the ID
+                    # By changing this to the full file path, previously fetched files
+                    # will not be marked as read anymore.
+                    "id": str(path),
+                    "filename": path.name,
+                    "size": path.stat().st_size,
+                    "fetched_at": now(),
```
jmaupetit, line 177: Can we be more generic?

```suggestion
                    "timestamp": now(),
```

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+
+        # The file has been created, add a new entry to the history.
+        self.append_to_history(
+            {
+                "backend": self.name,
+                "command": "push",
+                "id": str(path),
+                "filename": path.name,
+                "size": path.stat().st_size,
+                "pushed_at": now(),
```
jmaupetit, line 254: ditto

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+
+class FileSystemDataBackendSettings(BaseDataBackendSettings):
+    """Represent the FileSystem data backend default configuration."""
+
+    class Config(BaseSettingsConfig):
+        """Pydantic Configuration."""
+
+        env_prefix = "RALPH_BACKENDS__DATA__FS__"
+
+    DEFAULT_DIRECTORY_PATH: str = str(core_settings.APP_DIR / "archives")
```
jmaupetit, line 35: Shouldn't it be the current directory instead?

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+        if not operation_type:
+            operation_type = self.default_operation_type
+
+        if operation_type == BaseOperationType.DELETE:
+            msg = "Delete operation_type is not allowed."
+            logger.error(msg)
+            raise BackendParameterException(msg)
+
+        if not target:
+            target = str(uuid4())
```
jmaupetit, line 220: :thinking: I hesitate to write to `stdout` if `target` is missing...

If we choose to write to a random file name, shouldn't we prefix it with something like the following:

```suggestion
            target = f"{now()}-{uuid4()}"
```

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+                "id": str(path),
+                "filename": path.name,
+                "size": path.stat().st_size,
+                "pushed_at": now(),
+            }
+        )
+        return 1
+
+    @staticmethod
+    def _read_raw(file, chunk_size):
```
jmaupetit, line 260:
```suggestion
    def _read_raw(file: IO, chunk_size: int) -> Generator[bytes, None, None]:
```

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+        """Read the `file` in chunks of size `chunk_size` and yield them."""
+
+        while chunk := file.read(chunk_size):
+            yield chunk
+
+    @staticmethod
+    def _read_dict(file, _chunk_size):
```
jmaupetit, line 267:
```suggestion
    def _read_dict(file: IO, _chunk_size: int) -> Generator[dict, None, None]:
```

jmaupetit (commented):
That's really great! Some decisions for the FS backend new implementation should be discussed, but I really appreciate your approach! Ping @quitterie-lcs

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+            int: The number of written files. (Always equal to 1).
+        """
+
+        if not operation_type:
+            operation_type = self.default_operation_type
+
+        if operation_type == BaseOperationType.DELETE:
+            msg = "Delete operation_type is not allowed."
+            logger.error(msg)
+            raise BackendParameterException(msg)
```
quitterie-lcs, line 217: Can we define this property directly in the Enum? A class method returning the sub enum allowed for write method as this constraint will be common to a lot of backends. Same for read in LDP

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+            self.default_directory.mkdir(parents=True)
+
+        logger.debug("Default directory: %s", self.default_directory)
+
+    def status(self) -> DataBackendStatus:
+        """Check whether the default directory has appropriate write permissions."""
+
+        if os.access(self.default_directory, os.W_OK):
+            return DataBackendStatus.OK
+
```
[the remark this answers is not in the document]
SergioSim: Nice catch! Thanks) Indeed, read and execute permission checks were missing) Updated this part)

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,274 @@
...
+            details (bool): Get detailed file information instead of just file paths.
+            new (bool): Given the history, list only not already read files.
+
+        Yields:
+            str: The next file path. (If details is False).
+            dict: The next file details. (If details is True).
+        """
+
+        target = Path(target) if target else self.default_directory
+        paths = [str(path) for path in target.iterdir()]
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks) Updated this part)

src/ralph/backends/data/fs.py
SergioSim: Great Idea) Thanks) Updated this part) had to change the current HistoryMixin a bit to work with `action` too)

src/ralph/backends/data/fs.py
SergioSim: Good point) Thanks) Updated this part with `"."`)

src/ralph/backends/data/fs.py
SergioSim: Good Idea) Thanks)

src/ralph/backends/data/base.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Yields:
+            dict: If `raw_output` is False.
+            bytes: If `raw_output` is True.
+        """
+
+    @abstractmethod
+    def write(  # pylint: disable=too-many-arguments
+        self,
+        stream: Iterable[Union[bytes, str, dict]],
```
wilbrdt, line 142: Seen with Julien, wouldn't it be less prone to confusion to change the 'stream' parameter to something more generic, like maybe 'data', or 'records', since a stream is not required anymore ?
And we should probably expand its type to just 'Iterable', since it's the only criteria to be parsed and wrote to a container, WDYT ?

wilbrdt (commented):
Hey, just one small remark after writing an example of Ralph's backends API usage as a lib.

src/ralph/backends/data/fs.py
SergioSim: Well spotted) Thanks) updated this part)

src/ralph/backends/data/base.py
SergioSim: Good Idea) Thanks) Renamed the 'stream' parameter to 'data') Do we need to support more types than bytes and dictionaries? I can't find for now an easy way to support all types(

src/ralph/backends/data/fs.py
SergioSim: I'm not sure if I understood it right. Do you mean to extract this if condition in a separate method?
Something like:
```python
def is_operation_type_allowed(self, operation_type: BaseOperationType) -> bool:
    # Raise an exception if not allowed
```
?

src/ralph/backends/data/base.py
```diff
@@ -33,7 +33,7 @@ class Config:
 
         extra = "forbid"
 
-    query_string: Union[str, None] = None
+    query_string: None | str = None
```
wilbrdt, line 36: Discussed with @jmaupetit, Ralph is still compatible with Python3.9 for now, and it would be nice to stay compatible. WDYT?

SergioSim: Good point! Thanks) Updated the changes to stay compatible with python3.9) It might also be worth testing the compatibility in the CI, added a separate PR (#256) for that.

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,291 @@
...
+        # The file has been created, add a new entry to the history.
+        self.append_to_history(
+            {
+                "backend": self.name,
+                "action": "write",
+                "id": str(path),
+                "filename": path.name,
+                "size": path.stat().st_size,
+                "timestamp": now(),
+            }
```
wilbrdt, line 265: Probably outside of this PR scope, but I think it would be nice to normalize history entries, at least for the core, with something looking like this:
``` python
class HistoryEntry(BaseModel):
""" Pydantic model for history entries."""
    backend: str
    action: str
    id: str
    size: int
    timestamp: datetime

    class Config:
        """HistoryEntry Model configuration."""
        extra = "allow"
```
It would avoid mistakes when adding new backends (`get_command_history` requires entry to have `backend` and `command` keys).
WDYT?

jmaupetit: Good idea!

src/ralph/backends/data/base.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Yields:
+            dict: If `raw_output` is False.
+            bytes: If `raw_output` is True.
+        """
+
+    @abstractmethod
+    def write(  # pylint: disable=too-many-arguments
+        self,
+        data: Iterable[Union[bytes, dict]],
```
wilbrdt, line 155: Quick question: with `data` being an Iterable, are we losing, using Ralph as the CLI, the ability to stream from a Linux pipe to write on a backend?

wilbrdt:
Depends on #228

src/ralph/backends/data/base.py
SergioSim: That's a good point) file-like objects like standard IO streams we use in the CLI are Iterable thus, this should be covered, however, it might indeed be useful to explicitly specify IO streams in the annotation to highlight this use case) Will update that)

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,315 @@
...
+        )
+        return 1
+
+    @staticmethod
+    def _read_raw(file: IO, chunk_size: int, _ignore_errors: bool) -> Iterator[bytes]:
```
quitterie-lcs, line 291: Should we be so specific for the `file` type ?

SergioSim: What would be a less specific type hint for a file-like object? We could use IOBase or Iterable in `_read_dict`, however in `_read_raw` we use the `read` method on the file which is part of the `IO` class.

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,326 @@
...
+            msg = "Default directory doesn't exist, creating: %s"
+            logger.info(msg, self.default_directory)
+            self.default_directory.mkdir(parents=True)
+
+        logger.debug("Default directory: %s", self.default_directory)
+
+    def status(self) -> DataBackendStatus:
+        """Checks whether the default directory has appropriate permissions."""
+        for mode in [os.R_OK, os.W_OK, os.X_OK]:
+            if not os.access(self.default_directory, mode):
```
jmaupetit, line 77: Maybe should we add an explicit error log?

tests/backends/data/test_fs.py
```diff
@@ -0,0 +1,979 @@
...
+    ],
+)
+def test_backends_data_fs_data_backend_list_method_with_details(
+    containers, target, expected, fs_backend, fs
+):
+    """Tests the `FSDataBackend.list` method with `details` set to `True`."""
+    # pylint: disable=invalid-name,too-many-arguments
+    for container in containers:
+        fs.create_file(container)
+        os.utime(container, (1, 1))
```
jmaupetit, line 173: After reading your tests, I am not sure container is an appropriate semantic for the FS backend. Should we be more specific, _e.g._ use `files` instead?

jmaupetit (commented):
Almost good to go!

src/ralph/backends/data/fs.py
```diff
@@ -0,0 +1,326 @@
...
+        settings = settings if settings else self.settings_class()
+        self.default_chunk_size = settings.DEFAULT_CHUNK_SIZE
+        self.default_directory = settings.DEFAULT_DIRECTORY_PATH
+        self.default_query_string = settings.DEFAULT_QUERY_STRING
+        self.locale_encoding = settings.LOCALE_ENCODING
+
+        if not self.default_directory.is_dir():
+            msg = "Default directory doesn't exist, creating: %s"
+            logger.info(msg, self.default_directory)
+            self.default_directory.mkdir(parents=True)
```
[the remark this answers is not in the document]
SergioSim: You are right) This part is only executed when the default directory is not a directory (as this value is defined in the settings, it could point to any location).

tests/backends/data/test_fs.py
SergioSim: Well spotted) The idea was to use common semantics across data backend tests, however, indeed, this makes the test less obvious( Updated this part)

src/ralph/backends/data/fs.py
SergioSim: Nice catch) Thanks) Added a error log)

src/ralph/backends/data/fs.py
jmaupetit: FTR: it will be changed in #297 by @wilbrdt

jmaupetit (approved):
Awesome!

PR #229, opened 2022-11-04
✅(tests) fix hypothesis warnings

SergioSim:
## Purpose

The new version of `hypothesis` emits a warning during boolean comparison of a hypothesis strategy which pollutes test logs.
Also, the new version of `pydantic` includes our fix for resolving nested JSON fields. thus we don't need our custom JSON resolver anymore.

## Proposal

- [x] fix `hypothesis` boolean comparison warning by comparing with `None` instead.
- [x] remove our custom JSON resolver.


quitterie-lcs (approved):
Looks good to me !

PR #237, opened 2022-11-29
🔧(tray) add CA cert file path for the LRS ES backend

jmaupetit:
## Purpose

When using the Elasticsearch (ES) backend in an LRS context, for SSL connections to the cluster, python's ES client requires access to the ES cluster self-generated CA root certificate.

## Proposal

- [x] configure ES backend option using an injected environment variable when required


SergioSim (approved):
LGTM) :+1: 
image

PR #241, opened 2022-12-05
🔧(project) improve Sentry integration

jmaupetit:
## Purpose

Sentry integration is too basic and requires to be a bit more configurable and adapted to used web framework (FastAPI).

## Proposal

We now install the FastAPI sentry integration to better track server performance and post mortem report.

Traces sample rate is now configurable for the CLI and the LRS.


SergioSim (commented):
LGTM :+1: 

image

quitterie-lcs (approved):

PR #251, opened 2023-01-09
CLI: add --version option

jmaupetit:
## Purpose

It's often useful to be able to quickly check version number of installed Ralph instance.

## Proposal

- [x] add `--version` CLI option

During this work, I also improved the following:

- [x] pydoctyle: ignore development folders
- [x] cli: homogenize commands help

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ backend-ldp =
     requests>=2.0.0
 backend-mongo =
     pymongo[srv]>=4.0.0
-backend-s3 = 
+backend-s3 =
     boto3>=1.24.70
     botocore>=1.27.71
 backend-swift =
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -119,7 +119,7 @@ exclude =
 
 [pydocstyle]
 convention = google
-match_dir = ^(?!tests).*
+match_dir = ^(?!tests|venv|build|scripts).*
```
quitterie-lcs, line 122: Why did you add this ? I don't get clearly the point

jmaupetit: I don't want to spend time linting those local directories only used for development and containing third-party sources.

wilbrdt (approved):
Looks good to me! 🎉

Issue #246, opened 2022-12-21
`make bootstrap` is failing due to missing build dependences


bmtcril:
## Bug Report

Hi all, I'm excited to get started developing with Ralph. Getting set up I ran into an issue, however. On the master branch `make bootstrap` errors out pip installing psutil due to gcc not being available.

**Expected behavior/code**
`make bootstrap` builds the docker images cleanly.

**Actual Behavior**
An error occurs:

```
#0 22.22       creating build/temp.linux-aarch64-cpython-39/psutil
#0 22.22       gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DPSUTIL_POSIX=1 -DPSUTIL_SIZEOF_PID_T=4 -DPSUTIL_VERSION=594 -DPy_LIMITED_API=0x03060000 -DPSUTIL_LINUX=1 -DPSUTIL_ETHTOOL_MISSING_TYPES=1 -I/usr/local/include/python3.9 -c psutil/_psutil_common.c -o build/temp.linux-aarch64-cpython-39/psutil/_psutil_common.o
#0 22.22       C compiler or Python headers are not installed on this system. Try to run:
#0 22.22       sudo apt-get install gcc python3-dev
#0 22.22       error: command 'gcc' failed: No such file or directory
```

**Steps to Reproduce**
1. Sync the Ralph master branch
2. run `make bootstrap`
3. The above error occurs

**Environment**
- Ralph version: master branch
- Platform: Docker Desktop on Mac OS 13.0.1

**Possible Solution**
I was able to get a successful build by adding the required development tools, see the PR here: https://github.com/bmtcril/ralph/pull/1

jmaupetit:
Thanks for your feedback! :heart: 

I cannot reproduce this behavior, not sure if it's platform-specific or not.

1. Did you add a new dependency before trying to build / bootstrap?
2. Can you provide the full building logs?
3. Which version of Docker + compose are you running?

bmtcril:
Thanks for the reply @jmaupetit ...

1. No new dependencies, just pulled master and ran `make bootstrap`
2. It's pretty long, but I just reproduced it. Here's a gist with the log: https://gist.github.com/bmtcril/6f5b5653318a7baf045750eee73bc079
3. Docker Desktop for Mac, `Docker version 20.10.21, build baeda1f`, `Docker Compose version v2.13.0`

jmaupetit:
[...]

PR #252, opened 2023-01-09
:bug: Fix M1 Mac `make bootstrap` failures

bmtcril:
## Purpose
Fixes https://github.com/openfun/ralph/issues/246 which occurs on M1 Macs during `make bootstrap`.

## Proposal
Add the missing dependencies only for the `linux/arm64` target platform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,17 @@ FROM core as development
 # Copy all sources, not only runtime-required files
 COPY . /app/
 
+# Only the M1 Mac images need these packages installed
+ARG TARGETPLATFORM
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; \
+    then apt-get update && \
+        apt-get install -y \
+            build-essential \
+            gcc \
+            python-dev && \
+        rm -rf /var/lib/apt/lists/*; \
+    fi;
+
 # Uninstall ralph and re-install it in editable mode along with development
 # dependencies
 RUN pip uninstall -y ralph-malph


Dockerfile
```diff
@@ -50,6 +50,18 @@ FROM core as development
...
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; \
+    then apt-get update && \
+        apt-get install -y \
+        build-essential \
+        gcc \
+        python-dev && \
+        rm -rf /var/lib/apt/lists/*; \
+    fi;
+
+
```
jmaupetit, line 64: A single blank line would be sufficient.

```suggestion

```

Dockerfile
```diff
@@ -50,6 +50,18 @@ FROM core as development
...
 COPY . /app/
 
+# Only the M1 Mac images need these packages installed
+ARG TARGETPLATFORM
+RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; \
+    then apt-get update && \
+        apt-get install -y \
+        build-essential \
+        gcc \
+        python-dev && \
```
jmaupetit, line 60: We usually indent packages to distinguish them from commands:

```suggestion
          build-essential \
          gcc \
          python-dev && \
```

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ and this project adheres to
 - Improve Sentry integration for the LRS
 - Update handbook link to `https://handbook.openfun.fr`
 
+### Fixed
+
+- `make bootstrap` errors on M1 Macs
+
```
jmaupetit, line 30: We don't usually log changes related to the development environment or project that are not related to end-users concerns. So I would suggest to remove this. Note that the CHANGELOG check in the CI is allowed to fail :wink:

jmaupetit (commented):
Thank you for this fix :heart: 

I only have minor suggestions. Almost good to go :muscle:

bmtcril:
@jmaupetit Thanks for the feedback, I think it should be set now

jmaupetit (approved):
Yay, thanks!

PR #255, opened 2023-01-17
🚧 (project) add rich library for a user friendly CLI

quitterie-lcs:
## Purpose

`ralph` can be used as a CLI and integrating the rich library makes the tool more user friendly and easy to read.

## Proposal

The [provided handler from rich](https://rich.readthedocs.io/en/stable/reference/logging.html?rich.logging.RichHandler) replaces the custom made stream handler.

:test_tube: logger tests are failing, as for now, I have not updated them with the new handler. The rich integration enforces message wrapping. It is not yet possible to run correctly the test

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -1,5 +1,7 @@
 # Configuration
 RALPH_APP_DIR=/app/.ralph
+FORCE_COLOR=1
+TERM=xterm-256
 
 # Uncomment lines (by removing # characters at the beginning of target lines)
 # to define environment variables associated to the backend(s) you need.
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,6 +54,7 @@ cli =
     bcrypt>=4.0.0
     click>=8.1.0
     click-option-group>=0.5.0
+    rich>=13.1.0
     sentry-sdk[fastapi]>=1.9.0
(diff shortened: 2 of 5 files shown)


quitterie-lcs:
Seen and discussed with @wilbrdt, we should consider configure the logger in a other way. To get the coloration with the RichHandler, a console should be configured.
Will configure the logger with a direct instantiation of the handler

setup.cfg
```diff
@@ -54,6 +54,7 @@ cli =
     bcrypt>=4.0.0
     click>=8.1.0
     click-option-group>=0.5.0
+    rich>=13.1.0
```
wilbrdt, line 57: 13.2.0 now released!

quitterie-lcs:
@jmaupetit what do you think of this configuration for `rich` ?

.env.dist
```diff
@@ -1,5 +1,7 @@
 # Configuration
 RALPH_APP_DIR=/app/.ralph
+FORCE_COLOR=1
+TERM=xterm-256
```
wilbrdt, line 4: I personally don't mind with this env var approach. Maybe add a comment just to be clear for Rich.

PR #202, opened 2022-07-04
⬆️(project) upgrade base python version to 3.11

jmaupetit:
## Purpose

Python 3.11 is the most recent python stable release. We should support it.

## Proposal

- [x] upgrade project base Docker image to python 3.11
- [x] upgrade CI base Docker image to python 3.11
- [x] update the documentation

PR #256, opened 2023-01-18
👷(circle) run tests with multiple python versions in the CI

SergioSim:
## Purpose

We want to make sure that Ralph is compatible with python 3.9, 3.10, and 3.11.

## Proposal

- [x] Run tests in the CI with different python docker images

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,8 +124,11 @@ jobs:
   # ---- Backend jobs ----
   # Build backend development environment
   build:
+    parameters:
+      python-image:
+        type: string
     docker:
-      - image: cimg/python:3.11
+      - image: cimg/<< parameters.python-image >>
         auth:
           username: $DOCKER_HUB_USER
           password: $DOCKER_HUB_PASSWORD
@@ -134,14 +137,14 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v1-dependencies-{{ .Revision }}
+            - v1-dependencies-<< parameters.python-image >>-{{ .Revision }}
       - run:
           name: Install development dependencies
           command: pip install --user .[backend-es,backend-ldp,backend-mongo,backend-swift,backend-ws,cli,dev,lrs]
       - save_cache:
           paths:
             - ~/.local
-          key: v1-dependencies-{{ .Revision }}
+          key: v1-dependencies-<< parameters.python-image >>-{{ .Revision }}
 
   lint:
     docker:
@@ -154,7 +157,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v1-dependencies-{{ .Revision }}
+            - v1-dependencies-python:3.11-{{ .Revision }}
       - run:
           name: Lint code with flake8
           command: ~/.local/bin/flake8
@@ -175,8 +178,11 @@ jobs:
           command: ~/.local/bin/pydocstyle
 
   test:
+    parameters:
+      python-image:
+        type: string
     docker:
-      - image: cimg/python:3.11
+      - image: cimg/<< parameters.python-image >>
         auth:
           username: $DOCKER_HUB_USER
           password: $DOCKER_HUB_PASSWORD
@@ -198,7 +204,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v1-dependencies-{{ .Revision }}
+            - v1-dependencies-<< parameters.python-image >>-{{ .Revision }}
       - run:
           name: Run tests
           command: |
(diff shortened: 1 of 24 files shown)


.circleci/config.yml
```diff
@@ -154,7 +157,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v1-dependencies-{{ .Revision }}
+            - v1-dependencies-python:3.11-{{ .Revision }}
```
quitterie-lcs, line 160: Maybe it's a silly issue, but shouldn't we ensure that it lints correctly also in the previous Python version ?

quitterie-lcs (commented):
Looks good! 
Having a question about the test library job? 
Shouldn't we ensure that it passes also with previous in test-library job? (As a lot of users won't be necessarily in Python 3.11)

.circleci/config.yml
SergioSim: This might become tricky; what do we do in case of a conflict, e.g. if python3.9 linter fails - we fix the issue, but then python3.10 linter fails because of the fix?

SergioSim:
> Looks good! Having a question about the test library job? Shouldn't we ensure that it passes also with previous in test-library job? (As a lot of users won't be necessarily in Python 3.11)

Well Spotted! Thanks) I haven't thought about the test-library job - updated it too

.circleci/config.yml
wilbrdt: Then should we just lint on the oldest version, aka python3.9?

wilbrdt (commented):
Have you tried factoring these in a fixture? It would go nicely along the fixture _moto_fs_, another incompatibility between moto and pyfakefs :grimacing:

tests/backends/storage/test_fs.py
```diff
@@ -69,10 +69,16 @@ def test_backends_storage_fs_url(fs):
 
 # pylint: disable=invalid-name
 # pylint: disable=unused-argument
-def test_backends_storage_fs_list(fs):
+def test_backends_storage_fs_list(fs, monkeypatch):
     """Tests archives listing in FSStorage."""
     fs.create_dir(settings.APP_DIR)
 
+    # Force Path instantiation with fake FS
+    monkeypatch.setattr(
```
wilbrdt, line 77: Have you tried factoring these in a fixture? It would go nicely along the fixture moto_fs, another incompatibility between moto and pyfakefs :grimacing:

SergioSim: Good Idea) Thanks) Updated this part for the swift fixture too)

.circleci/config.yml
SergioSim: I find it convenient when the CI linter version matches the development image version as it might avoid conflicts - should we use the oldest version in development too? Reverting most changes from #202? Or maybe make the development image configurable with a make file variable setting the python version, something like `make bootstrap PYTHON_VERSION=3.9`?

tests/backends/storage/test_fs.py
wilbrdt: Whoops sorry I wasn't quite clear about the fixture: I should have said "alongside" haha.
I would prefer a fixture created just for this incompatibility, that would be common for all the backends. A fixture named for example "settings_fs" which would only contain 
```
    # Force Path instantiation with fake FS
    monkeypatch.setattr(
        "ralph.backends.mixins.settings",
        Settings(HISTORY_FILE=Path(settings.APP_DIR / "history.json")),
    )
```
This fixture would be called wherever necessary.
WDYT?

.circleci/config.yml
quitterie-lcs: > I find it convenient when the CI linter version matches the development image version as it might avoid conflicts - should we use the oldest version in development too? Reverting most changes from #202? Or maybe make the development image configurable with a make file variable setting the python version, something like `make bootstrap PYTHON_VERSION=3.9`?

@SergioSim I did not have this in mind. In this case, we should keep the linter with Python 3.11 version

tests/backends/storage/test_fs.py
SergioSim: Sorry for the misunderstanding; this approach looks even better) Thanks) Updated this part)

.circleci/config.yml
jmaupetit: I think we could be opinionated regarding the python version used in quality controls (linters, etc). Only tests should reflect supported python versions.

.circleci/config.yml
```diff
@@ -503,6 +512,9 @@ workflows:
       # Build, lint and test production and development Docker images
       # (debian-based)
       - build:
+          matrix:
+            parameters:
+              python-image: [python:3.9, python:3.10, python:3.11]
```
jmaupetit, line 517: We should maybe add python 3.8...

jmaupetit (approved):

Issue #254, opened 2023-01-17
xAPI Validation error when `result.extensions` is specified


wilbrdt:
## Bug Report

**Expected behavior/code**
When executing command `validate` on a (known-to-be-correct) xAPI statement, Ralph should return 
`Validating 1 events (ignore_errors=0 | fail-on-unknown=0`

**Actual Behavior**
When executing command `validate` on a (known-to-be-correct) xAPI statement containing **result extensions**, Ralph returns a BadFormatException:
```
2023-01-17 14:25:58,440 INFO     ralph.cli Validating xapi events (ignore_errors=False | fail-on-unknown=False)
2023-01-17 14:25:58,650 ERROR    ralph.models.validator Input event is not a valid VideoSeeked event.
Traceback (most recent call last):
  File "/app/src/ralph/models/validator.py", line 30, in validate
    yield self._validate_event(event_str)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/ralph/models/validator.py", line 77, in _validate_event
    return self.get_first_valid_model(event).json()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/ralph/models/validator.py", line 62, in get_first_valid_model
    raise error
  File "/app/src/ralph/models/validator.py", line 58, in get_first_valid_model
    return model(**event)
           ^^^^^^^^^^^^^^
  File "pydantic/main.py", line 342, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 4 validation errors for VideoSeeked
result -> extensions -> https://w3id.org/xapi/video/extensions/length
  extra fields not permitted (type=value_error.extra)
result -> extensions -> https://w3id.org/xapi/video/extensions/played-segments
  extra fields not permitted (type=value_error.extra)
result -> extensions -> https://w3id.org/xapi/video/extensions/progress
  extra fields not permitted (type=value_error.extra)
@timestamp
  extra fields not permitted (type=value_error.extra)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/ralph", line 33, in <module>
    sys.exit(load_entry_point('ralph-malph', 'console_scripts', 'ralph')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #258, opened 2023-01-23
🩹(xapi) allow extra fields for xAPI extensions

wilbrdt:
## Purpose

Fix issue #254

## Proposal

After reading xAPI profile spec, Ralph should not forbid additional extensions.
Base models of Extension fields now allow extra fields.

### DIFF
--- a/src/ralph/models/xapi/config.py
+++ b/src/ralph/models/xapi/config.py
@@ -9,3 +9,11 @@ class BaseModelWithConfig(BaseModel):
     class Config:  # pylint: disable=missing-class-docstring # noqa: D106
         extra = Extra.forbid
         min_anystr_length = 1
+
+
+class BaseExtensionModelWithConfig(BaseModel):
+    """Pydantic model for extension configuration shared among all models."""
+
+    class Config:  # pylint: disable=missing-class-docstring # noqa: D106
+        extra = Extra.allow
+        min_anystr_length = 0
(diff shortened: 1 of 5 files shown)


src/ralph/models/xapi/config.py
```diff
@@ -9,3 +9,11 @@ class BaseModelWithConfig(BaseModel):
...
         extra = Extra.forbid
         min_anystr_length = 1
+
+
+class BaseExtensionModelWithConfig(BaseModel):
+    """Pydantic model for extension configuration shared among all models."""
+
+    class Config:  # pylint: disable=missing-class-docstring # noqa: D106
+        extra = Extra.allow
+        min_anystr_length = 1
```
SergioSim, line 19: Aren't empty strings allowed in extension fields? Maybe we could use the `BaseModel` directly? (Extra.allow is used by default) WDYT?

wilbrdt: You're right, empty strings should be allowed for all extension fields, even for `object.definition.extensions`. 

Default extra seems to be `ignore` instead of `allow`. It should be best to allow extra extensions as the `validate` command returns a JSON string of the model parsed/validated.

quitterie-lcs (approved):
LGTM ! 
It is a convenient solution for now, before evaluating if we need to reorganize xAPI models without FUN dependencies

PR #262, opened 2023-01-30
✨(authentication) add OpenId Connect authentication

wilbrdt:
## Purpose

Add OpenID Connect ID tokens verification issued by third party Identity Provider.
It works with Keycloak but would probably work with others Identity Providers.

This is the last step of the Authorization Code Flow, described in the following:
1. Client app redirects to the authorization endpoint at the Token Service of the Authorization Server.
2. If the user is not yet authenticated, the Token Service redirects to the Authentication Service. Note that these two entities, while running in the same product, are separate conceptually.
3. The User authenticates, and is redirected back to the Token Service.
4. The Authorization Server issues a one time token called the authorization code. This token needs to be passed to every API calls.

First three steps need to be done outside of Ralph.

## Proposal

- [x] Add ID tokens verification
- [x] Make both BasicAuth and OpenId Connect live together
- [x] Add OpenId Connect documentation
- [x] Test OpenID Connect auth


src/ralph/conf.py
```diff
@@ -262,6 +262,8 @@ class Config(BaseSettingsConfig):
 
     _CORE: CoreSettings = core_settings
     AUTH_FILE: Path = _CORE.APP_DIR / "auth.json"
+    AUTH_PROVIDER_URL: str = "http://localhost:8081/auth/realms/ralph"
```
jmaupetit, line 265: I think we can be more pydantic on this :wink: https://docs.pydantic.dev/usage/types/#urls

jmaupetit (commented):
That's a great start!

Quick remark, I would prefer having an `api.auth` module as:

```
.
└── api
    ├── auth
    │   ├── basic.py
    │   ├── __init__.py
    │   └── oidc.py
    └── __init__.py

2 directories, 4 files
```

WDYT?

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,116 @@
...
+
+    ID Tokens are polymorphic and may have many attributes not defined in the
+    spec thus this model accepts all addition fields. Only required fields are
+    listed in the attributes section of this docstring or enforced by pydantic.
+
+    Attributes:
+        iss (str): Issuer Identifier for the Issuer of the response.
+        sub (str): Subject Identifier.
+        aud (str): Audience(s) that this ID Token is intended for.
+        exp (str): Expiration time on or after which the ID Token MUST NOT be
```
jmaupetit, line 37:
```suggestion
        exp (int): Expiration time on or after which the ID Token MUST NOT be
```

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,116 @@
...
+                   accepted for processing.
+        iat (iat): Time at which the JWT was issued.
```
jmaupetit, line 39:
```suggestion
        iat (int): Time at which the JWT was issued.
```

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,116 @@
...
+
+    class Config:  # pylint: disable=missing-class-docstring # noqa: D106
+        extra = Extra.allow
+
+
+def requests_get(url: str) -> dict:
+    """Send a get request to the url.
+
+    Args:
+        url (str): URL endpoint  to send the get request to.
```
jmaupetit, line 56:
```suggestion
        url (str): URL endpoint to send the get request to.
```

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,116 @@
...
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Could not validate credentials",
+            headers={"WWW-Authenticate": "Bearer"},
+        ) from exc
+
+
+@lru_cache()
+def discover_provider(base_url: str) -> dict:
+    """Discover the authentication server (or OpenId Provider) configuration."""
+    return requests_get(f"{base_url}/.well-known/openid-configuration")
```
jmaupetit, line 79: I think the `requests_get` helper is not required. As it's only used twice, you can duplicate this snippet and add specific logging messages (`logger.ERROR`) so that it's easier to debug OIDC configuration for operators.

jmaupetit (commented):
I see great ideas in this PR! :muscle:

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,116 @@
...
+
+def authenticated_user(auth_header: Union[str, None] = Depends(oauth2_scheme)) -> Union[IDToken, None]:
+    """Decode and validate OpenId Connect ID token against issuer in config.
+
+    Args:
+        token (str): Base64 encoded OIDC Token. This is invoked behind the
+            scenes by Depends.
+
+    Return:
+        IDToken (IDToken):
```
quitterie-lcs, line 96: Missing description

docs/api.md
```diff
@@ -81,7 +92,70 @@ $ curl --user john.doe@example.com:PASSWORD http://localhost:8100/whoami
...
...UzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICJSTWlLMFhESmRaVnhSNUI5dmVOMzdpMXlIWVk3ZmluQXdwU0pjRnRTeE44In0.eyJleHAiOjE2NzYzOTI5NjgsImlhdCI6MTY3NjM5MjY2OCwianRpIjoiZWJlNjlmN2YtYjRiZC00Yzc0LTk4ODYtZDNlZDg4ODhlZWQ4IiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL2F1dGgvcmVhbG1zL2Z1bi1tb29jIiwiYXVkIjoiaHR0cDovL2xvY2FsaG9zdDo4MTAwIiwic3ViIjoiNjhmYTU0MGMtMjdjZi00ZGE4LWJjNmQtMTYzNDU4MDQ1MmUxIiwidHlwIjoiQmVhcmVyIiwiYXpwIjoicmFscGgiLCJzZXNzaW9uX3N0YXRlIjoiMjJhMzY3MzUtZTM1Zi00OTZiLWEyNDMtMTUyZDMyZWJmZjQ1IiwiYWNyIjoiMSIsImFsbG93ZWQtb3JpZ2lucyI6WyJodHRwOi8vbG9jYWxob3N0OjgxMDAiXSwicmVzb3VyY2VfYWNjZXNzIjp7InJhbHBoIjp7InJvbGVzIjpbImFsbCJdfX0sInNjb3BlIjoiZW1haWwgcHJvZmlsZSIsInNpZCI6IjIyYTM2NzM1LWUzNWYtNDk2Yi1hMjQzLTE1MmQzMmViZmY0NSIsImVtYWlsX3ZlcmlmaWVkIjpmYWxzZSwicm9sZXMiOlsiYWxsIl0sInByZWZlcnJlZF91c2VybmFtZSI6InJhbHBoX2FkbWluIiwiZW1haWwiOiJyYWxwaF9hZG1pbkBleGFtcGxlLm9yZyJ9.MljNbx8msaVVaD0FriEceWcBESzjIch2ZVwgFq_yeoBdXo-fT7w4S2MVGM1Vm7Cbo1oaXj_udPDRd0MgSDF-cCj9GpQ4b58mEQLxI4V0gVDVLT7ZFUKTRGVx3oRLYPs9uswr_BjgHOvxTTYnAQMDxkGeDeN1dllgwhV1uuy6NtaiMHYW7nGFO04cqgHpWUbzJZTtS-edz-V-J3eZWULUpV2PDEepjGDwp8SAVxZ0hTwJPrW2JesYlecjgBztWNa_Bh03nImci5eX0a60mWhwTGLko2q_QlI51zd1r_9q0VFgB0WcEtnv3MyGuDatcDhb5J6GE9yxxQcz4hLErRwuZQ"
```
jmaupetit, line 153: ditto

src/ralph/conf.py
```diff
@@ -319,6 +319,8 @@ class Config(BaseSettingsConfig):
         },
     }
     PARSERS: ParserSettings = ParserSettings()
+    RUNSERVER_AUTH_OIDC_AUDIENCE: str = None
+    RUNSERVER_AUTH_OIDC_ISSUER_URI: AnyHttpUrl = None
```
jmaupetit, line 323: Maybe we should be more explicit and define a `RUNSERVER_AUTH_BACKEND` configuration variable? And thus list `RUNSERVER_AUTH_BACKENDS` as an Enum?

```suggestion
    RUNSERVER_AUTH_BACKEND = RUNSERVER_AUTH_BACKENDS.BASIC
    RUNSERVER_AUTH_OIDC_AUDIENCE: str = None
    RUNSERVER_AUTH_OIDC_ISSUER_URI: AnyHttpUrl = None
```

src/ralph/api/auth/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Main module for Ralph's LRS API authentication."""
+
+from ralph.api.auth.basic import get_authenticated_user as get_basic_user
+from ralph.api.auth.oidc import get_authenticated_user as get_oidc_user
+from ralph.conf import settings
+
+# At startup, select the authentication mode that will be used
+get_authenticated_user = (
+    get_oidc_user if settings.RUNSERVER_AUTH_OIDC_ISSUER_URI else get_basic_user
+)
```
jmaupetit, line 10:
```suggestion
get_authenticated_user = (
    get_oidc_user if settings.RUNSERVER_AUTH_BACKEND ==  settings.RUNSERVER_AUTH_BACKENDS.OIDC else get_basic_user
)
```

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,148 @@
...
+from pydantic import BaseModel, Extra
+
+from ralph.api.auth.user import AuthenticatedUser
+from ralph.conf import settings
+
+oauth2_scheme = OpenIdConnect(
+    openIdConnectUrl=f"{settings.RUNSERVER_AUTH_OIDC_ISSUER_URI}"
+    "/.well-known/openid-configuration",
+    auto_error=False,
+)
```
jmaupetit, line 21:
```suggestion
OPENID_CONFIGURATION_PATH = "/.well-known/openid-configuration"
oauth2_scheme = OpenIdConnect(
    openIdConnectUrl=f"{settings.RUNSERVER_AUTH_OIDC_ISSUER_URI}{OPENID_CONFIGURATION_PATH}",
    auto_error=False,
)
```

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,148 @@
...
+    class Config:  # pylint: disable=missing-class-docstring # noqa: D106
+        extra = Extra.allow
+
+
+@lru_cache()
+def discover_provider(base_url: str) -> dict:
+    """Discover the authentication server (or OpenId Provider) configuration."""
+    try:
+        response = requests.get(
+            f"{base_url}/.well-known/openid-configuration", timeout=5
```
jmaupetit, line 57:
```suggestion
            f"{base_url}{OPENID_CONFIGURATION_PATH}", timeout=5
```

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Could not validate credentials",
+            headers={"WWW-Authenticate": "Bearer"},
+        ) from exc
+
+
+@lru_cache()
+def get_public_keys(jwks_uri: str):
```
jmaupetit, line 73: Can we be more specific and use Pydantic types?

```suggestion
def get_public_keys(jwks_uri: AnyUrl):
```

jmaupetit: Can we also type/define a model for returned value?

tests/api/test_forwarding.py
```diff
@@ -167,7 +167,9 @@ async def post_success(*args, **kwargs):  # pylint: disable=unused-argument
     ] == [message for source, _, message in caplog.record_tuples if source != "asyncio"]
 
 
-@hypothesis_settings(suppress_health_check=(HealthCheck.function_scoped_fixture,))
+@hypothesis_settings(
+    deadline=None, suppress_health_check=(HealthCheck.function_scoped_fixture,)
+)
```
jmaupetit, line 172: Maybe we can restore the default deadline as we've upgraded CI execution class?

tests/fixtures/auth.py
```diff
@@ -35,3 +48,158 @@ def auth_credentials(fs):
...
+    """Select basic auth and return a TestClient."""
+    monkeypatch.delenv("RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI", raising=False)
+    monkeypatch.delenv("RALPH_RUNSERVER_AUTH_OIDC_AUDIENCE", raising=False)
+    reload(conf)
+    app = reload(api).app
+    return TestClient(app)
+
+
+@pytest.fixture
+def oidc_client(monkeypatch):
```
jmaupetit, line 64: I would suggest something more explicit:

```suggestion
def oidc_test_client(monkeypatch):
```

tests/fixtures/auth.py
```diff
@@ -35,3 +48,158 @@ def auth_credentials(fs):
     )
 
     return credentials
+
+
+@pytest.fixture
+def basic_client(monkeypatch):
```
jmaupetit, line 54: Also:

```suggestion
def basic_auth_test_client(monkeypatch):
```

tests/fixtures/auth.py
```diff
@@ -35,3 +48,158 @@ def auth_credentials(fs):
...
+@pytest.fixture
+def oidc_client(monkeypatch):
+    """Select oidc auth and return a TestClient."""
+    monkeypatch.setenv("RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI", ISSUER_URI)
+    monkeypatch.setenv("RALPH_RUNSERVER_AUTH_OIDC_AUDIENCE", AUDIENCE)
+    reload(conf)
+    reload(api.auth.oidc)
+    reload(api.auth)
+    app = reload(api).app
+    return TestClient(app)
```
jmaupetit, line 72: Another idea would be to use a configurable single fixture for authenticated test client using markers [1]. WDYT?


[1] https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures

jmaupetit (commented):
Yay, almost done :muscle:

src/ralph/conf.py
wilbrdt: It is more clear indeed!

src/ralph/api/auth/oidc.py
wilbrdt: For sure for type hinting the return value, but I didn't define models, I don't know if it is worth the effort and if it won't slow down the auth.

tests/api/test_forwarding.py
wilbrdt: Good catch!

tests/fixtures/auth.py
wilbrdt: Pytest markers are mostly used for test configuration (parametrize, skipif, filterwarning etc). It would also mean adding a custom mark related to authentication in the pytest configuration under `setup.cfg`, which I'm not very keen on.

But I like the idea of having a common fixture returning a TestClient, I let you take a look, I instead used the direct test parametrization to change the fixture.

tests/api/test_forwarding.py
wilbrdt: We're still exceeding the deadline even on a large resource on this test, I leave `deadline=None` as such

docs/api.md
```diff
@@ -81,7 +94,73 @@ $ curl --user john.doe@example.com:PASSWORD http://localhost:8100/whoami
 < {"username":"john.doe@example.com","scopes":["authenticated","example_scope"]}
 ```
 
-### Forwarding statements
+### OpenID Connect authentication
+
+Ralph LRS API server supports OpenID Connect (OIDC) on top of OAuth 2.0 for authentication and authorization.
+
+To enable OIDC auth, you should set the `RALPH_RUNSERVER_AUTH_BACKEND` environment variable to `oidc`, and you should define the `RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI` environment variable with your identity provider's Issuer Identifier URI as such:
```
quitterie-lcs, line 101: I suggest to have a more visual documentation for setting environment variables. User just have to copy/paste the proposed configuration (and adapt it if necessary)

```suggestion
To enable OIDC auth, you should set the `RALPH_RUNSERVER_AUTH_BACKEND` environment variable as follows:
`
RALPH_RUNSERVER_AUTH_BACKEND=oidc

and you should define the `RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI` environment variable with your identity provider's Issuer Identifier URI as follows:
`
RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI=http://{provider_host}:{provider_port}/auth/realms/{realm_name}
`
```

docs/api.md
```diff
@@ -81,7 +94,73 @@ $ curl --user john.doe@example.com:PASSWORD http://localhost:8100/whoami
...
+`http://{provider_host}:{provider_port}/auth/realms/{realm_name}`
```
quitterie-lcs, line 102:
```suggestion
```

docs/api.md
```diff
@@ -81,7 +94,73 @@ $ curl --user john.doe@example.com:PASSWORD http://localhost:8100/whoami
...
+
+This address must be accessible to the LRS on startup as it will perform OIDC Discovery to retrieve public keys and other information about the OIDC environment.
+
+It is also strongly recommended that you set the optional `RALPH_RUNSERVER_AUTH_OIDC_AUDIENCE` environment variable to the origin address of the LRS itself (ex. "http://localhost:8100") to enable verification that a given token was issued specifically for the LRS.
+
+#### Identity Providers
+
+OIDC support is currently developed and tested against Keycloak but may work with other identity providers that implement the specification.
+
+The repository [Learning analytics playground](https://github.com/openfun/learning-analytics-playground/) contains a Docker Compose file and configuration for a demo instance of Keycloak with a `ralph` client.
```
quitterie-lcs, line 112:
```suggestion
The [Learning analytics playground](https://github.com/openfun/learning-analytics-playground/) repository contains a Docker Compose file and configuration for a demo instance of Keycloak with a `ralph` client.
```

tests/api/auth/test_basic.py
```diff
@@ -0,0 +1,162 @@
...
+        ]
+    )
+
+    # Test addition operator
+    users += other_users
+
+    # Test len
+    assert len(users) == 3
+
+    # Test getitem
```
quitterie-lcs, line 48: `Test getitem` is actually tested in `Test iterator`. Is it necessary to have both ?

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,147 @@
...
+    auth_header: Union[str, None] = Depends(oauth2_scheme)
+) -> AuthenticatedUser:
+    """Decode and validate OpenId Connect ID token against issuer in config.
+
+    Args:
+        token (str): Base64 encoded OIDC Token. This is invoked behind the
+            scenes by Depends.
+
+    Return:
+        IDToken (IDToken)
```
quitterie-lcs, line 103:
```suggestion
        AuthenticatedUser (AuthenticatedUser)
```

src/ralph/api/auth/basic.py
```diff
@@ -108,7 +97,9 @@ def get_stored_credentials(auth_file: Path) -> ServerUsersCredentials:
     return ServerUsersCredentials.parse_file(auth_file)
 
 
-def authenticated_user(credentials: HTTPBasicCredentials = Depends(security)):
+def get_authenticated_user(
+    credentials: Union[HTTPBasicCredentials, None] = Depends(security),
+) -> AuthenticatedUser:
     """Checks valid auth parameters.
```
quitterie-lcs, line 103: `Return` and `Raises` sections are missing

tests/api/auth/test_basic.py
wilbrdt: Behind the scene, it does tests different methods (`__getitem__`, `__iter__`), isn't it?

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,147 @@
...
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Could not validate credentials",
+            headers={"WWW-Authenticate": "Bearer"},
+        ) from exc
+
+    id_token = IDToken.parse_obj(decoded_token)
+
+    return AuthenticatedUser(
+        username=id_token.sub,
```
Leobouloc, line 145: Independently of the last comment. An example for an `openid` field from the xAPI spec is:

```
{
     "name": "Aaron Silvers",
     "openid": "http://aaron.openid.example.org",
     "objectType": "Agent"
}
``` 

Shouldn't this be (at least): 

```
username=id_token.iss + "/" + id_token.sub
```

(I assumed we are putting `openid` in `username`)

?

src/ralph/api/auth/oidc.py
```diff
@@ -0,0 +1,147 @@
...
+        logger.error("Unable to decode the ID token: %s", exc)
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Could not validate credentials",
+            headers={"WWW-Authenticate": "Bearer"},
+        ) from exc
+
+    id_token = IDToken.parse_obj(decoded_token)
+
+    return AuthenticatedUser(
```
Leobouloc, line 144: I think `AuthenticatedUser` should be refactored to make a distinction between the various IFI available. It would make more sense to use the `openid` IFI here. Perhaps we could add an `ifi`  or `agent` field to the class ?

tests/api/auth/test_basic.py
```diff
@@ -0,0 +1,162 @@
...
+                UserCredentials(
+                    username="foo", hash="notsorealhash", scopes=["foo_scope"]
+                ),
+            ]
+        )
+
+
+@pytest.mark.parametrize("auth_mode", ["basic"])
+def test_get_whoami_no_credentials(auth_test_client):
+    """Whoami route returns a 401 error when no credentials are sent."""
```
Leobouloc, line 74: maybe we could have a `basic_auth_test_client` fixture instead of parametrizing all tests ? (I am not sure this use is in the spirit of `parametrize` but I might be wrong).

Leobouloc (commented):
This looks neat !

Leobouloc (commented):
This looks neat !

Leobouloc (commented):
This looks neat !

tests/api/auth/test_basic.py
wilbrdt: I agree it's not in the spirit of `parametrize`, re-splitting into two distinct fixtures as I find it more clear.

src/ralph/api/auth/oidc.py
wilbrdt: Letting as it is for now, but I take note to do this for v4.0!

quitterie-lcs (approved):
Looks good to me !

PR #266, opened 2023-02-02
🏷️(library) restore python 3.7+ compatibility for models

jmaupetit:
## Purpose

Projects still running python 3.7 should be able to integrate Ralph models in their code base.

## Proposal

- [x] fix typing issues
- [x] automate library testing for various python versions from 3.7 to 3.11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ docker-login: &docker-login
         echo "$DOCKER_HUB_PASSWORD" | docker login -u "$DOCKER_HUB_USER" --password-stdin ||
         echo "Docker Hub anonymous mode"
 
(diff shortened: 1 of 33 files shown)


wilbrdt (approved):
LGTM :tada: 
Alt Text

But I really think it would need more documentation for lib users or for potential new contributors, indicating how Ralph can be used as a lib, and that only the lib parts (i.e. models) needs to be compatible with python3.7.

jmaupetit:
> But I really think it would need more documentation for lib users or for potential new contributors, indicating how Ralph can be used as a lib, and that only the lib parts (i.e. models) needs to be compatible with python3.7.

:100: yes!

PR #276, opened 2023-03-01
🐛(api) fix sentry integration

jmaupetit:
## Purpose

When the SENTRY_DSN setting is active, the LRS server cannot start as the sentry_sdk module is not imported.

## Proposal

- [x] import the `sentry_sdk` in the LRS server module


SergioSim (approved):
Well Spotted! Looks good) :+1: 

Could we remove the `# noqa: F821` comment on line 12 of `src/ralph/api/__init__.py`?

image

jmaupetit:
> Could we remove the # noqa: F821 comment on line 12 of src/ralph/api/__init__.py?

Well spotted!

PR #281, opened 2023-03-08
🐛(api) fix ralph runserver on pre-built images

bmtcril:
When `ralph runserver` was run from the command line on a pre-built image uvicorn threw an error to prevent forking recursion. This check allows startup to continue as expected.

## Purpose

Fix a bug with running the `ralph` executable instead of via `python -m ralph` on pre-built images resulting in uvicorn throwing a `RuntimeError` while forking the process:

`An attempt has been made to start a new process before the current process has finished its bootstrapping phase.`

## Proposal

Wrap the `cli` command in a check that we're running from `__main__` to prevent the recursion.


bmtcril:
@jmaupetit updated, thank you!

jmaupetit (approved):

PR #289, opened 2023-03-21
Implement synchronous LRS backend

quitterie-lcs:
## Purpose

In FUN perspective, we want to use our LRS in `warren` backend and intends to
make available `warren` for every one. Hence it should not directly depend on
the FUN made LRS. 

## Proposal

In the perspective of having a generic call to any LRS in data application, we
propose the implementation in the backends stack of `ralph` a new instance of
backend: the LRS one. It will allow anyone to be able to use any LRS in one's
application

Implementation to do list: 
- [x] add http base backend
- [x] test http base backend
- [x] add lrs backend (get, post methods)
- [x] test lrs backend
- [x] add development settings
- [x] update cli commands
- [x] update CI configuration
- [x] update documentation for `ralph`
- [x] update CHANGELOG.md

### DIFF
--- a/docs/backends.md
+++ b/docs/backends.md
@@ -183,7 +183,6 @@ The MongoDB client options supported in Ralph are:
 - `document_class`: default class to use for documents returned from queries
 - `tz_aware`: if True, datetime instances returned as values in a document will be timezone aware (otherwise they will be naive)
 
-
 ### ClickHouse
 
 The ClickHouse backend can be used as a data lake and to fetch collections of
@@ -192,6 +191,7 @@ documents from it.
 #### Backend parameters
 
 ClickHouse parameters required to connect are:
+
 - `host`: the connection hostname to connect to (_e.g._ `"clickhouse.yourhost.com"`)
 - `port`: the port to the ClickHouse HTTPS interface (_e.g._ `8123`)
 - `database`: the name of the database to connect to
@@ -199,14 +199,34 @@ ClickHouse parameters required to connect are:
 - `client_options`: a comma separated key=value list of ClickHouse client options
 
 Secondary parameters are needed if not using the default ClickHouse user:
+
 - `username`: the username to connect as
 - `password`: the password for the given ClickHouse username
 
 By default, the following client options are set, if you override the default 
 client options you must also set these:
+
 - `"date_time_input_format": "best_effort"` allows RFC date parsing
 - `"allow_experimental_object_type": 1` allows the JSON data type we use to store statements
 
 The ClickHouse client options supported in Ralph can be found in these locations:
 - [Python driver specific](https://clickhouse.com/docs/en/integrations/language-clients/python/driver-api#settings-argument)
 - [General ClickHouse client settings](https://clickhouse.com/docs/en/operations/settings/settings/)
+
+## HTTP backends
+
+### LRS
+
+The LRS backend is used to store and retrieve xAPI statements from various systems mostly used in e-learning infrastructures.
+
+#### Backend parameters
+
+LRS parameters required to connect are: 
+
+- `url`: the URL to connect to the server (_e.g._ [`http://ralph:secret@lrs:8100`])
+- `username`: the username to connect as
+- `password`: the password for the given LRS username
+
+Optional parameters can be configured if necessary: 
+
+- `headers`: a comma-separated key=value list of LRS server headers
--- a/setup.cfg
+++ b/setup.cfg
@@ -81,7 +81,8 @@ dev =
     pylint==2.17.4
     pytest==7.3.2
     pytest-asyncio==0.21.0
-    pytest-cov==4.1.0
+    pytest-cov==4.0.0
+    pytest-httpx==0.22.0
 ci =
     twine==4.0.2
 lrs =
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,6 +47,9 @@ backend-es =
 backend-ldp =
     ovh>=1.0.0
     requests>=2.0.0
+backend-lrs =
+    httpx==0.24.1
+    more-itertools==9.1.0
 backend-mongo =
     pymongo[srv]>=4.0.0
     python-dateutil>=2.8.2
--- a/src/ralph/backends/http/lrs.py
+++ b/src/ralph/backends/http/lrs.py
@@ -1,0 +1,275 @@
+"""LRS HTTP backend for Ralph."""
+
+import json
+import logging
+from itertools import chain
+from typing import Iterable, Iterator, List, Optional, Union
+from urllib.parse import ParseResult, parse_qs, urljoin, urlparse
+
+import httpx
+from httpx import Client, HTTPError, HTTPStatusError, RequestError
+from more_itertools import chunked
+from pydantic import AnyHttpUrl, BaseModel, parse_obj_as
+
+from ralph.conf import LRSHeaders, settings
+from ralph.exceptions import BackendException, BackendParameterException
+
+from .base import (
+    BaseHTTP,
+    BaseQuery,
+    HTTPBackendStatus,
+    OperationType,
+    enforce_query_checks,
+)
+
+lrs_settings = settings.BACKENDS.HTTP.LRS
+logger = logging.getLogger(__name__)
+
+
+class StatementResponse(BaseModel):
+    """Pydantic model for `get` statements response."""
+
+    statements: Union[List[dict], dict]
+    more: Optional[str]
+
+
+class LRSQuery(BaseQuery):
+    """LRS body query model."""
+
+    query: Optional[dict]
+
+
+class LRSHTTP(BaseHTTP):
+    """LRS HTTP backend."""
+
+    name = "lrs"
+    query = LRSQuery
+    default_operation_type = OperationType.CREATE
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        base_url: str = lrs_settings.BASE_URL,
+        username: str = lrs_settings.USERNAME,
+        password: str = lrs_settings.PASSWORD,
+        headers: LRSHeaders = lrs_settings.HEADERS,
+        status_endpoint: str = lrs_settings.STATUS_ENDPOINT,
+        statements_endpoint: str = lrs_settings.STATEMENTS_ENDPOINT,
+    ):
+        """Instantiates the LRS client.
+
+        Args:
+            base_url (AnyHttpUrl): LRS server URL.
+            username (str): Basic auth username for LRS authentication.
+            password (str): Basic auth password for LRS authentication.
+            headers (dict): Headers defined for the LRS server connection.
+            status_endpoint (str): Endpoint used to check server status.
+            statements_endpoint (str): Default endpoint for LRS statements resource.
+        """
+        self.base_url = parse_obj_as(AnyHttpUrl, base_url)
+        self.auth = (username, password)
+        self.headers = headers
+        self.status_endpoint = status_endpoint
+        self.statements_endpoint = statements_endpoint
+
+    def status(self):
+        """Implements HTTP backend check for server status."""
+        status_url = urljoin(self.base_url, self.status_endpoint)
+
+        try:
+            response = httpx.get(status_url)
+            response.raise_for_status()
+        except RequestError:
+            msg = "Unable to request the server"
+            logger.error(msg)
+            return HTTPBackendStatus.AWAY
+        except HTTPStatusError:
+            msg = "Response raised an HTTP status of 4xx or 5xx"
+            logger.error(msg)
+            return HTTPBackendStatus.ERROR
+
+        return HTTPBackendStatus.OK
+
+    def list(
+        self, target: str = None, details: bool = False, new: bool = False
+    ) -> Iterator[Union[str, dict]]:
+        # noqa: D205, D415
+        """LRS HTTP backend does not support `list` method,
+        calling this method will raise an error.
+        """
+        msg = "LRS HTTP backend does not support `list` method, cannot list from %s"
+        logger.error(msg, target)
+        raise NotImplementedError(msg % target)
+
+    @enforce_query_checks
+    def read(  # pylint: disable=too-many-arguments
+        self,
+        query: Union[str, LRSQuery] = None,
+        target: str = None,
+        chunk_size: Union[None, int] = 500,
+        raw_output: bool = False,
+        ignore_errors: bool = False,
+    ) -> Iterator[Union[bytes, dict]]:
+        """Get statements from LRS `target` endpoint.
+
+        The `read` method defined in the LRS specification returns `statements` array
+        and `more` IRL. The latter is required when the returned `statement` list has
+        been limited.
+
+        Args:
+            query (str, LRSQuery):  The query to select records to read.
+            target (str): Endpoint from which to read data (e.g. `/statements`).
+                If target is `None`, `/xAPI/statements` default endpoint is used.
+            chunk_size (int or None): The number of records or bytes to read in one
+                batch, depending on whether the records are dictionaries or bytes.
+            raw_output (bool): Controls whether to yield bytes or dictionaries.
+                If the records are dictionaries and `raw_output` is set to `True`, they
+                are encoded as JSON.
+                If the records are bytes and `raw_output` is set to `False`, they are
+                decoded as JSON by line.
+            ignore_errors (bool): If `True`, errors during the read operation
+                are be ignored and logged. If `False` (default), a `BackendException`
+                is raised if an error occurs.
+        """
+        if not target:
+            target = self.statements_endpoint
+
+        query_params = {}
+        if query.query:
+            query_params.update(query.query)
+
+        # Set chunk_size to limit parameter if not exist in the query
+        query_params.setdefault("limit", chunk_size)
+
+        # Create request URL
+        target = ParseResult(
+            scheme=urlparse(self.base_url).scheme,
+            netloc=urlparse(self.base_url).netloc,
+            path=target,
+            query="",
+            params="",
+            fragment="",
+        ).geturl()
+
+        def fetch_statements(params: dict):
+            while True:
+                with Client(auth=self.auth, headers=self.headers) as client:
+                    response = client.get(target, params=params)
+                    response.raise_for_status()
+
+                    statements_response = StatementResponse.parse_obj(response.json())
+                    statements = statements_response.statements
+                    statements = (
+                        [statements] if not isinstance(statements, list) else statements
+                    )
+                    if raw_output:
+                        for statement in statements:
+                            yield bytes(json.dumps(statement), encoding="utf-8")
+                    else:
+                        for statement in statements:
+                            yield statement
+
+                    if not statements_response.more:
+                        break
+
+                    params.update(parse_qs(urlparse(statements_response.more).query))
+
+        try:
+            yield from fetch_statements(params=query_params)
+        except HTTPError as error:
+            msg = "Failed to fetch statements."
+            logger.error("%s. %s", msg, error)
+            raise BackendException(msg, *error.args) from error
+
+    def write(  # pylint: disable=too-many-arguments
+        self,
+        data: Union[Iterable[bytes], Iterable[dict]],
+        target: Union[None, str] = None,
+        chunk_size: Union[None, int] = 500,
+        ignore_errors: bool = False,
+        operation_type: Union[None, OperationType] = None,
+    ) -> int:
+        """Writes `data` records to the `target` endpoint and returns their count.
+
+        Args:
+            data: (Iterable): The data to write.
+            target (str or None): Endpoint in which to write data (e.g. `/statements`).
+                If `target` is `None`, `/xAPI/statements` default endpoint is used.
+            chunk_size (int or None): The number of records or bytes to write in one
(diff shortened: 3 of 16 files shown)


src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,89 @@
...
+
+        The `get` method defined in the LRS spefication returns `statements` array and
+        `more` IRL. The latter is required when the `statement` list returned has been
+        limited.
+        """
+
+        origin = urlparse(target).scheme + "://" + urlparse(target).netloc
+
+        def fetch_statements(url, origin=origin):
+            with Client(auth=self.auth, headers=self.headers) as client:
```
Leobouloc, line 51: This part (`Client(auth=self.auth, headers=self.headers)`) will return an error if `self.auth == (None, None)` which would be the case if no `user:password` is not provided in the `connection_url`.

Perhaps this should be a bit more generic ? I am not sure the `connection_url` in the class instantiation is necessary.

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        statements = 0
+
+        try:
+            for chunk in chunked(data, chunk_size):
+                with Client(auth=self.auth, headers=self.headers) as client:
+                    request = client.post(self.origin + target, content=chunk)
+                    request.raise_for_status()
+                    statements += len(chunk)
+                    logger.debug("Posted %d statements", statements)
```
wilbrdt, line 100:
```suggestion
                    logger.debug("Posted %d statements", statements_count)
```

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,106 @@
...
+        except HTTPError as error:
+            msg = "Failed to post statements"
+            logger.error("%s. %s", msg, error)
+            raise BackendException(msg, *error.args) from error
+
+        return statements
```
wilbrdt, line 106:
```suggestion
        return statements_count
```

wilbrdt (commented):
Alt text

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,106 @@
...
+        target: str,
+        data: List[Union[bytes, dict]],
+        chunk_size: Union[None, int] = 500,
+        ignore_errors: bool = False,
+    ) -> int:
+        """Post statements to LRS.
+
+        Stores in an LRS a statement or a set of statements.
+        """
+        statements = 0
```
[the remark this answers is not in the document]
Leobouloc: I agree with the naming change, but I wonder if it is the appropriate output for this function, as I am not sure how this would be used.

As is, the function will either raise an Exception, or return the number of statements provided by the user (which the user already has access to). Returning a count would seem more appropriate if we would also return the count even in the event of an exception (something like `{"status": "error", "statements_posted_count": 1000; "remaining_statements_count": 150}`), but this would prevent us from raising an error.

Maybe we could return the output of the requests, in case the LRS provides feedback ?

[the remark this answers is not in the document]
wilbrdt: The objective is to be able to use Ralph as a FastAPI server that can GET and POST statements from other LRS. If you look at the POST statements route [here](https://github.com/openfun/ralph/blob/639e2456a44b9acd0e196a57dbf4567f9b063fc7/src/ralph/api/routers/statements.py#L349), we use the number of statements posted for logging information.

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return HTTPBackendStatus.OK
+
+    def get(self, target: str, chunk_size: Union[None, int] = None):
+        """Get statements from LRS.
+
+        The `get` method defined in the LRS spefication returns `statements` array and
+        `more` IRL. The latter is required when the `statement` list returned has been
+        limited.
+        """
```
jmaupetit, line 62: The `target` parameter is not documented.

jmaupetit (commented):
Looks great! :muscle:

jmaupetit:
Also: maybe we should move to the unified backends interface!

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,106 @@
...
+        self.origin = (
+            urlparse(self.connection_url).scheme
+            + "://"
+            + urlparse(self.connection_url).netloc
+        )
+
+    def status(self):
+        """Implements HTTP backend check for server status."""
+        try:
+            response = httpx.get(self.origin + "/__heartbeat__")
```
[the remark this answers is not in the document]
wilbrdt: This is the way

src/ralph/conf.py
```diff
@@ -80,34 +80,41 @@ def get_instance(self, **init_parameters):
...
     class Config:  # pylint: disable=missing-class-docstring # noqa: D106
         extra = Extra.forbid
 
 
+class ClickhouseClientOptions(BaseModel):
+    """Pydantic model for `clickhouse` client options."""
+
+    class Config:  # pylint: disable=missing-class-docstring # noqa: D106
+        extra = Extra.allow
+
```
wilbrdt, line 95: We could add the parameters used as default in the backend implementation:
```suggestion
class ClickhouseClientOptions(ClientOptions):
    """Pydantic model for `clickhouse` client options."""
    
    date_time_input_format: str = "best_effort"
    allow_experimental_object_type: Literal[0,1] = 1
```

src/ralph/conf.py
```diff
@@ -145,6 +152,39 @@ class DatabaseBackendSettings(BaseModel):
...
+    """Pydantic model for additionnal client options."""
+
+    class Config:  # pylint: disable=missing-class-docstring # noqa: D106
+        extra = Extra.allow
+
+
+class LRSHeaders(HeaderParameters):
+    """Pydantic model for LRS headers."""
+
+    X_EXPERIENCE_API_VERSION: str = "1.0.0"
```
wilbrdt, line 168: Why not the last version as default?
```suggestion
    X_EXPERIENCE_API_VERSION: str = "1.0.3"
```

tests/backends/http/test_lrs.py
```diff
@@ -0,0 +1,521 @@
...
+    )
+
+    result = backend.read()
+    assert list(result) == statements["statements"]
+
+
+def test_backends_http_lrs_read_method_without_statements(httpx_mock: HTTPXMock):
+    """Test the LRS backend `read` method without statements in the target endpoint
+    raises a `BackendException`.
+    """
```
Leobouloc, line 163: It seems like the name and description do not match what the test does. I suggest:

`test_backends_http_lrs_read_backend_error`

tests/backends/http/test_lrs.py
```diff
@@ -0,0 +1,521 @@
...
+        json=more_statements_with_query_played_verb,
+    )
+
+    # Return an iterable of dict
+    result = backend.read(query=query_params, target=target, raw_output=False)
+    assert (
+        list(result)
+        == statements_with_query_played_verb["statements"]
+        + more_statements_with_query_played_verb["statements"]
+    )
```
Leobouloc, line 378: perhaps add assertion on number of results for security.

tests/backends/http/test_lrs.py
```diff
@@ -0,0 +1,521 @@
...
+    """
+
+    url = "http://fake-lrs.com"
+    target = "/xAPI/statements"
+
+    backend = LRSHTTP(url=url, basic_username="user", basic_password="pass")
+
+    msg = "Delete operation_type is not supported."
+    with pytest.raises(BackendParameterException, match=msg):
+        backend.write(target=target, data=[b"foo"], operation_type=OperationType.DELETE)
```
Leobouloc, line 468: For conciseness, these could be parametrized.

Leobouloc (commented):
Seems good to me :) I left a few minor comments and suggestions.

Leobouloc (commented):
Seems good to me :) I left a few minor comments and suggestions.

docs/backends.md
```diff
@@ -191,21 +190,40 @@ documents from it.
...
+
+### LRS
+
+The LRS backend is used to store and retrieve xAPI statements from various systems mostly used in e-learning infrastructures.
+
+#### Backend parameters
+
+LRS parameter required to connect is: 
+
+- `connection_url`: the connection URL to connect to the server (_e.g._ [`http://ralph:secret@lrs:8100`])
```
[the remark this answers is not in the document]
quitterie-lcs: The doc was not updated, as I split the `connection_url` into several parameters.

src/ralph/cli.py
```diff
@@ -130,6 +130,28 @@ def convert(self, value, param, ctx):
         return self.client_options_type(**super().convert(value, param, ctx))
 
 
+class HeadersParametersParamType(CommaSeparatedKeyValueParamType):
```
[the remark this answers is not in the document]
quitterie-lcs: We keep the pattern from `click.ParamType` as it inherits from

tests/backends/http/test_lrs.py
quitterie-lcs: Without statements, it shoud return an empty list, it is not the correct behavior being described

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,254 @@
...
+            query (str, LRSQuery):  The query to select records to read.
+            target (str): Endpoint from which to read data (e.g. `/statements`).
+                If target is `None`, `/xAPI/statements` default endpoint is used.
+            chunk_size (int or None): The number of records or bytes to read in one
+                batch, depending on whether the records are dictionaries or bytes.
+            raw_output (bool): Controls whether to yield bytes or dictionaries.
+                If the records are dictionaries and `raw_output` is set to `True`, they
+                are encoded as JSON.
+                If the records are bytes and `raw_output` is set to `False`, they are
+                decoded as JSON by line.
```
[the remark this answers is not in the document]
quitterie-lcs: We set this parameter in backend unification, to yield either bytes or dictionaries
https://github.com/openfun/ralph/blob/unify-data-backends/src/ralph/backends/data/base.py#LL174C13-L178C41

tests/backends/http/test_lrs.py
```diff
@@ -0,0 +1,521 @@
...
+        X_EXPERIENCE_API_VERSION="1.0.3", CONTENT_TYPE="application/json"
+    )
+    status_endpoint = "/fake-status-endpoint"
+    backend = LRSHTTP(
+        url="http://fake-lrs.com",
+        basic_username="user",
+        basic_password="pass",
+        headers=headers,
+        status_endpoint=status_endpoint,
+    )
```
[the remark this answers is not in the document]
quitterie-lcs: Maybe it's something we can together discuss, it has to be taken into account in the backend unification

SergioSim (commented):
This looks great :+1: 
I have only minor questions/suggestions)

image

setup.cfg
```diff
@@ -47,6 +47,8 @@ backend-es =
 backend-ldp =
     ovh>=1.0.0
     requests>=2.0.0
+backend-lrs = 
+    more-itertools==9.1.0
```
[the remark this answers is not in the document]
quitterie-lcs: Good point !

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,254 @@
...
+
+        return HTTPBackendStatus.OK
+
+    def list(
+        self, target: str = None, details: bool = False, new: bool = False
+    ) -> Iterator[Union[str, dict]]:
+        # noqa: D205, D415
+        """LRS HTTP backend does not support `list` method,
+        calling this method will raise an error.
+        """
```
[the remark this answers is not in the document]
quitterie-lcs: The LRS backend we are implementing is dedicared to statements. It means that the method would return only '/xAPI/statements' endpoint. 

From Ralph LRS server side, we could be able to implement a method to be able to track all implemented endpoints but what about other LRSs? 

I am not sure it seems relevant as it depends too much from the LRS and as the backend is statement-oriented. However, the message returned can be enrich with an info about '/xAPI/statements' being the available endpoint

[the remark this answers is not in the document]
SergioSim: Good point) I thought the scope would be supporting more xAPI routes. Indeed, if solely `/xAPI/statements` is used, there is little added value to list available routes :+1:

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,255 @@
...
+        target = ParseResult(
+            scheme=urlparse(self.url).scheme,
+            netloc=urlparse(self.url).netloc,
+            path=target,
+            query="",
+            params="",
+            fragment="",
+        ).geturl()
+
+        def fetch_statements(url, query_params):
```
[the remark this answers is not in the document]
quitterie-lcs: That's true. With modifications applied on the initial code, this parameter is no longer necessary

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,255 @@
...
+        data = chain((first_record,), data)
+
+        logger.debug(
+            "Start writing to the %s endpoint (chunk size: %d)", target, chunk_size
+        )
+
+        if operation_type in [OperationType.INDEX, OperationType.CREATE]:
+            with Client(auth=self.auth, headers=self.headers) as client:
+                for chunk in chunked(data, chunk_size):
+                    request = client.post(target, content=chunk)
```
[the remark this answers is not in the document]
quitterie-lcs: Fixed by adding a static method to `_parse_dict_to_bytes`

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,134 @@
...
+                return HTTPBackendStatus.AWAY
+            if response.status_code in (500, 404):
+                return HTTPBackendStatus.ERROR
+
+        return HTTPBackendStatus.OK
+
+    def read(
+        self,
+        target: str = None,
+        chunk_size: Union[None, int] = None,
```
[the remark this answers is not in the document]
quitterie-lcs: Set in the `query_param` dict value with `setdefault()` method

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,279 @@
...
+                If `operation_type` is `None`, the `default_operation_type` is used
+                instead. See `OperationType`.
+        """
+        statements_count = 0
+
+        data = iter(data)
+        try:
+            first_record = next(data)
+        except StopIteration:
+            logger.info("Data Iterator is empty; skipping write to target.")
```
jmaupetit, line 211:
```suggestion
            logger.debug("Data Iterator is empty; skipping write to target.")
```

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,279 @@
...
+                request = client.post(target, content=chunk)
+                try:
+                    request.raise_for_status()
+                except HTTPError as error:
+                    msg = "Failed to post statements"
+                    logger.error("%s. %s", msg, error)
+                    raise BackendException(msg, *error.args) from error
+                statements_count += len(chunk)
+                logger.debug("Posted %d statements", statements_count)
+
```
jmaupetit, line 263: You can log (info) the total number of statements posted once it's done.

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,279 @@
...
+        if operation_type == OperationType.UPDATE:
+            msg = "Update operation_type is not supported."
+            logger.error(msg)
+            raise BackendParameterException(msg)
+
+        if operation_type == OperationType.DELETE:
+            msg = "Delete operation_type is not supported."
+            logger.error(msg)
+            raise BackendParameterException(msg)
```
jmaupetit, line 230:
```suggestion
        if operation_type in (OperationType.APPEND, OperationType.UPDATE, OperationType.DELETE):
            msg = f"{operation_type.value} operation type is not supported."
            logger.error(msg)
            raise BackendParameterException(msg)
```

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,279 @@
...
+    """Pydantic model for `get` statements response."""
+
+    statements: Union[List[dict], dict]
+    more: Optional[str]
+
+
+class LRSQuery(BaseQuery):
+    """LRS body query model."""
+
+    query: Optional[dict]
```
jmaupetit, line 39: I think we can be more specific about expected/allowed query parameters, WDYT?

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,279 @@
...
+        except HTTPError as error:
+            msg = "Failed to fetch statements."
+            logger.error("%s. %s", msg, error)
+            raise BackendException(msg, *error.args) from error
+
+    def write(  # pylint: disable=too-many-arguments
+        self,
+        data: Union[List[bytes], List[dict]],
+        target: Union[None, str] = None,
+        chunk_size: Union[None, int] = 500,
```
jmaupetit, line 185: Should be configurable?

.env.dist
```diff
@@ -71,6 +71,10 @@ RALPH_BACKENDS__DATABASE__CLICKHOUSE__TEST_HOST=clickhouse
 RALPH_BACKENDS__DATABASE__CLICKHOUSE__TEST_PORT=8123
 RALPH_BACKENDS__DATABASE__CLICKHOUSE__TEST_TABLE_NAME=test_xapi_events_all
 
+# LRS HTTP backend
+
+RALPH_BACKENDS__HTTP__LRS__CONNECTION_URL=http://ralph:secret@localhost:8100/
+
```
wilbrdt, line 77: Seems unused now?
```suggestion
```

wilbrdt (commented):
Just some small remarks, but otherwise it looks good to me!

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,277 @@
...
+    ) -> Iterator[Union[bytes, dict]]:
+        """Get statements from LRS `target` endpoint.
+
+        The `read` method defined in the LRS specification returns `statements` array
+        and `more` IRL. The latter is required when the returned `statement` list has
+        been limited.
+
+        Args:
+            query (str, LRSQuery):  The query to select records to read.
+            target (str): Endpoint from which to read data (e.g. `/statements`).
```
Leobouloc, line 120: perhaps include a helper for users that forget the slash ? (seems like it could be an easy mistake to make)

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,277 @@
...
+        target: Union[None, str] = None,
+        chunk_size: Union[None, int] = 500,
+        ignore_errors: bool = False,
+        operation_type: Union[None, OperationType] = None,
+    ) -> int:
+        """Writes `data` records to the `target` endpoint and returns their count.
+
+        Args:
+            data: (Iterable): The data to write.
+            target (str or None): Endpoint where to write data (e.g. `/statements`).
```
Leobouloc, line 197: same suggestion: write helper ?

src/ralph/cli.py
```diff
@@ -426,23 +452,31 @@ def convert(from_, to_, ignore_errors, fail_on_unknown, **conversion_set_kwargs)
     default=settings.DEFAULT_BACKEND_CHUNK_SIZE,
     help="Get events by chunks of size #",
 )
+@click.option(
+    "-t",
+    "--target",
+    type=str,
+    default=None,
+    help="Target on the backend where to fetch events",
```
Leobouloc, line 460:
```suggestion
    help=Endpoint from which to fetch events (e.g. `/statements`)",
```

src/ralph/backends/http/lrs.py
Leobouloc, line 197:
```suggestion
            target (str or None): Endpoint in which to write data (e.g. `/statements`).
```

src/ralph/cli.py
```diff
@@ -495,7 +543,14 @@ def fetch(backend, archive, chunk_size, query, **options):
     is_flag=True,
     help="Continue writing regardless of raised errors",
 )
-def push(backend, archive, chunk_size, force, ignore_errors, **options):
+@click.option(
+    "-t",
+    "--target",
+    type=str,
+    default=None,
+    help="Target on the backend where to push events",
```
Leobouloc, line 551:
```suggestion
    help="Endpoint in which to push events (e.g. `statements`)",
```

src/ralph/backends/http/lrs.py
quitterie-lcs: The `target` parameter is used in the `ParseResult` method. Whether the slash is missing, the target is correctly parsed for the URL

src/ralph/backends/http/lrs.py
```diff
@@ -0,0 +1,277 @@
...
+        try:
+            yield from fetch_statements(params=query_params)
+        except HTTPError as error:
+            msg = "Failed to fetch statements."
+            logger.error("%s. %s", msg, error)
+            raise BackendException(msg, *error.args) from error
+
+    def write(  # pylint: disable=too-many-arguments
+        self,
+        data: Union[List[bytes], List[dict]],
```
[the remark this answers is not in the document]
quitterie-lcs: I would prefer to keep an `Iterable` to be able to pass stream as input

wilbrdt (approved):
Alt text

PR #293, opened 2023-04-06
👽️(backends) fix double quoting in ClickHouse backend server parameters

SergioSim:
## Purpose

In the current `clickhouse-connect` release (0.5.19) (see [CHANGELOG](https://github.com/ClickHouse/clickhouse-connect/blob/main/CHANGELOG.md#bug-fixes)), the quoting and escaping of array literals in server parameters is handled internally. Thus, we do not need to add quotes on our side.

## Proposal

- [x] remove quoting of `ids` in ClickHouse backend

### DIFF
--- a/tests/models/test_converter.py
+++ b/tests/models/test_converter.py
@@ -372,7 +372,7 @@ def test_converter_convert_with_invalid_page_close_event_raises_an_exception(
             list(result)
 
 
-@settings(suppress_health_check=(HealthCheck.function_scoped_fixture,))
+@settings(deadline=None, suppress_health_check=(HealthCheck.function_scoped_fixture,))
 @pytest.mark.parametrize("valid_uuid", ["ee241f8b-174f-5bdb-bae9-c09de5fe017f"])
 @pytest.mark.parametrize("invalid_platform_url", ["", "not an URL"])
 @custom_given(UIPageClose)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to
 ### Fixed
 
 - An issue with starting Ralph in pre-built Docker containers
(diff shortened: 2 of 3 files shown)


tests/models/test_converter.py
```diff
@@ -372,7 +372,7 @@ def test_converter_convert_with_invalid_page_close_event_raises_an_exception(
             list(result)
 
 
-@settings(suppress_health_check=(HealthCheck.function_scoped_fixture,))
+@settings(deadline=None, suppress_health_check=(HealthCheck.function_scoped_fixture,))
```
quitterie-lcs, line 375: Why did you remove this ? Useless?

SergioSim: I got the following error during the first run in the CI:

```
Unreliable test timings! On an initial run, this test took 308.32ms, which exceeded the deadline of 200.00ms, but on a subsequent run it took 2.27 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.

You can reproduce this example by temporarily adding @reproduce_failure('6.70.2', b'AXicY2BkYAAiKGBkQObBmIzIHHSAQxgTAAAC6QAJ') as a decorator on your test case
============ 1 failed, 775 passed, 55 warnings in 145.33s (0:02:25) ============
```

I suspect the "suspicious" speedup flagged by `hypothesis` is due to caching performed in our `custom_given` fixture. However, I'm not 100% sure; unfortunately, this error only occurs sometimes.

wilbrdt (approved):

jmaupetit (approved):
Don't forget to fill the CHANGELOG!

SergioSim:
> Don't forget to fill the CHANGELOG!

Well spotted, thanks) updated the CHANGELOG)

PR #297, opened 2023-04-13
🏗️(backends) unify data backends s3

wilbrdt:
## Purpose

Storage and Database backends had similar interfaces and usage, a new Data backend interface has been created.

## Proposal

Update S3 backend to follow new Data backend interface.
~~Add pydantic model for a history entry.~~ &rarr; moved to #317 

- [x] Add S3DataBackend
- [x] Add tests for S3

### DIFF
--- a/src/ralph/backends/data/s3.py
+++ b/src/ralph/backends/data/s3.py
@@ -1,0 +1,388 @@
+"""S3 data backend for Ralph."""
+
+import json
+import logging
+from io import IOBase
+from itertools import chain
+from typing import Iterable, Iterator, Union
+from uuid import uuid4
+
+import boto3
+from boto3.s3.transfer import TransferConfig
+from botocore.exceptions import (
+    ClientError,
+    ParamValidationError,
+    ReadTimeoutError,
+    ResponseStreamingError,
+)
+from botocore.response import StreamingBody
+from requests_toolbelt import StreamingIterator
+
+from ralph.backends.data.base import (
+    BaseDataBackend,
+    BaseDataBackendSettings,
+    BaseOperationType,
+    BaseQuery,
+    DataBackendStatus,
+    enforce_query_checks,
+)
+from ralph.backends.mixins import HistoryMixin
+from ralph.conf import BaseSettingsConfig
+from ralph.exceptions import BackendException, BackendParameterException
+from ralph.utils import now
+
+logger = logging.getLogger(__name__)
+
+
+class S3DataBackendSettings(BaseDataBackendSettings):
+    """S3 data backend default configuration.
+
+    Attributes:
+        ACCESS_KEY_ID (str): The access key id for the S3 account.
+        SECRET_ACCESS_KEY (str): The secret key for the S3 account.
+        SESSION_TOKEN (str): The session token for the S3 account.
+        ENDPOINT_URL (str): The endpoint URL of the S3.
+        DEFAULT_REGION (str): The default region used in instantiating the client.
+        DEFAULT_BUCKET_NAME (str): The default bucket name targeted.
+        DEFAULT_CHUNK_SIZE (str): The default chunk size for reading and writing
+            objects.
+        LOCALE_ENCODING (str): The encoding used for writing dictionaries to objects.
+    """
+
+    class Config(BaseSettingsConfig):
+        """Pydantic Configuration."""
+
+        env_prefix = "RALPH_BACKENDS__DATA__S3__"
+
+    ACCESS_KEY_ID: str = None
+    SECRET_ACCESS_KEY: str = None
+    SESSION_TOKEN: str = None
+    ENDPOINT_URL: str = None
+    DEFAULT_REGION: str = None
+    DEFAULT_BUCKET_NAME: str = None
+    DEFAULT_CHUNK_SIZE: int = 4096
+    LOCALE_ENCODING: str = "utf8"
+
+
+class S3DataBackend(HistoryMixin, BaseDataBackend):
+    """S3 data backend."""
+
+    name = "s3"
+    default_operation_type = BaseOperationType.CREATE
+    settings_class = S3DataBackendSettings
+
+    def __init__(self, settings: settings_class = None):
+        """Instantiate the AWS S3 client."""
+        self.settings = settings if settings else self.settings_class()
+
+        self.default_bucket_name = self.settings.DEFAULT_BUCKET_NAME
+        self.default_chunk_size = self.settings.DEFAULT_CHUNK_SIZE
+        self.locale_encoding = self.settings.LOCALE_ENCODING
+        self._client = None
+
+    @property
+    def client(self):
+        """Create a boto3 client if it doesn't exist."""
+        if not self._client:
+            self._client = boto3.client(
+                "s3",
+                aws_access_key_id=self.settings.ACCESS_KEY_ID,
+                aws_secret_access_key=self.settings.SECRET_ACCESS_KEY,
+                aws_session_token=self.settings.SESSION_TOKEN,
+                region_name=self.settings.DEFAULT_REGION,
+                endpoint_url=self.settings.ENDPOINT_URL,
+            )
+        return self._client
+
+    def status(self) -> DataBackendStatus:
+        """Implement data backend checks (e.g. connection, cluster status).
+
+        Return:
+            DataBackendStatus: The status of the data backend.
+        """
+        try:
+            self.client.head_bucket(Bucket=self.default_bucket_name)
+        except ClientError:
+            return DataBackendStatus.ERROR
+
+        return DataBackendStatus.OK
+
+    def list(
+        self, target: str = None, details: bool = False, new: bool = False
+    ) -> Iterator[Union[str, dict]]:
+        """List objects for the target bucket.
+
+        Args:
+            target (str or None): The target bucket to list from.
+                If target is `None`, the `default_bucket_name` is used instead.
+            details (bool): Get detailed object information instead of just object name.
+            new (bool): Given the history, list only unread files.
+
+        Yields:
+            str: The next object name. (If details is False).
+            dict: The next object details. (If details is True).
+
+        Raises:
+            BackendException: If a failure occurs.
+        """
+        if target is None:
+            target = self.default_bucket_name
+
+        objects_to_skip = set()
+        if new:
+            objects_to_skip = set(self.get_command_history(self.name, "read"))
+
+        try:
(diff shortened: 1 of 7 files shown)


wilbrdt:
Depends on #228

tests/backends/data/test_s3.py
```diff
@@ -0,0 +1,547 @@
...
+        s3_backend()
+    except Exception:  # pylint:disable=broad-except
+        pytest.fail("S3 backend should not raise exception on successful connection")
+
+
+@mock_s3
+def test_backends_data_s3_data_backend_status_method(
+    s3_backend,
+):  # pylint:disable=unused-argument, invalid-name
+    """Test the `FSDataBackend.status` method."""
```
quitterie-lcs, line 42:
```suggestion
    """Test the `S3DataBackend.status` method."""
```

SergioSim (commented):
Looks great) :+1: 
Have only minor questions/suggestions)
image

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,362 @@
...
+        if operation_type in [
+            BaseOperationType.APPEND,
+            BaseOperationType.DELETE,
+            BaseOperationType.UPDATE,
+        ]:
+            msg = "%s operation_type is not allowed."
+            logger.error(msg, operation_type.name)
+            raise BackendParameterException(msg % operation_type.name)
+
+        if operation_type in [BaseOperationType.CREATE, BaseOperationType.INDEX]:
```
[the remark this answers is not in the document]
wilbrdt: You're right!

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,362 @@
...
+            if target_object in list(self.list(target=target_bucket)):
+                msg = "%s already exists and overwrite is not allowed for operation %s"
+                logger.error(msg, target_object, operation_type)
+                if not ignore_errors:
+                    raise BackendException(msg % (target_object, operation_type))
```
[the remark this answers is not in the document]
wilbrdt: You're right!

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,362 @@
...
+            BackendException: If a failure during the read operation occurs and
+                `ignore_errors` is set to `False`.
+            BackendParameterException: If a backend argument value is not valid and
+                `ignore_errors` is set to `False`.
+        """
+        if query.query_string is None:
+            msg = "Invalid query. The query should be a valid object name."
+            logger.error(msg)
+            if not ignore_errors:
+                raise BackendParameterException(msg)
```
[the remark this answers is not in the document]
wilbrdt: Indeed!

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,362 @@
...
+        except ClientError as err:
+            error_msg = err.response["Error"]["Message"]
+            msg = "Failed to download %s: %s"
+            logger.error(msg, query.query_string, error_msg)
+            if not ignore_errors:
+                raise BackendException(msg % (query.query_string, error_msg)) from err
+
+        reader = self._read_raw if raw_output else self._read_dict
+        for chunk in reader(response["Body"], chunk_size, ignore_errors):
+            yield chunk
```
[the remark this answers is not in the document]
wilbrdt: You're right! I think it should be recoverable on the next chunks/lines, so we could silence the exception.

tests/backends/data/test_s3.py
```diff
@@ -0,0 +1,568 @@
...
+    monkeypatch.setattr("ralph.backends.data.s3.now", lambda: freezed_now)
+
+    object_name = "new-archive.gz"
+    some_content = b"some contents in the stream file to upload"
+
+    s3 = s3_backend()
+    s3.clean_history(lambda *_: True)
+
+    response = s3.write(
+        data=[some_content],
```
SergioSim, line 481:
```suggestion
        data=[some_content, some_content, some_content],
```
Could you try to test the write method with multiple chunks?
Isn't the `put_object` method overwriting the last chunk on each iteration?
If that's the case, maybe we could use the `upload_fileobj` method instead (converting `data` to a stream if it's an Iterable)?

tests/backends/data/test_s3.py
```diff
@@ -0,0 +1,568 @@
...
+
+    body = b"some contents in the body"
+
+    s3_client.put_object(
+        Bucket=bucket_name,
+        Key="2022-09-29.gz",
+        Body=body,
+    )
+
+    # fs.create_file(settings.HISTORY_FILE, contents=json.dumps([]))
```
SergioSim, line 304:
```suggestion
```

SergioSim (commented):
Looks Cool) :+1: 
Have only one question about the use of `put_object` in the `write` method.
image

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,360 @@
...
+            for chunk in reader(response["Body"], chunk_size, ignore_errors):
+                yield chunk
+        except (ReadTimeoutError, ResponseStreamingError) as err:
+            error_msg = err.response["Error"]["Message"]
+            msg = "Failed to read chunk from object %s: %s"
+            logger.error(msg, query.query_string, error_msg)
+            if not ignore_errors:
+                raise BackendException(msg % (query.query_string, error_msg)) from err
+
+        # Archive fetched, add a new entry to the historyResponseStreamingError
```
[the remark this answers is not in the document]
wilbrdt: :sweat_smile:

tests/backends/data/test_s3.py
wilbrdt: Dang, I don't know what I was doing, but you are right, thank you for spotting this! Reverting to using `upload_fileobj`, with a conversion from an iterable to a file-like object. Let me know what you think!

quitterie-lcs (approved):

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,360 @@
...
+                file_obj.write(
+                    bytes(f"{json.dumps(chunk)}\n", encoding=self.locale_encoding)
+                )
+                count += 1
+        else:
+            for chunk in data:
+                file_obj.write(chunk)
+                count += 1
+
+        file_obj.seek(0)
```
SergioSim, line 312: Storing a copy of the whole file in memory (`BytesIO`) might be problematic when working with large files.
I think it would be preferable to write the `data` content to a temporary file on disk and then use it with `upload_file`.
Alternatively, maybe we could implement a class with a `read` method wrapping the `data` iterable?
WDYT?

Also, `data` could be an instance of `IOBase`, if that's the case (and it contains binary data), could we pass it directly to `upload_fileobj`?

wilbrdt: Using now `StreamingIterator` from `requests-toolbelt`, to be able to use `upload_fileobj`, let me know what you think of it!

SergioSim (commented):
Awesome! That's great news that it works with `StreamingIterator`! Thanks for going the extra mile)
Have only one final question about the support of Iterators that are generators.

image

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,372 @@
...
+
+        Return:
+            int: The number of written objects.
+
+        Raise:
+            BackendException: If a failure during the write operation occurs.
+            BackendParameterException: If a backend argument value is not valid.
+        """
+        try:
+            first_record = next(iter(data))
```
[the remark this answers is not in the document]
wilbrdt: You're right I forgot about streams being consumed. Reverting to using `chain`!

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,372 @@
...
+            raise BackendParameterException(msg % operation_type.name)
+
+        if target_object in list(self.list(target=target_bucket)):
+            msg = "%s already exists and overwrite is not allowed for operation %s"
+            logger.error(msg, target_object, operation_type)
+            raise BackendException(msg % (target_object, operation_type))
+
+        logger.info("Creating archive: %s", target_object)
+
+        count = sum(1 for _ in data)
```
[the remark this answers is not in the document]
wilbrdt: Thanks for pointing this out!
Let me know what you think!

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,386 @@
...
+            raise BackendParameterException(msg)
+
+        if not chunk_size:
+            chunk_size = self.default_chunk_size
+
+        if target is None:
+            target = self.default_bucket_name
+
+        try:
+            response = self.client.get_object(Bucket=target, Key=query.query_string)
```
SergioSim, line 198: I'm curious about the `get_object` method - is the returned `StreamingBody()` fetched in lazy or greedy manner?
If it's greedy, we might want to use (in future) the [download_fileobj](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/download_fileobj.html) method, which can be configured with `multipart_chunksize`.

SergioSim (approved):
Looks great! :+1: 

image

src/ralph/backends/data/s3.py
```diff
@@ -0,0 +1,386 @@
...
+        # Using StreamingIterator from requests-toolbelt but without specifying a size
+        # as we will not use it. It implements the `read` method for iterators.
+        data = StreamingIterator(0, data)
+
+        try:
+            self.client.upload_fileobj(
+                Bucket=target_bucket,
+                Key=target_object,
+                Fileobj=data,
+            )
```
SergioSim, line 320: It seems the [upload_fileobj](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/upload_fileobj.html) method can be configured with a [TransferConfig](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.TransferConfig) object including a `multipart_chunksize` parameter similar to our `chunk_size` parameter.

src/ralph/backends/data/s3.py
wilbrdt: It seems to return a `StreamingBody` stream that you would then `iter_lines` or `iter_chunks` over.
Wouldn't using `download_fileobj` means storing the whole file into memory? It would be faster though indeed, for files that can fit into memory.

SergioSim: You are right, indeed the interface of the `download_fileobj` unfortunately doesn't let us yield results by chunks - `get_object` should be the right choice

PR #305, opened 2023-04-20
🏗️(backends) unify data backends ldp

SergioSim:
## Purpose

## Proposal

- [x] add LDPDataBackend

### DIFF
--- a/src/ralph/backends/data/ldp.py
+++ b/src/ralph/backends/data/ldp.py
@@ -1,0 +1,264 @@
+"""OVH's LDP data backend for Ralph."""
+
+import logging
+from typing import Iterable, Iterator, Literal, Union
+
+import ovh
+import requests
+
+from ralph.backends.data.base import (
+    BaseDataBackend,
+    BaseDataBackendSettings,
+    BaseOperationType,
+    BaseQuery,
+    DataBackendStatus,
+    enforce_query_checks,
+)
+from ralph.backends.mixins import HistoryMixin
+from ralph.conf import BaseSettingsConfig
+from ralph.exceptions import BackendException, BackendParameterException
+from ralph.utils import now
+
+logger = logging.getLogger(__name__)
+
+
+class LDPDataBackendSettings(BaseDataBackendSettings):
+    """OVH LDP (Log Data Platform) data backend default configuration.
+
+    Attributes:
+        APPLICATION_KEY (str): The OVH API application key (AK).
+        APPLICATION_SECRET (str): The OVH API application secret (AS).
+        CONSUMER_KEY (str): The OVH API consumer key (CK).
+        DEFAULT_STREAM_ID (str):  The default stream identifier to query.
+        ENDPOINT (str): The OVH API endpoint.
+        REQUEST_TIMEOUT (int): HTTP request timeout in seconds.
+        SERVICE_NAME (str): The default LDP account name.
+    """
+
+    class Config(BaseSettingsConfig):
+        """Pydantic Configuration."""
+
+        env_prefix = "RALPH_BACKENDS__DATA__LDP__"
+
+    APPLICATION_KEY: str = None
+    APPLICATION_SECRET: str = None
+    CONSUMER_KEY: str = None
+    DEFAULT_STREAM_ID: str = None
+    ENDPOINT: Literal[
+        "ovh-eu",
+        "ovh-us",
+        "ovh-ca",
+        "kimsufi-eu",
+        "kimsufi-ca",
+        "soyoustart-eu",
+        "soyoustart-ca",
+    ] = "ovh-eu"
+    REQUEST_TIMEOUT: int = None
+    SERVICE_NAME: str = None
+
+
+class LDPDataBackend(HistoryMixin, BaseDataBackend):
+    """OVH LDP (Log Data Platform) data backend."""
+
+    name = "ldp"
+    settings_class = LDPDataBackendSettings
+
+    def __init__(self, settings: settings_class = None):
+        """Instantiate the OVH LDP client.
+
+        Args:
+            settings (LDPDataBackendSettings or None): The data backend settings.
+                If `settings` is `None`, a default settings instance is used instead.
+        """
+        self.settings = settings if settings else self.settings_class()
+        self.service_name = self.settings.SERVICE_NAME
+        self.stream_id = self.settings.DEFAULT_STREAM_ID
+        self.timeout = self.settings.REQUEST_TIMEOUT
+        self._client = None
+
+    @property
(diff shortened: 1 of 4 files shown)


src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Raises:
+            BackendException: If a failure during the read operation occurs.
+            BackendParameterException: If the `query` argument is not an archive name.
+        """
+        if query.query_string is None:
+            msg = "Invalid query. The query should be a valid archive name"
+            raise BackendParameterException(msg)
+
+        if not raw_output or not ignore_errors:
```
SergioSim, line 160: For now, only reading raw (binary) output is implemented.
Do we decompress the (.gz/.zip/.7z or .zst and potentially PGP encrypted) archive and JSON parse the (GELF or plain) content if `raw_output` is `False`?

src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,244 @@
...
+
+    @enforce_query_checks
+    def read(
+        self,
+        *,
+        query: Union[str, BaseQuery] = None,
+        target: str = None,
+        chunk_size: Union[None, int] = 4096,
+        raw_output: bool = True,
+        ignore_errors: bool = True,
```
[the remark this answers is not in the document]
wilbrdt: I agree for `raw_output = True`, as we always returns bytes, but I would have think that we don't ignore errors, so `ignore_errors = False`. WDYT?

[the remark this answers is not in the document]
SergioSim: Ok, let's keep `raw_output = True`

I like the reasoning that if a boolean argument is not used (has any effect), then its default value should be `False`.
On the other hand, setting `raw_output = True` would be helpful to indicate the only supported behavior.
Then, regarding `ignore_errors`, IMO, the behavior would be closer to `True` than `False`, although both work:
We don't ignore errors, which implies that we handle them, but since we don't have any errors to handle, we don't handle them and thus ignore them.

src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,244 @@
...
+    class Config(BaseSettingsConfig):
+        """Pydantic Configuration."""
+
+        env_prefix = "RALPH_BACKENDS__DATA__LDP__"
+
+    APPLICATION_KEY: str = None
+    APPLICATION_SECRET: str = None
+    CONSUMER_KEY: str = None
+    DEFAULT_STREAM_ID: str = None
+    ENDPOINT: str = "ovh-eu"
```
SergioSim, line 47: Looks like the rebase discarded some of the fixup commits, will push them back

src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def list(
+        self, target: str = None, details: bool = False, new: bool = False
+    ) -> Iterator[Union[str, dict]]:
+        """List archives for a given target stream_id.
+
+        Args:
+            target (str or None): The target stream_id where to list the archives.
+                If target is `None`, the `DEFAULT_STREAM_ID` is used instead.
+            details (bool): Get detailed archive information instead of just ID's.
```
quitterie-lcs, line 106:
```suggestion
            details (bool): Get detailed archive information in addition to archive IDs.
```

src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,259 @@
...
+            dict: If `details` is True.
+
+        Raises:
+            BackendParameterException: If the `target` is `None` and no
+                `DEFAULT_STREAM_ID` is given.
+            BackendException: If a failure during retrieval of archives list occurs.
+        """
+        list_archives_endpoint = self._get_archive_endpoint(stream_id=target)
+        logger.debug("List archives endpoint: %s", list_archives_endpoint)
+        logger.debug("List archives details: %s", str(details))
```
quitterie-lcs, line 120: I think it's better to log this info it if the `details` parameter is set to True

quitterie-lcs, line 120: Shouldn't it be on `INFO` level ?

src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,259 @@
...
+
+        try:
+            archives = self.client.get(list_archives_endpoint)
+        except ovh.exceptions.APIError as error:
+            msg = "Failed to get archives list: %s"
+            logger.error(msg, error)
+            raise BackendException(msg % error) from error
+
+        logger.debug("Found %d archives", len(archives))
```
quitterie-lcs, line 129: Ditto about `INFO` level

tests/backends/data/test_ldp.py
```diff
@@ -0,0 +1,770 @@
...
+        """Mock the OVH client get request."""
+        # pylint: disable=unused-argument
+        return {"filename": "archive_name", "size": 10}
+
+    backend = ldp_backend()
+
+    monkeypatch.setattr(requests, "get", mock_requests_get)
+    backend = ldp_backend()
+    monkeypatch.setattr(backend, "_url", lambda *_: "/")
+    monkeypatch.setattr(backend.client, "get", mock_get)
```
quitterie-lcs, line 470: Could `requests_mock` fixture fit to mock the `requests` response in this testcase? :point_right: [requests_mock from Pytest](https://requests-mock.readthedocs.io/en/latest/pytest.html)

SergioSim: That's a great idea) This would indeed simplify the current tests that use the MockResponse instead.
However, I'm hesitating about adding a new dependency to `setup.cfg` in this branch as this file frequently changes (to avoid merge conflicts). Could this be done after or at the end of the v4 release? WDYT?

src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,259 @@
...
+        """Instantiate the OVH LDP client."""
+        self.settings = settings if settings else self.settings_class()
+        self.service_name = self.settings.SERVICE_NAME
+        self.stream_id = self.settings.DEFAULT_STREAM_ID
+        self.timeout = self.settings.REQUEST_TIMEOUT
+        self._client = None
+
+    @property
+    def client(self):
+        """Create an ovh.Client if it doesn't exists."""
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks)

tests/backends/data/test_ldp.py
quitterie-lcs: That's not a problem for me. We can add it as an improvement in an issue to not forget to do it

Issue #315, opened 2023-05-03
Wrong "more" url on the second request or when using query parameters


wilbrdt:
## Bug Report

**Expected behavior/code**
When making a GET request to the "/xAPI/statements" endpoint, a new endpoint for the next "page" of statements is received via the "more" key on the response.

Making a second GET request to this new endpoint, should return a new response with another valid endpoint in the "more" key.

**Actual Behavior**
An invalid endpoint in the "more" key is returned, it seems to be missing the `?`.

**Steps to Reproduce**
1. Make a simple request to the /xAPI/statements endpoint:
```
curl -s \
    --user ralph:secret \
    -H "Content-Type: application/json" \
    http://localhost:8100/xAPI/statements/ \ | jq
```
2. The response contains a "more" key:
```
"more": "/xAPI/statements?pit_id=s6vrAwEKc3RhdGVtZW50cxZ3UHZkY25GQlNPeTJfWVdiSkxLc3ZRABZ2TFhuaFBLbVNmdS0yVXRSQkdWRy1BAAAAAAAAGtelFnZ1bzM0MERaVGxTSXNDZzZvTGRObUEAARZ3UHZkY25GQlNPeTJfWVdiSkxLc3ZRAAA=&search_after=1682373022000|150"
```
3. Make a new request to the previous returned endpoint:
```
curl -s \
    --user ralph:secret \
    -H "Content-Type: application/json" \
    "http://localhost:8100/xAPI/statements?pit_id=s6vrAwEKc3RhdGVtZW50cxZ3UHZkY25GQlNPeTJfWVdiSkxLc3ZRABZ2TFhuaFBLbVNmdS0yVXRSQkdWRy1BAAAAAAAAGtelFnZ1bzM0MERaVGxTSXNDZzZvTGRObUEAARZ3UHZkY25GQlNPeTJfWVdiSkxLc3ZRAAA=&search_after=1682373022000|150" \ | jq
```
4. The "more" key on the response is as following (notice the lack of a "?" character on "statementspit_id")
```
[...]

PR #316, opened 2023-05-03
🐛(api) fix `more` irl on second request

wilbrdt:
## Purpose
Following #315, making a second GET request to the `more` irl, or querying with a query parameter, was failing due to a missing `?` after the endpoint name.

## Proposal

- [x] '?' should now be present for every `more` irl generated, even when query parameters are passed. 
- [x] Tests updated to cover this issue

### DIFF
--- a/src/ralph/api/routers/statements.py
+++ b/src/ralph/api/routers/statements.py
@@ -3,6 +3,7 @@
 import logging
 from datetime import datetime
 from typing import List, Literal, Optional, Union
+from urllib.parse import ParseResult, urlencode
 from uuid import UUID, uuid4
 
 from fastapi import (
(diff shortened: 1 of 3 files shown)


src/ralph/api/routers/statements.py
```diff
@@ -251,7 +251,7 @@ async def get(
         response.update(
             {
                 "more": (
-                    f"{path}{query + '&' if query else '?'}pit_id={query_result.pit_id}"
+                    f"{path}?{query + '&' if query else ''}pit_id={query_result.pit_id}"
```
jmaupetit, line 254: Maybe it means that we need to build the `more` URL with "more" reliable tools such as `urllib` [1]. WDYT?

[1] https://docs.python.org/3/library/urllib.parse.html

tests/api/test_statements_get.py
```diff
@@ -372,8 +380,81 @@ def test_api_statements_get_statements_with_pagination(
...
+
+    # Second response gets the missing result from the first response.
+    second_response = client.get(
+        first_response.json()["more"],
+        headers={"Authorization": f"Basic {auth_credentials}"},
+    )
+    assert second_response.status_code == 200
+    assert second_response.json()["statements"] == [statements[2], statements[1]]
     more_regex = re.compile(r"^/xAPI/statements/\?pit_id=.*&search_after=.*$")
+    assert more_regex.match(second_response.json()["more"])
```
jmaupetit, line 395: ditto: maybe we can test the parsed URL using `urllib`?

jmaupetit (commented):
Wow, that was quick. Thanks!

quitterie-lcs (commented):
LGTM

src/ralph/api/routers/statements.py
```diff
@@ -246,14 +247,19 @@ async def get(
...
+            }
+        )
+
         response.update(
             {
-                "more": (
-                    f"{path}{query + '&' if query else '?'}pit_id={query_result.pit_id}"
-                    f"&search_after={query_result.search_after}"
-                )
+                "more": urlunparse(("", "", path, "", urlencode(query), "")),
```
jmaupetit, line 262: Maybe to make it a bit more explicit:

```suggestion
                "more": ParseResult(
                    scheme="", 
                    netloc="", 
                    path=path, 
                    params="", 
                    query=urlencode(query), 
                    fragment="",
                ).geturl(),
```

jmaupetit (commented):
I have two minor remarks, thanks!

PR #317, opened 2023-05-03
💥(history) change command history

wilbrdt:
## Purpose

Add a common HistoryEntry pydantic model for command history.

## Proposal
Each backend will be adapted to use HistoryEntry when merged:
- [x] fs
- [ ] s3
- [ ] mongo
- [ ] es
- [ ] clickhouse
- [ ] ldp
- [ ] swift

### DIFF
--- a/src/ralph/backends/data/fs.py
+++ b/src/ralph/backends/data/fs.py
@@ -18,7 +18,7 @@ from ralph.backends.data.base import (
     DataBackendStatus,
     enforce_query_checks,
 )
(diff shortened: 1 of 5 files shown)


jmaupetit:
Maybe adding a feature flag on the history would be a great idea? When using a backend as a library there is a high probability that you don't need it!

wilbrdt:
Opening issue https://github.com/openfun/ralph/issues/474 to add it in the future, closing this as it is too far from `master` branch.

Issue #288, opened 2023-03-20
Add authorization scope mechanism


wilbrdt:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
With the addition of the [OpenId Connect authentication](https://github.com/openfun/ralph/pull/262) into Ralph, we should, as mentionned by the [xAPI spec](https://github.com/adlnet/xAPI-Spec), implement scopes.

**Describe the solution you'd like**
In a first implementation, we could implement the following scopes:
- `all/read`
- `all`
- `statements/write`
- `statements/read`
- `statements/read/mine`

**Discovery, Documentation, Adoption, Migration Strategy**
We will probably have to add an [Authority](https://github.com/yetanalytics/lrsql/blob/main/doc/authority.md#oidc-authority) mechanism, necessary for the `statements/read/mine` scope.

PR #318, opened 2023-05-03
(api) Implement scopes

Leobouloc:
# STALE BRANCH, PR KEPT OPEN FOR REFERENCE

## Purpose

Currently, the API has no restriction to the data / endpoints available. All authenticated users have unlimited read and write access, through all endpoints. The purpose, of this PR, (initiated by #288) is to add security mechanism including scopes and permissions. The xAPI specification suggests [these scopes](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Communication.md#details-15).

**NB:** The term "scope" is used losely, as the authentification mode (for now) is Basic Auth ("scope" is usually used in the context of OAuth). We use the term in a similar sense but applied to a user. (eg. "a user has the right to access scope `/statements/read/mine`"). 

## Proposal

Adding scopes is relatively straightforward except for the scope `/statements/read/mine`, which requires to add a notion of belonging to the statements being stored. This notion is [covered by the xAPI specification](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#231-statement-immutability) under the name "Authority", which is a field that SHOULD be created by the LRS upon receiving statements and populated with the requesting user account information. This `authority`  field may then be used to filter incoming queries and/or results. 

The proposed solution (a first pass) includes pre-processing the data to add "Authority", implementing scopes, and adding a permissions mechanism which forces user to query using an `authority` field, when the broadest scope available is `/statements/read/mine`.

The proposed pre-processing also handles annex [LRS conformity issues,](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#231-statement-immutability) by adding to the statements the fields `timestamp`, `stored` and `id` (when not present).

Description...

- [x] Add statement pre-processing on `POST` and `PUT` for `timestamp`, `stored`, `id`
- [x] Test pre-processing for `timestamp`, `stored`, `id`  for `PUT`
- [x] Test pre-processing for `timestamp`, `stored`, `id`  for `POST`
- [x] Refactor endpoint tests to work with pre-processing (`GET`, `POST`, `PUT`)
- [x] Test new helper functions in utils
- [x] Add authority to databases and query statements
- [x] Decide whether or not to authorize same authority for different users
- [x] Test querying by authority
- [x] Add statement pre-processing on `POST` and `PUT` for `authority` (separate as it is a more complex issue)
- [x] Test pre-processing for `authority` for `POST`
- [x] Test pre-processing for `authority` for `PUT`
- [x] Add "scope" (see above) mechanism to auth, to restrict access to endpoints
- [ ] Test "scope" mechanism (especially test permissions vulnerabilities)
- [ ] Add permissions mechanism to get
- [ ] Test permission mechanism
- [ ] Update CHANGELOG

### DIFF
--- a/src/ralph/api/auth.py
+++ b/src/ralph/api/auth.py
@@ -3,11 +3,11 @@
 import logging
 from functools import lru_cache
 from pathlib import Path
-from typing import List
+from typing import List, Tuple
 
 import bcrypt
 from fastapi import Depends, HTTPException, status
-from fastapi.security import HTTPBasic, HTTPBasicCredentials
+from fastapi.security import HTTPBasic, HTTPBasicCredentials, SecurityScopes
 from pydantic import BaseModel, root_validator
 from starlette.authentication import AuthenticationError
 
@@ -17,7 +17,6 @@ from ralph.conf import settings
 # with invalid credentials to something innocuous with the same method as if
 # it were a legitimate user.
 UNUSED_PASSWORD = bcrypt.hashpw(b"ralph", bcrypt.gensalt())
-
 
 security = HTTPBasic()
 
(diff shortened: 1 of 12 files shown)


src/ralph/api/auth.py
```diff
@@ -107,8 +106,50 @@ def get_stored_credentials(auth_file: Path) -> ServerUsersCredentials:
...
     return ServerUsersCredentials.parse_file(auth_file)
 
-
-def authenticated_user(credentials: HTTPBasicCredentials = Depends(security)):
+@lru_cache()
+def _scope_is_authorized(requested_scope, user_scopes: Tuple):
+    """Check if the requested scope can be accessed"""
+
+    # List of all expanded scopes
+    expanded_scopes = {
```
quitterie-lcs, line 114: Maybe we can use Pydantic to define these scopes

Leobouloc: Yes ! I will do this in a future iteration

quitterie-lcs: :sunglasses:

Leobouloc:
Stale branch. See: 
https://github.com/openfun/ralph/pull/441

PR #336, opened 2023-05-22
🏗️(backends) unify data backends es

SergioSim:
## Purpose

## Proposal

- [x] add ESDataBackend


src/ralph/backends/lrs/base.py
```diff
@@ -20,36 +24,52 @@ class StatementQueryResult:
...
 
+    settings_class = BaseLRSBackendSettings
+
     @abstractmethod
     def query_statements(self, params: StatementParameters) -> StatementQueryResult:
         """Returns the statements query payload using xAPI parameters."""
 
     @abstractmethod
-    def query_statements_by_ids(self, ids: List[str]) -> list:
+    def query_statements_by_ids(self, ids: List[str]) -> List:
```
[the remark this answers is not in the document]
SergioSim: Aye, I missed [this part](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Communication.md#211-put-statements) of the xAPI LRS spec (which we do not support for now):

> If the LRS receives a Statement with an id it already has a Statement for, it SHOULD verify the received Statement matches the existing one and SHOULD return 409 Conflict if they do not match.

Thus it seems we do need to retrieve the statements for future comparison.

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,320 @@
...
+        elif operation_type in (BaseOperationType.CREATE, BaseOperationType.INDEX):
+            for item in data:
+                yield {
+                    "_index": target,
+                    "_id": item.get("id", None),
+                    "_op_type": operation_type.value,
+                    "_source": item,
+                }
+        else:
+            # operation_type == BaseOperationType.DELETE (by exclusion)
```
wilbrdt, line 294: What about BaseOperationType.APPEND?

src/ralph/backends/lrs/es.py
wilbrdt, line 1: I like the idea of having ESLRSBackend in a separate file :1st_place_medal:

wilbrdt (commented):
Looking really good! 
Alt text

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,622 @@
...
+
+    # Find every odd item.
+    query = {"query": {"query": {"term": {"modulo": 1}}}}
+    results = list(backend.read(query=query))
+    assert len(results) == 5
+    assert results[0]["_source"]["id"] == 1
+    assert results[1]["_source"]["id"] == 3
+    assert results[2]["_source"]["id"] == 5
+    assert results[3]["_source"]["id"] == 7
+    assert results[4]["_source"]["id"] == 9
```
Leobouloc, line 329: Maybe only go up to 5 statements and remove a few lines of code ? ☁️

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,320 @@
...
+    def read(
+        self,
+        *,
+        query: Union[str, ESQuery] = None,
+        target: str = None,
+        chunk_size: Union[None, int] = 500,
+        raw_output: bool = False,
+        ignore_errors: bool = False,
+    ) -> Iterator[Union[bytes, dict]]:
+        """Reads documents matching the query in the target index and yields them.
```
Leobouloc, line 158:
```suggestion
        """Read documents matching the query in the target index and yield them.
```

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,320 @@
...
+        if "green" not in cluster_status:
+            logger.error("Cluster status is not green: %s", cluster_status)
+            return DataBackendStatus.ERROR
+
+        return DataBackendStatus.OK
+
+    def list(
+        self, target: str = None, details: bool = False, new: bool = False
+    ) -> Iterator[Union[str, dict]]:
+        """Lists available Elasticsearch indices, data streams and aliases.
```
Leobouloc, line 111:
```suggestion
        """List available Elasticsearch indices, data streams and aliases.
```

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def write(  # pylint: disable=too-many-arguments
+        self,
+        data: Union[IOBase, Iterable[bytes], Iterable[dict]],
+        target: Union[None, str] = None,
+        chunk_size: Union[None, int] = 500,
+        ignore_errors: bool = False,
+        operation_type: Union[None, BaseOperationType] = None,
+    ) -> int:
+        """Writes data documents to the target index and returns their count.
```
Leobouloc, line 210:
```suggestion
        """Write data documents to the target index and returns their count.
```

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,320 @@
...
+            raise BackendException(*error.args, f"{count} succeeded writes") from error
+        return count
+
+    @staticmethod
+    def to_documents(
+        data: Iterable[dict],
+        target: str,
+        operation_type: BaseOperationType,
+    ) -> Iterator[dict]:
+        """Converts dictionaries from `data` to ES documents and yields them."""
```
Leobouloc, line 276:
```suggestion
        """Convert dictionaries from `data` to ES documents and yield them."""
```

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,320 @@
...
+            # operation_type == BaseOperationType.DELETE (by exclusion)
+            for item in data:
+                yield {
+                    "_index": target,
+                    "_id": item.get("id", None),
+                    "_op_type": operation_type.value,
+                }
+
+    def _read_raw(self, documents: Iterable[Dict[str, Any]]) -> Iterator[bytes]:
+        """Reads the `documents` Iterable and yields bytes."""
```
Leobouloc, line 303:
```suggestion
        """Read the `documents` Iterable and yields bytes."""
```

Leobouloc (commented):
This looks very neat ! (I didn't go over all of it yet). My only real question is whether or not "yellow" should be considered OK.

I also started putting docstrings in the imperative form. These changes may be discarded.

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,622 @@
...
+    elasticsearch_node = backend.client.transport.node_pool.get()
+    assert elasticsearch_node.config.ca_certs == Path("/path/to/ca/bundle")
+    assert elasticsearch_node.config.verify_certs is True
+    assert elasticsearch_node.host == "elasticsearch_hostname"
+    assert elasticsearch_node.port == 9200
+
+    try:
+        ESDataBackend(settings)
+    except Exception as err:  # pylint:disable=broad-except
+        pytest.fail(f"ESDataBackend should not raise exceptions: {err}")
```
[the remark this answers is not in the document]
SergioSim: It's a suggestion from @quitterie-lcs mentioned in https://github.com/openfun/ralph/pull/228#discussion_r1014251182
Indeed we don't need the try/except mechanism, however, it makes the error message more obvious if an exception occurs.

src/ralph/backends/data/es.py
SergioSim: Good point) It seems that in most of ralph's modules, we currently use the descriptive docstring style.
However, in other parts we do use the imperative style.

The [google style guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) is not very strict on this part:

> The docstring may be descriptive-style ("""Fetches rows from a Bigtable.""") or imperative-style ("""Fetch rows from a Bigtable."""), but the style should be consistent within a file.

I like the idea to follow a single docstring-style across the project, was a decision already made about that?

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,320 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ESClientOptions(BaseModel):
+    """Pydantic model for Elasticsearch additional client options."""
+
+    class Config:
+        """Base query model configuration."""
+
+        extra = "forbid"
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Removed the `extra = "forbid"`
It might not work in Ralph's cli context as it requires proper type conversion from strings.
However, it's indeed useful for library usage.
Alternatively, we could define all 100ish available ES client initialization options with their proper types; however, this might be difficult to maintain.

src/ralph/backends/data/es.py
SergioSim: Good point, `BaseOperationType.APPEND` is already excluded in the `write` method.
However, if `to_documents` is called in another context, indeed, this might occur.
Should we make this method private?
Or maybe move the:
```
if operation_type == BaseOperationType.APPEND:
            msg = "Append operation_type is not supported."
            logger.error(msg)
```
from the `write` method in `to_documents`?

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,622 @@
...
+        es_status = "1664532320 10:05:20 docker-cluster green 1 1 2 2 0 0 1 0 - 66.7%"
+        elasticsearch_patch.setattr(backend.client, "info", lambda: None)
+        elasticsearch_patch.setattr(backend.client.cat, "health", lambda: es_status)
+        assert backend.status() == DataBackendStatus.OK
+
+    with monkeypatch.context() as elasticsearch_patch:
+        es_status = "1664532320 10:05:20 docker-cluster yellow 1 1 2 2 0 0 1 0 - 66.7%"
+        elasticsearch_patch.setattr(backend.client, "info", lambda: None)
+        elasticsearch_patch.setattr(backend.client.cat, "health", lambda: es_status)
+        assert backend.status() == DataBackendStatus.ERROR
```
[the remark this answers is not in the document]
SergioSim: Good point) Indeed, in some deployments, the "yellow" status might be considered `OK`.
I would suggest making it configurable (to keep users aware of the actual ES status), [see this fixup commit](https://github.com/openfun/ralph/pull/336/commits/3b8c5853027e5ce8617fce5ed233d5a8c16ad657).
WDYT?

src/ralph/backends/data/es.py
SergioSim: Thanks) yes, let's make the `to_documents` private, updated it)

tests/backends/data/test_es.py
SergioSim: Good Idea) Thanks) Updated this part)

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,321 @@
...
+                dictionary defining a search definition using the Elasticsearch Query
+                DSL. The Lucene query overrides the query DSL if present.
+                See Elasticsearch search reference for Lucene query syntax:
+                https://www.elastic.co/guide/en/elasticsearch/reference/8.6/search-search.html#search-api-query-params-q
+                See Elasticsearch search reference for query DSL syntax:
+                https://www.elastic.co/guide/en/elasticsearch/reference/8.6/search-search.html#request-body-search-query
+            target (str or None): The target Elasticsearch index name to query.
+                If target is `None`, the `default_index` is used instead.
+            chunk_size (int or None): The chunk size for reading batches of documents.
+                If chunk_size is `None` it defaults to 10.
```
wilbrdt, line 172:
```suggestion
                If chunk_size is `None` it defaults to 500.
```

SergioSim: The default is indeed `500`; however, it seems that when `chunk_size` is set explicitly to `None` the `size` body field of the request to Elasticsearch `_search` path is not added and defaults to `10`: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html#search-size-param

SergioSim: However, this is indeed confusing, should we set `chunk_size` to `500` if it is set to `None`?

wilbrdt: It would be better yes!

src/ralph/backends/data/es.py
```diff
@@ -169,7 +170,7 @@ def read(
             target (str or None): The target Elasticsearch index name to query.
                 If target is `None`, the `default_index` is used instead.
             chunk_size (int or None): The chunk size for reading batches of documents.
-                If chunk_size is `None` it defaults to 10.
+                If chunk_size is `None` it defaults to `DEFAULT_CHUNK_SIZE=500`.
```
wilbrdt, line 173:
```suggestion
                If chunk_size is `None` it defaults to `DEFAULT_CHUNK_SIZE`.
```

src/ralph/backends/data/es.py
```diff
@@ -215,6 +217,7 @@ def write(  # pylint: disable=too-many-arguments
             target (str or None): The target Elasticsearch index name.
                 If target is `None`, the `default_index` is used instead.
             chunk_size (int or None): The number of documents to write in one batch.
+                If chunk_size is `None` it defaults to `DEFAULT_CHUNK_SIZE=500`.
```
wilbrdt, line 220:
```suggestion
                If chunk_size is `None` it defaults to `DEFAULT_CHUNK_SIZE`.
```

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,636 @@
...
+    elasticsearch_node = backend.client.transport.node_pool.get()
+    assert elasticsearch_node.config.ca_certs == Path("/path/to/ca/bundle")
+    assert elasticsearch_node.config.verify_certs is True
+    assert elasticsearch_node.host == "elasticsearch_hostname"
+    assert elasticsearch_node.port == 9200
+
+    try:
+        ESDataBackend(settings)
+    except Exception as err:  # pylint:disable=broad-except
+        pytest.fail(f"ESDataBackend should not raise exceptions: {err}")
```
quitterie-lcs, line 84: We should be consistent between backend tests. I don't get why we care about checking if an exception occurs in this test. Shouldn't we use the `es_backend` fixture ? cf [mongoDB test](https://github.com/openfun/ralph/pull/372/files#diff-580781a4f0491f1f6b6f56d67c9ca90002a16eaa4e0ecc36cb09573492bf3963R56)

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,325 @@
...
+    DataBackendStatus,
+    enforce_query_checks,
+)
+from ralph.conf import BaseSettingsConfig, CommaSeparatedTuple
+from ralph.exceptions import BackendException, BackendParameterException
+
+logger = logging.getLogger(__name__)
+
+
+class ESClientOptions(BaseModel):
```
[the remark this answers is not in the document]
SergioSim: Yes, agreed) This should be done when removing the `database.es` backend as it depends on it for now.

tests/backends/data/test_es.py
SergioSim: Well spotted! Thanks) I have forgotten this test in the MongoDB pull request, I will update it.
The idea of the test is to check that instantiating a second instance of the backend does not raise an exception.
Do we update the error message to something like: "Two instances of ESDataBackend should not raise exceptions"?

src/ralph/backends/data/es.py
SergioSim: Well spotted) Thanks) updated)

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,325 @@
...
+    def __init__(self, settings: settings_class = None):
+        """Instantiate the Elasticsearch client."""
+        self.settings = settings if settings else self.settings_class()
+        self.default_index = self.settings.DEFAULT_INDEX
+        self.locale_encoding = self.settings.LOCALE_ENCODING
+        self._client = None
+
+    @property
+    def client(self):
+        """Create an Elasticsearch client if it doesn't exists."""
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks) updated this part)

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,325 @@
...
+
+class ESDataBackend(BaseDataBackend):
+    """Elasticsearch data backend."""
+
+    name = "es"
+    query_model = ESQuery
+    settings_class = ESDataBackendSettings
+
+    def __init__(self, settings: settings_class = None):
+        """Instantiate the Elasticsearch client."""
```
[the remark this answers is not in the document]
SergioSim: Thanks) Updated)

src/ralph/backends/data/ldp.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+class LDPDataBackend(HistoryMixin, BaseDataBackend):
+    """OVH LDP (Log Data Platform) data backend."""
+
+    name = "ldp"
+    settings_class = LDPDataBackendSettings
+
+    def __init__(self, settings: settings_class = None):
+        """Instantiate the OVH LDP client."""
```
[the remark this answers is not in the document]
SergioSim: Thanks) Updated this in the LDP pull request)

tests/backends/lrs/test_es.py
```diff
@@ -0,0 +1,386 @@
...
+                "track_total_hits": False,
+                "search_after": ["1686557542970", "0"],
+            },
+        ),
+    ],
+)
+def test_backends_lrs_es_lrs_backend_query_statements_query(
+    params, expected_body, es_lrs_backend, monkeypatch
+):
+    """Tests the `ESLRSBackend.query_statements` method, given valid statement
```
Leobouloc, line 225:
```suggestion
    """Test the `ESLRSBackend.query_statements` method, given valid statement
```

And so on for the other functions of this file :)

tests/backends/lrs/test_es.py
```diff
@@ -0,0 +1,386 @@
...
+def test_backends_lrs_es_lrs_backend_query_statements_with_search_query_failure(
+    es, es_lrs_backend, monkeypatch, caplog
+):
+    """Tests the `ESLRSBackend.query_statements`, given a search query failure, should
+    raise a `BackendException` and log the error.
+    """
+    # pylint: disable=invalid-name,unused-argument
+
+    def mock_search(**_):
+        """Mocks the Elasticsearch.search method."""
```
Leobouloc, line 309:
```suggestion
        """Mock the Elasticsearch.search method."""
```

tests/backends/lrs/test_es.py
```diff
@@ -0,0 +1,386 @@
...
+    logger_name = "ralph.backends.lrs.es"
+    msg = "Failed to execute ElasticSearch query: ApiError(None, 'Query error')"
+    assert caplog.record_tuples == [(logger_name, logging.ERROR, msg)]
+
+
+def test_backends_lrs_es_lrs_backend_query_statements_by_ids_with_multiple_indexes(
+    es, es_forwarding, es_lrs_backend
+):
+    """Tests the `ESLRSBackend.query_statements_by_ids` method, given a valid search
+    query, should execute the query uniquely on the specified index and return the
```
Leobouloc, line 356:
```suggestion
    query, should execute the query only on the specified index and return the
```
?

tests/backends/data/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        (None, BaseQuery()),
+        ("foo", BaseQuery(query_string="foo")),
+        (BaseQuery(query_string="foo"), BaseQuery(query_string="foo")),
+    ],
+)
+def test_backends_data_base_enforce_query_checks_with_valid_input(value, expected):
+    """Tests the enforce_query_checks function given valid input."""
```
Leobouloc, line 18:
```suggestion
    """Test the enforce_query_checks function given valid input."""
```

And so on for tests :)

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,636 @@
...
+        i for i in range(10) if i != 2
+    ]
+
+    # Given an unparsable binary JSON document, the write method should skip it.
+    data = [
+        json.dumps({"foo": "bar"}).encode("utf-8"),
+        "This is invalid JSON".encode("utf-8"),
+        json.dumps({"foo": "baz"}).encode("utf-8"),
+    ]
+    backend.write(data, chunk_size=2, ignore_errors=True)
```
Leobouloc, line 611: perhaps we could also check that `backend.write` returns the value `2` ?

Leobouloc (commented):
Reviewed mostly ES and base related tests for now. This looks very neat ! I only have minor comments. The one recurrent suggestion is to change docstrings to imperative, which is not always the case across files.

tests/backends/lrs/test_es.py
SergioSim: Well spotted) Thanks) updated the docstrings)

tests/backends/lrs/test_es.py
SergioSim: Nice catch) Thanks) Updated)

tests/backends/data/test_base.py
SergioSim: Well spotted) Thanks) Included this change in the #365 PR

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,638 @@
...
+            raise ESConnectionError("Mocked connection error")
+
+        elasticsearch_patch.setattr(backend.client, "info", mock_connection_error)
+        with caplog.at_level(logging.ERROR):
+            assert backend.status() == DataBackendStatus.AWAY
+
+    assert (
+        "ralph.backends.data.es",
+        logging.ERROR,
+        "Failed to connect to Elasticsearch: Connection error",
```
quitterie-lcs, line 131: Why is the message used in the mocked method not returned? It should be the following message, no ? 
```
Failed to connect to Elasticsearch: Mocked connection error
```

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,638 @@
...
+    """
+
+    def mock_es_scan(*args, **kwargs):
+        """Mock the ES.client.scan method always raising an exception."""
+        raise exception
+
+    backend = es_backend()
+    monkeypatch.setattr("ralph.backends.data.es.scan", mock_es_scan)
+    with pytest.raises(BackendException, match=f"Failed to read documents: {error}"):
+        next(backend.read())
```
quitterie-lcs, line 247:
```suggestion
        with caplog.at_level(logging.ERROR):
            next(backend.read())
    
    assert (
        "ralph.backends.data.es",
        logging.ERROR,
        f"Failed to read documents: {error}",
    ) in caplog.record_tuples
```

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,638 @@
...
+    results = list(backend.read(query=query))
+    assert len(results) == 1
+    assert results[0]["_source"]["id"] == 1
+
+    # Check query argument type
+    with pytest.raises(
+        BackendParameterException,
+        match="'query' argument is expected to be a ESQuery instance.",
+    ):
+        list(backend.read(query={"not_query": "foo"}))
```
quitterie-lcs, line 357:
```suggestion
        with caplog.at_level(logging.ERROR):
            list(backend.read(query={"not_query": "foo"}))

    assert (
        "ralph.backends.data.es",
        logging.ERROR,
       "'query' argument is expected to be a ESQuery instance.",
    ) in caplog.record_tuples
```

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,638 @@
...
+        json.dumps({"foo": "baz"}).encode("utf-8"),
+    ]
+
+    # By default, we should raise an error and stop the importation.
+    msg = (
+        r"Failed to decode JSON: Expecting value: line 1 column 1 \(char 0\), "
+        r"for document: b'This is invalid JSON'"
+    )
+    with pytest.raises(BackendException, match=msg) as exception_info:
+        backend.write(data, chunk_size=2)
```
quitterie-lcs, line 575:
```suggestion
    with pytest.raises(BackendException, match=msg) as exception_info:
        with caplog.at_level(logging.ERROR):
            backend.write(data, chunk_size=2)
    
    assert (
        "ralph.backends.data.es",
        logging.ERROR,
        msg,
    ) in caplog.record_tuples 
```

src/ralph/backends/data/base.py
```diff
@@ -0,0 +1,225 @@
...
+    return wrapper
+
+
+class BaseDataBackend(ABC):
+    """Base data backend interface."""
+
+    name = "base"
+    query_model = BaseQuery
+    default_operation_type = BaseOperationType.INDEX
+    settings_class = BaseDataBackendSettings
```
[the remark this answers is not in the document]
wilbrdt: I prefer the first one, i.e. `settings_class`

src/ralph/backends/lrs/es.py
```diff
@@ -0,0 +1,158 @@
...
+        return StatementQueryResult(
+            statements=[document["_source"] for document in es_documents],
+            pit_id=es_response["pit_id"],
+            search_after=search_after,
+        )
+
+    def query_statements_by_ids(self, ids: List[str]) -> List:
+        """Returns the list of matching statement IDs from the database."""
+        body = {"query": {"terms": {"_id": ids}}}
+        return self._search(index=self.default_index, body=body)["hits"]["hits"]
```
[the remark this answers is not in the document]
wilbrdt: With `Iterator[dict]` as the returned type hint!

src/ralph/backends/lrs/base.py
```diff
@@ -20,36 +24,52 @@ class StatementQueryResult:
...
 
 
 class BaseLRSBackend(BaseDataBackend):
     """Base LRS backend interface."""
 
+    settings_class = BaseLRSBackendSettings
+
     @abstractmethod
     def query_statements(self, params: StatementParameters) -> StatementQueryResult:
         """Returns the statements query payload using xAPI parameters."""
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks) This was updated in the clickhouse PR) Now rebased)

src/ralph/backends/lrs/base.py
```diff
@@ -20,36 +24,52 @@ class StatementQueryResult:
...
 
     @abstractmethod
-    def query_statements_by_ids(self, ids: List[str]) -> list:
+    def query_statements_by_ids(self, ids: List[str]) -> List:
         """Returns the list of matching statement IDs from the database."""
```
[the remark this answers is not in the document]
SergioSim: Also updated in the clickhouse PR) Now rebased)

src/ralph/backends/lrs/base.py
```diff
@@ -20,36 +24,52 @@ class StatementQueryResult:
...
+
+    mbox: Optional[str]
+    mbox_sha1sum: Optional[str]
+    openid: Optional[str]
+    account__name: Optional[str]
+    account__home_page: Optional[str]
+
+
 class StatementParameters(BaseModel):
     """Represents a dictionary of possible LRS query parameters."""
```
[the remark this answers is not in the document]
SergioSim: Also updated in the clickhouse PR) Now rebased)

src/ralph/backends/lrs/es.py
```diff
@@ -0,0 +1,158 @@
...
+    POINT_IN_TIME_KEEP_ALIVE: str = "1m"
+
+
+class ESLRSBackend(BaseLRSBackend, ESDataBackend):
+    """Elasticsearch LRS backend implementation."""
+
+    settings_class = ESLRSBackendSettings
+
+    def query_statements(self, params: StatementParameters) -> StatementQueryResult:
+        """Returns the statements query payload using xAPI parameters."""
```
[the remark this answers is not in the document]
SergioSim: Also updated in the clickhouse PR) Now rebased)

src/ralph/backends/data/es.py
```diff
@@ -0,0 +1,397 @@
...
+
+    def status(self) -> DataBackendStatus:
+        """Check Elasticsearch cluster connection and status."""
+        try:
+            self.client.info()
+        except ESConnectionError as error:
+            logger.error("Failed to connect to Elasticsearch: %s", error)
+            return DataBackendStatus.AWAY
+
+        cluster_status = self.client.cat.health()
```
SergioSim, line 148: Do we need to handle the possibility that Elasticsearch becomes unavailable just after the `self.client.info()` request?

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,638 @@
...
+    )
+
+
+def test_backends_data_es_data_backend_write_method_with_append_operation(es_backend):
+    """Test the `ESDataBackend.write` method, given an `APPEND` `operation_type`,
+    should raise a `BackendParameterException`.
+    """
+    backend = es_backend()
+    msg = "Append operation_type is not supported."
+    with pytest.raises(BackendParameterException, match=msg):
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks) updated this part)

src/ralph/backends/lrs/es.py
```diff
@@ -0,0 +1,158 @@
...
+            field = f"{target_field}.openid.keyword"
+            es_query_filters += [{"term": {field: agent_params.openid}}]
+        elif agent_params.account__name:
+            field = f"{target_field}.account.name.keyword"
+            es_query_filters += [{"term": {field: agent_params.account__name}}]
+            field = f"{target_field}.account.homePage.keyword"
+            es_query_filters += [{"term": {field: agent_params.account__home_page}}]
+
+    def _search(self, **kwargs):
+        """Wraps the ElasticSearch.search method.
```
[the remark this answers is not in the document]
SergioSim: This method is now removed as we use the `backend.read` method instead)

src/ralph/backends/lrs/es.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        try:
+            return self.client.search(**kwargs)
+        except (ApiError, ESConnectionError) as error:
+            msg = "Failed to execute ElasticSearch query: %s"
+            logger.error(msg, error)
+            raise BackendException(msg % error) from error
+
+    def _open_point_in_time(self, **kwargs):
+        """Wraps the ElasticSearch.open_point_in_time method.
```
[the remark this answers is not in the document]
SergioSim: Also removed this method - replaced by the read method)

src/ralph/conf.py
```diff
@@ -48,14 +48,14 @@ class Config(BaseSettingsConfig):
...
+    """Pydantic field type validating comma separated strings or lists/tuples."""
 
     @classmethod
     def __get_validators__(cls):  # noqa: D105
-        def validate(value: Union[str, Tuple[str]]) -> Tuple[str]:
-            """Checks whether the value is a comma separated string or a tuple."""
-            if isinstance(value, tuple):
-                return value
+        def validate(value: Union[str, Tuple[str], List[str]]) -> Tuple[str]:
+            """Checks whether the value is a comma separated string or a list/tuple."""
```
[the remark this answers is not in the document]
SergioSim: Nice catch) Thanks)

tests/backends/data/test_es.py
```diff
@@ -0,0 +1,638 @@
...
+        # the `status` method should return `DataBackendStatus.OK`.
+        backend.settings.ALLOW_YELLOW_STATUS = True
+        with caplog.at_level(logging.INFO):
+            assert backend.status() == DataBackendStatus.OK
+
+        assert (
+            "ralph.backends.data.es",
+            logging.INFO,
+            "Cluster status is yellow.",
+        ) in caplog.record_tuples
```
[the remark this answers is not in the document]
SergioSim: You are right) Updated this part)

tests/backends/data/test_es.py
SergioSim: Good point) This was due to the way the `ESConnectionError` was initialized (missing second argument), updated the test to make the error message less confusing)

src/ralph/backends/lrs/es.py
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks) Updated this part)

src/ralph/backends/data/es.py
wilbrdt: If it becomes unavailable, we reach the end of the method where we log and return a `DataBackendStatus.ERROR`, no?

wilbrdt (approved):
Alt text
Nothing to add, looks good to me, nice job!

PR #342, opened 2023-05-31
💥(cli) Add mandatory "agent" field to credentials

Leobouloc:
## Purpose

In link with #288 , we need a mechanism to assign an "Agent" ([in the sense of the xAPI specification](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#2421-when-the-actor-objecttype-is-agent)) to a user (to later be able to infer the "Authority" field when writing a statement). 

It was proposed  to add the "agent" representation as a field in the user credentials, which is what this PR implements. There is no contraint as to which of the 4 valid [IFI's](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#2423-inverse-functional-identifier) must be used.

## Proposal

Add "agent" to user credentials.

- [x] update make file with default agent (using "mbox" IFI)
- [x] update cli to create "agent" field when creating a new user
- [x] write tests 
- [x] update documentation

PR #343, opened 2023-05-31
✨(api) Add "mine" query parameter

Leobouloc:
## Purpose

In link with #288 , we need a mechanism to fetch statements by authority. It has been decided that the `GET /statements` endpoint will implement a `mine` option (not in xAPI specification), which is used to return only the statements for which the "authority" field matches the "agent" field of the current user (stored in credentials since #342). 

With future implementation of permissions, it is expected that the use of `mine` option will be made mandatory for users of the API that do not have full read access. The idea behind this is to make any filtering obvious to the user ("explicit is better than implicit").

## Proposal

Add `mine` by adding fetching user "agent" and querying using a new database query field.

- [x] add option to API
- [x] add backend query for authority
- [x] add test for API that covers `mine` option


src/ralph/api/routers/statements.py
```diff
@@ -61,11 +63,34 @@
 }
 
 
+def _parse_agent_parameters(agent_obj: dict):
```
wilbrdt, line 66: We could even LRU-cache this method, WDYT?

tests/api/test_auth.py
```diff
@@ -66,3 +66,5 @@ def test_api_auth_model_serveruserscredentials():
                 ),
             ]
         )
+
+    # TODO: should we verify agent uniqueness ?
```
wilbrdt, line 70:
```suggestion
```

tests/api/test_statements_get.py
```diff
@@ -76,36 +84,137 @@ def create_mock_agent(ifi: str, id_: int, home_page_id=None):
...
+    )
+    assert response.status_code == 200
+    assert response.json() == {"statements": [statements[0]]}
+
+    # Check that invalid parameters returns an error
+    response = client.get(
+        "/xAPI/statements/?mine=BigBoat",
+        headers={"Authorization": f"Basic {credentials_1_bis}"},
+    )
+    assert response.status_code == 422  # TODO: is this the proper code
```
wilbrdt, line 215: What would you propose? I think 422 is good

tests/fixtures/auth.py
```diff
@@ -9,17 +9,17 @@
 from ralph.conf import settings
 
 
-def create_user(fs_, user: str, pwd: str, scopes: list, agent: dict):
+def create_user(fs_, username: str, password: str, scopes: list, agent: dict):
     """Create a user using Basic Auth in the (fake) file system.
 
     Args:
         fs_: fixture provided by pyfakefs
         user: username used for auth
```
wilbrdt, line 17:
```suggestion
        username: username used for auth
```

wilbrdt (commented):
Just a few remarks but looks good!

src/ralph/api/routers/statements.py
Leobouloc: Good idea ! I will add this.

src/ralph/conf.py
```diff
@@ -320,6 +320,7 @@ class Config(BaseSettingsConfig):
         },
     }
     PARSERS: ParserSettings = ParserSettings()
+    QUERY_AGENTS_CACHE_MAX_SIZE: int = 100
```
wilbrdt, line 323:
```suggestion
```

docs/api.md
```diff
@@ -169,4 +169,18 @@ RALPH_SENTRY_LRS_TRACES_SAMPLE_RATE=0.3
...
+
+Queries by `agent` or `authority` require pre-processing of the query 
+parameters. This pre-processing may be cached to avoid unnecessary 
+computing. You can set the maximum number of entries in your cache, which 
+should be greater than the number of unique agents queried for in a 
+sequence of requests. Defaults to 100. 
+
+```
+RALPH_QUERY_AGENTS_CACHE_MAX_SIZE=100
+```
```
wilbrdt, line 186:
```suggestion
```

wilbrdt (commented):
Some reminiscences of caching, but otherwise looks good to me!

PR #349, opened 2023-06-14
⚡️(cli) allow to use a query for HTTP backends

jmaupetit:
## Purpose

When using the LRS backend to fetch statements with the CLI, it seems mandatory to be able to filter statements using a standard LRS query (`since`, `until`, `verb`, etc...).

## Proposal

- [x] activate the `query` parameter for HTTP backends


wilbrdt (approved):
Alt text

PR #365, opened 2023-06-20
🏗️(backends) unify data backends fs lrs

SergioSim:
## Purpose

We want to provide an LRS backend that could be added to Ralph's LRS without additional dependencies.

## Proposal

- [X] add FSLRSBackend

Note: depends on changes from #336

### DIFF
--- a/src/ralph/backends/lrs/fs.py
+++ b/src/ralph/backends/lrs/fs.py
@@ -1,0 +1,378 @@
+"""FileSystem LRS backend for Ralph."""
+
+import logging
+from datetime import datetime
+from io import IOBase
+from typing import Iterable, List, Literal, Union
+from uuid import UUID
+
+from ralph.backends.data.base import BaseOperationType
+from ralph.backends.data.fs import FSDataBackend, FSDataBackendSettings
+from ralph.backends.lrs.base import (
+    AgentParameters,
+    BaseLRSBackend,
+    BaseLRSBackendSettings,
+    StatementParameters,
+    StatementQueryResult,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class FSLRSBackendSettings(BaseLRSBackendSettings, FSDataBackendSettings):
+    """FileSystem LRS backend default configuration.
+
+    Attributes:
+        DEFAULT_LRS_FILE (str): The default LRS filename to store statements.
+    """
+
+    DEFAULT_LRS_FILE: str = "fs_lrs.jsonl"
+
+
+class FSLRSBackend(BaseLRSBackend, FSDataBackend):
+    """FileSystem LRS Backend."""
+
+    settings_class = FSLRSBackendSettings
+
+    def write(  # pylint: disable=too-many-arguments
+        self,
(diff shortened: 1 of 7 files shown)


src/ralph/backends/lrs/fs.py
```diff
@@ -0,0 +1,348 @@
...
+                account = agent.get("account", {})
+                if account.get("name") == name and account.get("homePage") == home_page:
+                    return True
+
+            statement_object = statement.get("object", {})
+            if statement_object.get("objectType") == "SubStatement":
+                return match_related_account(statement_object)
+            return False
+
+        if name or home_page:
```
wilbrdt, line 234: Should both `name` **and** `homepage` be provided?

src/ralph/backends/lrs/fs.py
```diff
@@ -0,0 +1,348 @@
...
+
+        def match_since(statement: dict) -> bool:
+            """Return `True` if the statement was created after `timestamp`."""
+            try:
+                statement_timestamp = datetime.fromisoformat(statement.get("timestamp"))
+            except (TypeError, ValueError) as error:
+                msg = "Statement with id=%s contains unparsable timestamp=%s"
+                logger.debug(msg, statement.get("id"), error)
+                return False
+            return statement_timestamp > timestamp
```
wilbrdt, line 294: Quick question concerning all backends: do you know why we don't return, for a `since` request, timestamp that are equals? And only for the `until` parameter

wilbrdt (commented):
Alt text
Just a few suggestions/questions

src/ralph/backends/lrs/fs.py
```diff
@@ -0,0 +1,348 @@
...
+        See `FSDataBackend.write`.
+        """
+        target = target if target else self.settings.DEFAULT_LRS_FILE
+        return super().write(data, target, chunk_size, ignore_errors, operation_type)
+
+    def query_statements(self, params: StatementParameters) -> StatementQueryResult:
+        """Return the statements query payload using xAPI parameters."""
+        filters = []
+        self._add_filter_by_id(filters, params.statementId)
+        self._add_filter_by_agent(filters, params.agent, params.related_agents)
```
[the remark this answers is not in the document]
SergioSim: Well Spotted! Thanks) added the `authority` filters)

src/ralph/backends/lrs/fs.py
SergioSim: Good point! Thanks) Yes, indeed, if one is present, the other should be too) Updated this)

src/ralph/backends/lrs/fs.py
SergioSim: Good question) We follow the [xAPI spec](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Communication.md#user-content-2.1.3.s1.table1.row9) in this case) since is exclusive and until inclusive.

wilbrdt: Oh alright thanks!

wilbrdt (approved):
LGTM! :tada:

PR #385, opened 2023-07-07
🦺(models) make ProfileActivity presence mandatory

SergioSim:
## Purpose

The latest version of the xapi video profile (1.0.3) seems to enforce the presence of a video profile activity in the
`context.contextActivities.category` field, which was previously validated to optionally contain the profile activity value.

## Proposal

- [x] make the presence of a ProfileActivity mandatory in video and virtual classroom profile statements

### DIFF
--- a/src/ralph/models/xapi/video/contexts.py
+++ b/src/ralph/models/xapi/video/contexts.py
@@ -9,7 +9,7 @@ except ImportError:
 
 from uuid import UUID
 
-from pydantic import Field, NonNegativeFloat, PositiveInt
+from pydantic import Field, NonNegativeFloat, PositiveInt, validator
 
 from ..base.contexts import BaseXapiContext, BaseXapiContextContextActivities
 from ..base.unnested_objects import BaseXapiActivity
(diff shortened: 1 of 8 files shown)


src/ralph/models/xapi/video/contexts.py
```diff
@@ -32,17 +31,13 @@
...
-    property.
+    """Pydantic model for video `context`.`contextActivities`.`category` property.
 
     Attributes:
         id (str): Consists of the value `https://w3id.org/xapi/video`.
-        definition (dict): see ProfileActivity.
     """
 
     id: Literal["https://w3id.org/xapi/video"] = "https://w3id.org/xapi/video"
-    definition: ProfileActivity
```
quitterie-lcs, line 45: You're right, it is not a mandatory field to have for validation. But we can keep it as optional, no? WDYT?

cyrillay (approved):

wilbrdt (approved):
LGMT

src/ralph/models/xapi/video/contexts.py
```diff
@@ -31,32 +25,36 @@
...
 
     category: Union[
-        VideoContextContextActivitiesCategory,
-        List[VideoContextContextActivitiesCategory],
+        VideoProfileActivity, List[Union[VideoProfileActivity, BaseXapiActivity]]
     ]
 
+    @validator("category")
+    @classmethod
+    def check_presence_of_profile_activity_categoy(
```
quitterie-lcs, line 41:
```suggestion
    def check_presence_of_profile_activity_category(
```

quitterie-lcs (commented):
Some remarks and we are ready to go !

src/ralph/models/xapi/video/contexts.py
SergioSim: Well Spotted) Thanks)

src/ralph/models/xapi/video/contexts.py
```diff
@@ -1,19 +1,13 @@
...
-
 from uuid import UUID
 
-from pydantic import Field, NonNegativeFloat, PositiveInt
+from pydantic import Field, NonNegativeFloat, PositiveInt, validator
 
 from ..base.contexts import BaseXapiContext, BaseXapiContextContextActivities
 from ..base.unnested_objects import BaseXapiActivity
-from ..concepts.activity_types.scorm_profile import ProfileActivity
+from ..concepts.activity_types.video import VideoProfileActivity
```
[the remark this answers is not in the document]
SergioSim: Good point) You are right; indeed, they are not reusable) updated this part)

quitterie-lcs (approved):

PR #386, opened 2023-07-07
✅(tests) use requests_mock fixture in LDPDataBackend tests

SergioSim:
## Purpose

We want to simplify our tests that are mocking the `request` package.

## Proposal

- [x] use the `request_mock` library to mock the request.


wilbrdt (approved):
Alt text

PR #391, opened 2023-07-12
(auth) Integration of "agent" field and "mine" query param with OpenIDConnect

wilbrdt:
## Purpose

Rebasing and integrating the works on authentication:
- On one side was added:
	- PR #342 💥(cli) add mandatory "agent" field to credentials
	- PR #343 ✨(api) add "mine" option to fetch statements with matching authority
- On the other side was added the OpenIdConnect authentication method:
	- PR #262 ✨(authentication) add OpenId Connect authentication

### DIFF
--- a/tests/fixtures/auth.py
+++ b/tests/fixtures/auth.py
@@ -1,7 +1,7 @@
 """Test fixtures related to authentication on the API."""
 import base64
 import json
-from importlib import reload
+import os
 
 import bcrypt
 import pytest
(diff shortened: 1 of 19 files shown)


tests/fixtures/auth.py
```diff
@@ -21,58 +22,120 @@
...
-    yield TestClient(app)
+    # pylint:disable=import-outside-toplevel
+    # monkeypatch.setattr(
+    #     "ralph.api.auth.oidc.settings.RUNSERVER_AUTH_OIDC_ISSUER_URI",
+    #     ISSUER_URI,
+    # )
+    # monkeypatch.setattr(
+    #     "ralph.api.auth.oidc.settings.RUNSERVER_AUTH_OIDC_AUDIENCE",
+    #     AUDIENCE,
+    # )
```
Leobouloc, line 111: should we keep this ?

Leobouloc (commented):
The code seems good to me but I think this should be merged into alpha.

Issue #383, opened 2023-07-05
Rename `fetch`/`push` in cli to `read`/`write` for consistency ?


Leobouloc:
## Feature Request

Current cli functions are called `fetch` and `push`. Perhaps we should rename them to `read` and `write` to be consistent with the unified backend terminology.

PR #392, opened 2023-07-12
(cli) Rename cli fetch & push to read & write

Leobouloc:
## Purpose

As mentionned in #383 , cli currently uses `fetch` and `push`. This could be unified with the new terms `read` and `write`.

## Proposal

- [x] change names in cli
- [x] update tests and doc
- [x] add to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- CLI: change `push` to `write` and `fetch` to `read` [BC]
+
 ## [3.9.0] - 2023-07-21
 
 ### Changed
--- a/src/ralph/backends/storage/fs.py
+++ b/src/ralph/backends/storage/fs.py
@@ -62,7 +62,7 @@ class FSStorage(HistoryMixin, BaseStorage):
         logger.debug("Found %d archives", len(archives))
 
         if new:
-            archives = set(archives) - set(self.get_command_history(self.name, "fetch"))
+            archives = set(archives) - set(self.get_command_history(self.name, "read"))
             logger.debug("New archives: %d", len(archives))
 
         for archive in archives:
@@ -86,7 +86,7 @@ class FSStorage(HistoryMixin, BaseStorage):
         self.append_to_history(
             {
                 "backend": self.name,
-                "command": "fetch",
+                "command": "read",
                 "id": name,
                 "filename": details.get("filename"),
                 "size": details.get("size"),
@@ -114,7 +114,7 @@ class FSStorage(HistoryMixin, BaseStorage):
         self.append_to_history(
             {
                 "backend": self.name,
-                "command": "push",
+                "command": "write",
                 "id": name,
                 "filename": details.get("filename"),
                 "size": details.get("size"),
--- a/src/ralph/cli.py
+++ b/src/ralph/cli.py
@@ -304,7 +304,7 @@ def backends_options(name=None, backend_types: List[BaseModel] = None):
 )
 @click.option(
     "-w",
-    "--write",
+    "--write-to-disk",
     is_flag=True,
     default=False,
     help="Write new credentials to the LRS authentication file.",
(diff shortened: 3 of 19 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ General improvement for the Helm Chart:
 - add missing namespace label
 - make object name consistent
 
+- In cli, change `push` to `write` and `fetch` to `read`
```
quitterie-lcs, line 28: You can just put this bullet point on line 13

tests/test_cli.py
```diff
@@ -1,4 +1,4 @@
-"""Tests for Ralph cli."""
+"""Test for Ralph cli."""
```
quitterie-lcs, line 1:
```suggestion
"""Tests for Ralph cli."""
```

quitterie-lcs (commented):
Looks good! Only a few comments and it is ready to be merged !
Be careful that the linter raises W0621 for `write` parameter in auth. A proper way would be to change this parameter name

CHANGELOG.md
Leobouloc: yes !

tests/test_cli.py
Leobouloc: indeed

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to
 
 ### Changed
 
+- In cli, change `push` to `write` and `fetch` to `read`
```
jmaupetit, line 15: 1. It's often a good idea to flag breaking changes.
2. Also, maybe we need to prefix change's category, _i.e._ _Helm_ or _CLI_ etc. WDYT?

```suggestion
- CLI: change `push` to `write` and `fetch` to `read` [BC]
```

src/ralph/backends/storage/fs.py
```diff
@@ -62,7 +62,7 @@ def list(self, details=False, new=False):
         logger.debug("Found %d archives", len(archives))
 
         if new:
-            archives = set(archives) - set(self.get_command_history(self.name, "fetch"))
+            archives = set(archives) - set(self.get_command_history(self.name, "read"))
```
jmaupetit, line 65: That's tricky. Should we support legacy history?

jmaupetit: Or provide an history migration script.

src/ralph/cli.py
```diff
@@ -314,7 +314,7 @@ def auth(
     username,
     password,
     scope,
-    write,
+    write_credentials,
```
jmaupetit, line 317: I don't think it's required.

Leobouloc: since the function is also named `write`, this was done to conform to linter.

CHANGELOG.md
Leobouloc: I agree

src/ralph/cli.py
jmaupetit: What about calling it `write_to_disk`?

src/ralph/backends/storage/fs.py
Leobouloc: As decided, we shall provide a migration script in the documentation. 

For memory, the alternative was to provide an additional `migrate` command to cli. Also, it was decided NOT to support legacy.

wilbrdt (commented):
Alt text
Looks good, just a couple of remarks before merging it!

src/ralph/cli.py
```diff
@@ -304,7 +304,7 @@ def wrapper(command):
 )
 @click.option(
     "-w",
-    "--write",
+    "--write-credentials",
```
Leobouloc, line 307:
```suggestion
    "--write-to-disk",
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to
 
 ### Changed
 
+- CLI: change `push` to `write` and `fetch` to `read` [BC]
```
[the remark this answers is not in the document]
Leobouloc: my bad. will do !

wilbrdt (approved):
:tada:

Issue #375, opened 2023-06-28
500 error when passing wrong query parameters to /xAPI/statements


lebrunthibault:
## Bug Report

**Expected behavior/code**
The GET query : `/xAPI/statements?toto=2` should return either a 200 response with no statements or a 400 response with an error

The [LRS spec](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Communication.md#:~:text=The%20LRS%20MUST%20reject%20with%20400%20Bad%20Request%20status%20any%20requests%20that%20use%20any%20parameters%20which%20the%20LRS%20does%20not%20recognize%20in%20their%20intended%20context%20in%20this%20specification.%20(%20Note%3A%20LRSs%20MAY%20recognize%20and%20act%20on%20parameters%20not%20in%20this%20specification)) does not specify what should be done in such case : *LRSs MAY recognize and act on parameters not in this specification* 

So we should decide if we return a 200 or 400 response.

NB : [lrsql](https://github.com/yetanalytics/lrsql) returns a 400 in such case with the following body 
```
{
    "error": {
        "message": "Invalid Params for path: /xapi/statements",
        "params": {
            "toto": "2"
        }
    }
}
```

**Actual Behavior**
This query or any query using parameters not described in the specification returns a 500 error with the following server error : `TypeError: StatementParameters.__init__() got an unexpected keyword argument 'toto'`

**Steps to Reproduce**
1. GET : `<base_url>/xAPI/statements?toto=2`

**Environment**
- Ralph version: 3.8.0
- Platform: docker compose

wilbrdt:
It should return a `400 Bad Request` for parameters unknown to Ralph indeed!

PR #399, opened 2023-07-25
(api) Return 400 status on invalid parameter in `/statements`

Leobouloc:
## Purpose

As mentionned in https://github.com/openfun/ralph/issues/375 , invalid parameters passed to the LRS should prompt a 400 error code from the API (instead of 500 as is currently the case). This PR deals with this issue for GET, POST, and PUT.

Example: `/xAPI/statements?toto=2` should return 400.

## Solution

The proposed solution would return 400 with the following message:

```
{
        "detail": "The following parameter is not allowed: `toto`"
}
```

### DIFF
--- a/src/ralph/api/forwarding.py
+++ b/src/ralph/api/forwarding.py
@@ -2,7 +2,7 @@
 
 import logging
 from functools import lru_cache
-from typing import List
+from typing import List, Literal, Union
 
 from httpx import AsyncClient, AsyncHTTPTransport, HTTPStatusError, RequestError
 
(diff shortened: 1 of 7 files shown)


Leobouloc:
CHANGELOG was not changed to avoid clutter  as this does not seem like a noteworthy change. WDYT ?

jmaupetit:
> CHANGELOG was not changed to avoid clutter as this does not seem like a noteworthy change. WDYT ?

I do think it needs to be mentioned!

jmaupetit (approved):

tests/api/test_statements_get.py
```diff
@@ -616,14 +616,24 @@ def test_api_statements_get_statements_invalid_query_parameters(
     id_1 = "be67b160-d958-4f51-b8b8-1892002dbac6"
     id_2 = "66c81e98-1763-4730-8cfc-f5ab34f1bad5"
 
-    # Test error when both statementId and voidedStatementId are provided
+    # Check for 400 status code when unknown parameters are provided
+    response = client.get(
+        "/xAPI/statements/?mamamia=herewegoagain",
```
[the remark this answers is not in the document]
Leobouloc: :)

src/ralph/api/forwarding.py
```diff
@@ -32,13 +32,16 @@ def get_active_xapi_forwardings() -> List[XapiForwardingConfigurationSettings]:
     return active_forwardings
 
 
-async def forward_xapi_statements(statements: List[dict]):
+async def forward_xapi_statements(
+    statements: List[dict], method: Literal["post", "put"]
```
wilbrdt, line 36: Would it work if we call this function with a list of statements and selecting the "put" method?
We could probably accept either a list of statements for POST or a single statement for PUT. WDYT?

wilbrdt (commented):
One small remark and good to go!

wilbrdt (approved):
Alt text

src/ralph/api/forwarding.py
Leobouloc: This was changed to:

```async def forward_xapi_statements(statements: Union[dict, List[dict]], method: Literal["post", "put"]):```

Issue #400, opened 2023-07-26
(backend) LRSHTTP is not working properly when run inside an Async function.


Leobouloc:
## Bug Report

**Expected behavior/code**
`LRSHTTP.status()` and all other methods should work when called inside an async function. 

**Actual Behavior**
As reported https://github.com/openfun/warren/pull/44#discussion_r1275127259 , the `LRSHTTP` methods returns an error when called within an events loop. The problem is caused by nesting of `asyncio.get_event_loop().run_until_complete(...)`, which is not allowed.

**Steps to Reproduce**
Add and run test:

```

@pytest.mark.anyio
async def test_backend_http_lrs_in_async_setting(monkeypatch):
    """Test that backend works when run in async function."""

    async def async_function():
        lrs = LRSHTTP()
        status_mock_response = HTTPBackendStatus.OK
        status_mock = AsyncMock(return_value=status_mock_response)
        monkeypatch.setattr(AsyncLRSHTTP, "status", status_mock)
        assert lrs.status() == status_mock_response
    
    await async_function()
```
Results in:
```
FAILED tests/backends/http/test_lrs.py::test_backend_http_lrs_in_async_setting - RuntimeError: This event loop is already running
```

**Environment**
- Ralph version: 3.9

**Possible Solution**
The first answer of [this post](https://stackoverflow.com/questions/46827007/runtimeerror-this-event-loop-is-already-running-in-python) suggest to use `nest-asyncio`.

PR #402, opened 2023-07-27
(api) disable lrshttp when called in events loop

Leobouloc:
## Purpose

As described in https://github.com/openfun/ralph/issues/400 , the current `LRSHTTP` (which is derived from `AsyncLRSHTTP`) methods can't be called from within a running events loop (as is the case in warren). There is no obvious alternative (other than code duplication), as this pattern is not allowed by `asyncio`. As dicussed [here](https://discuss.python.org/t/calling-coroutines-from-sync-code-2/24093/19), it is also considered an anti pattern.

## Proposal

It has been decided to NOT change the code structure. Instead:
- if a method is called OUTSIDE of a running events loop (in a synchronous setting), the code still works as previously.
- if a method is called INSIDE of a running events loop (meaning the user is already working asynchronously), we raise an explicit error directing the user to `AsyncLRSHTTP`. 

The error message would be:
```"This event loop is already running. You must use `AsyncLRSHTTP.status` (instead of `LRSHTTP.status`), or run this code outside the current event loop." ```

## Todo

- [x] add Exception
- [x] test that Exceptions are properly raised

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ have an authority field matching that of the user
 - API: Forwarding PUT now uses PUT (instead of POST)
 - Models: The xAPI `context.contextActivities.category` field is now mandatory
   in the video and virtual classroom profiles. [BC]
+- Backends: `LRSHTTP` methods must not be used in `asyncio` events loop (BC) 
 
 ## [3.9.0] - 2023-07-21
 
--- a/src/ralph/backends/http/lrs.py
+++ b/src/ralph/backends/http/lrs.py
@@ -4,6 +4,27 @@ import asyncio
 from ralph.backends.http.async_lrs import AsyncLRSHTTP
 
 
+def _ensure_running_loop_uniqueness(func):
+    """Raise an error when methods are used in a running asyncio events loop."""
+
+    def wrap(*args, **kwargs):
+        """Wrapper for decorator function."""
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            return func(*args, **kwargs)
+        if loop.is_running():
+            raise RuntimeError(
+                f"This event loop is already running. You must use "
+                f"`AsyncLRSHTTP.{func.__name__}` (instead of `LRSHTTP."
+                f"{func.__name__}`), or run this code outside the current"
+                " event loop."
+            )
+        return func(*args, **kwargs)
+
+    return wrap
+
+
 class LRSHTTP(AsyncLRSHTTP):
     """LRS HTTP backend."""
 
@@ -11,18 +32,21 @@ class LRSHTTP(AsyncLRSHTTP):
 
     name = "lrs"
 
+    @_ensure_running_loop_uniqueness
     def status(self, *args, **kwargs):
         """HTTP backend check for server status."""
         return asyncio.get_event_loop().run_until_complete(
             super().status(*args, **kwargs)
         )
 
+    @_ensure_running_loop_uniqueness
     def list(self, *args, **kwargs):
         """Raise error for unsuported `list` method."""
         return asyncio.get_event_loop().run_until_complete(
             super().list(*args, **kwargs)
         )
 
+    @_ensure_running_loop_uniqueness
     def read(self, *args, **kwargs):
         """Get statements from LRS `target` endpoint.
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -11,6 +11,7 @@ and this project adheres to
 ### Changed
 
 - Helm chart: improve chart modularity
+- (backends) `LRSHTTP` methods must be used outside of asyncio events loop (bc)
```
```suggestion
- backends: `LRSHTTP` methods must be used outside of `asyncio` events loop (BC) 
```

src/ralph/backends/http/lrs.py
```diff
@@ -4,25 +4,47 @@
 from ralph.backends.http.async_lrs import AsyncLRSHTTP
 
 
+def _check_for_running_loop(func):
```
jmaupetit, line 7: Maybe this would be more explicit?

```suggestion
def _ensure_running_loop_uniqueness(func):
```

src/ralph/backends/http/lrs.py
```diff
@@ -4,25 +4,47 @@
...
+
+    def wrap(*args, **kwargs):
+        """Wrapper for decorator function."""
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            return func(*args, **kwargs)
+        if loop.is_running():
+            raise RuntimeError(
+                f"This event loop is already running. You must use "
```
jmaupetit, line 18:
```suggestion
                f"An event loop is already running. You must use "
```

jmaupetit (commented):
LGTM. I only have minor remarks.

Leobouloc:
> 

Thank you @jmaupetit for the review. Do you have specific tests in mind (regarding the "needs test" label) ?

wilbrdt (approved):
Alt text 
LGTM

src/ralph/backends/http/lrs.py
```diff
@@ -4,25 +4,47 @@
...
+
+    def wrap(*args, **kwargs):
+        """Wrapper for decorator function."""
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            return func(*args, **kwargs)
+        if loop.is_running():
+            raise RuntimeError(
+                f"An event loop is already running. You must use "
```
Leobouloc, line 18:
```suggestion
                f"This event loop is already running. You must use "
```

Leobouloc: Commited this a bit too quickly. 

Reverted to "This event loop [..;]" to be consistent with the syntax of the original error raised by `asyncio`.

src/ralph/backends/http/lrs.py
```diff
@@ -4,25 +4,47 @@
...
 
+    @_check_for_running_loop
     def list(self, *args, **kwargs):
         """Raise error for unsuported `list` method."""
         return asyncio.get_event_loop().run_until_complete(
             super().list(*args, **kwargs)
         )
 
+    @_check_for_running_loop
     def read(self, *args, **kwargs):
```
Leobouloc, line 48:
```suggestion
    @_ensure_running_loop_uniqueness
    def list(self, *args, **kwargs):
        """Raise error for unsuported `list` method."""
        return asyncio.get_event_loop().run_until_complete(
            super().list(*args, **kwargs)
        )

    @_ensure_running_loop_uniqueness
    def read(self, *args, **kwargs):
```

src/ralph/backends/http/lrs.py
```diff
@@ -36,6 +58,7 @@ def read(self, *args, **kwargs):
         except StopAsyncIteration:
             pass
 
+    @_check_for_running_loop
```
Leobouloc, line 61:
```suggestion
    @_ensure_running_loop_uniqueness
```

quitterie-lcs (approved):
LGTM!

tests/backends/http/test_lrs.py
```diff
@@ -13,6 +14,48 @@
...
+            # pylint: disable=unused-argument
+            for statement in read_mock_response:
+                yield statement
+
+    else:
+        response_mock = AsyncMock(return_value=HTTPBackendStatus.OK)
+    monkeypatch.setattr(AsyncLRSHTTP, method, response_mock)
+
+    async def async_function():
+        """Encapsulate the synchroneous method in an asynchroneous function."""
```
Leobouloc, line 38:
```suggestion
        """Encapsulate the synchronous method in an asynchronous function."""
```

PR #408, opened 2023-08-07
Add close method to synchronous backends

wilbrdt:
## Purpose

Synchronous backends such as Elasticsearch or Mongodb need their connection to
be close when finished.

## Proposal

- [x] Add abstract method close to BaseDataBackend
- [x] Implement close method for ES
- [x] Implement close method for MongoDB
- [x] Implement close method for ClickHouse
- [x] Implement close method for S3
- [x] Implement close method for Swift
- [x] Raise `NotImplementedError` in close method for backends FS and LDP

### DIFF
--- a/src/ralph/backends/data/async_mongo.py
+++ b/src/ralph/backends/data/async_mongo.py
@@ -100,7 +100,7 @@ class AsyncMongoDataBackend(BaseAsyncDataBackend):
 
         try:
             collections = await database.list_collections()
-            for collection_info in collections:
+            async for collection_info in collections:
                 if details:
                     yield collection_info
                 else:
--- a/src/ralph/backends/data/async_mongo.py
+++ b/src/ralph/backends/data/async_mongo.py
@@ -256,7 +256,7 @@ class AsyncMongoDataBackend(BaseAsyncDataBackend):
         """Close the AsyncIOMotorClient client.
 
         Raise:
-            BackendException: If a failure during the close operation occurs.
+            BackendException: If a failure occurs during the close operation.
         """
         try:
             self.client.close()
--- a/src/ralph/backends/data/base.py
+++ b/src/ralph/backends/data/base.py
@@ -227,6 +227,14 @@ class BaseDataBackend(ABC):
             BackendParameterException: If a backend argument value is not valid.
         """
 
+    @abstractmethod
+    def close(self) -> None:
+        """Close the data backend client.
+
+        Raise:
+            BackendException: If a failure occurs during the close operation.
+        """
+
 
 class BaseAsyncDataBackend(ABC):
     """Base async data backend interface."""
@@ -381,5 +389,5 @@ class BaseAsyncDataBackend(ABC):
         """Close the data backend client.
 
         Raise:
-            BackendException: If a failure during the close operation occurs.
+            BackendException: If a failure occurs during the close operation.
         """
--- a/src/ralph/backends/data/clickhouse.py
+++ b/src/ralph/backends/data/clickhouse.py
@@ -382,6 +382,23 @@ class ClickHouseDataBackend(BaseDataBackend):
 
         return count
 
+    def close(self) -> None:
+        """Close the ClickHouse backend client.
+
+        Raise:
+            BackendException: If a failure occurs during the close operation.
+        """
+        if not self._client:
+            logger.warning("No backend client to close.")
+            return
+
+        try:
+            self.client.close()
+        except ClickHouseError as error:
+            msg = "Failed to close ClickHouse client: %s"
+            logger.error(msg, error)
+            raise BackendException(msg % error) from error
+
     @staticmethod
     def _to_insert_tuples(
         data: Iterable[dict],
--- a/src/ralph/backends/data/mongo.py
+++ b/src/ralph/backends/data/mongo.py
@@ -16,7 +16,13 @@ from dateutil.parser import isoparse
 from pydantic import Json, MongoDsn, constr
 from pymongo import MongoClient, ReplaceOne
 from pymongo.collection import Collection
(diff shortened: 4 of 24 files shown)


src/ralph/backends/data/async_mongo.py
```diff
@@ -100,7 +100,7 @@ async def list(
 
         try:
             collections = await database.list_collections()
-            for collection_info in collections:
+            async for collection_info in collections:
```
Leobouloc, line 103: Doesn't this mean there was (is) a missing test ? Has this been added ?

src/ralph/backends/data/base.py
```diff
@@ -227,6 +227,14 @@ def write(  # pylint: disable=too-many-arguments
             BackendParameterException: If a backend argument value is not valid.
         """
 
+    @abstractmethod
+    def close(self) -> None:
+        """Close the data backend client.
+
+        Raise:
+            BackendException: If a failure during the close operation occurs.
```
Leobouloc, line 235:
```suggestion
            BackendException: If a failure occurs during the close operation.
```

src/ralph/backends/data/clickhouse.py
```diff
@@ -382,6 +382,22 @@ def write(  # pylint: disable=too-many-arguments
 
         return count
 
+    def close(self) -> None:
+        """Close the ClickHouse backend client.
+
+        Raise:
+            BackendException: If a failure during the close operation occurs.
+        """
+        if not self._client:
```
Leobouloc, line 391: Perhaps we can add a warning to say that no client has been created ?

(applicable to all `close` methods)

tests/backends/data/test_s3.py
```diff
@@ -647,12 +663,28 @@ def test_backends_data_s3_write_method_with_failure_should_log_the_error(
...
-        s3.write(
+        backend.write(
             data=[body],
             target=object_name,
             operation_type=BaseOperationType.CREATE,
         )
+    backend.close()
+
+
+def test_backends_data_s3_data_backend_close_method(s3_backend, monkeypatch):
```
Leobouloc, line 678:
```suggestion
def test_backends_data_s3_data_backend_close_error(s3_backend, monkeypatch):
```

Specify in the name that we are testing for an error.

Leobouloc (commented):
Perhaps we should test if close actually closes the connection (checking that a call after `close` does NOT work) ? 

Other than that, it seems good to me.

src/ralph/backends/data/async_mongo.py
wilbrdt: It shouldn't have work in the first place though, I didn't understand how it could work before.
Can't see how to test this without testing the `motor` package

SergioSim: Well spotted! That was surprising to me too.
The current tests should already cover this change.
It seems `motor` fixed the bug we were relying upon in their [3.3.0 release](https://motor.readthedocs.io/en/3.3.0/changelog.html) 5 days ago [here](https://github.com/mongodb/motor/compare/3.2.0...3.3.0#diff-7aaa9d1f95fbcd82603b33bee372d43a6e1398cfe67e082d3ac158e565f2d891R547).
Thus `list_collections` now returns a `AsyncIOMotorCommandCursor` instead of a `CommandCursor`.
As this change seems to be incompatible with previous `motor` versions, could we update `setup.cfg` with `motor[srv]>=3.3.0`?

wilbrdt: @SergioSim you're right, constraining it to `>=3.3.0` !

src/ralph/backends/data/mongo.py
```diff
@@ -296,6 +302,23 @@ def write(  # pylint: disable=too-many-arguments
...
+    def close(self) -> None:
+        """Close the MongoDB backend client.
+
+        Raise:
+            BackendException: If a failure occurs during the close operation.
+        """
+        if not self.client:
+            logger.warning("No backend client to close.")
+            return
+
```
SergioSim, line 314:
```suggestion
```
Isn't `self.client` always defined in the `MongoDataBackend`?

src/ralph/backends/data/clickhouse.py
```diff
@@ -382,6 +382,23 @@ def write(  # pylint: disable=too-many-arguments
...
+
+        Raise:
+            BackendException: If a failure occurs during the close operation.
+        """
+        if not self._client:
+            logger.warning("No backend client to close.")
+            return
+
+        try:
+            self.client.close()
```
SergioSim, line 396: What about setting `self._client` to `None` afterward, to allow backends to create a new connection if requested?

SergioSim (approved):
LGTM) :+1: 
Have only two minor questions/suggestions)
image

Leobouloc (approved):
Looks good to me !

src/ralph/backends/data/mongo.py
wilbrdt: You're right!

src/ralph/backends/data/clickhouse.py
wilbrdt: Not sure if it would be useful, as we would could call the `close` method of the backend either:
- when shutting down the FastAPI server,
- before garbage collecting the backend on a CLI command.
Either way, we would have to re-instantiate a backend.
Besides, it would create differences between backends, as `mongo` and `async_mongo` create their connections when being instantiated (no `_client` attribute mechanism)
WDYT?

SergioSim: Good point) Indeed, the usefulness of this change might be quite limited, probably when used in a library context?
I guess we could address this later on if needed)

PR #416, opened 2023-08-18
Integrate unified backends

wilbrdt:
## Purpose

After unifying `database` and `storage` backends under a common `data`
interface, backends settings and methods have changed.

## Proposal

- [x] Move `Stream` backends settings alongside backend class
- [x] Move `HTTP` backends settings alongside backend class
- [x] Delete `database` and `storage` backends
- [x] Move backends settings from `conf.py` to `backends/conf.py` to avoid a
  circular import (e.g. `conf.py` -> `es.py` -> `conf.py`)
- [x] Remove `_class_path` attributes from backends settings
- [x] Change `utils.py` methods to get backend type and backend instance 
- [x] Change `cli.py` to use new unified backends
- [x] Adapt `cli.py` to be able to use async backends
- [x] Rebase and make modifications when async mongo backend is merged on `unify-data-backends`
- [x] Change router to use new unified backends (both sync and async)
- [x] Adapt CLI and API tests
- [x] Add tests for new `utils` functions
- [x] Make sure to not forget any new features/changes added on database backends only, after the work on backends unification started (thinking about PR #426 , PR #436)


.env.dist
```diff
@@ -13,63 +13,90 @@ RALPH_APP_DIR=/app/.ralph
...
+# RALPH_BACKENDS__DATA__CLICKHOUSE__CLIENT_OPTIONS__allow_experimental_object_type=
+# RALPH_BACKENDS__DATA__CLICKHOUSE__DEFAULT_CHUNK_SIZE=500
+# RALPH_BACKENDS__DATA__CLICKHOUSE__LOCALE_ENCODING=utf8
+RALPH_BACKENDS__DATA__CLICKHOUSE__TEST_DATABASE=test_statements
+RALPH_BACKENDS__DATA__CLICKHOUSE__TEST_HOST=clickhouse
+RALPH_BACKENDS__DATA__CLICKHOUSE__TEST_PORT=8123
+RALPH_BACKENDS__DATA__CLICKHOUSE__TEST_TABLE_NAME=test_xapi_events_all
+
+
+# LRS HTTP backend
```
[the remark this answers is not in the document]
wilbrdt: Method `write` have different prototypes between `data` and `http` interfaces. Should we update `data` to include `http` backends in it?

src/ralph/backends/http/__init__.py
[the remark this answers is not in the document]
wilbrdt: I harmonized with `data` backends as, for now, we don't need to import from the "package" but we can be more specific (from `ralph.backends.data.es import ESDataBackend` instead of `ralph.backends.data import ESDataBackend`). It will probably change with the plugin architecture though.

src/ralph/utils.py
```diff
@@ -37,23 +56,66 @@ def import_string(dotted_path):
...
+        raise BackendException(
+            f'No backend named "{backend_name}" '
+            f'under the backend type "{backend_type_name}"'
+        )
+
+    return backend_class
+
+
+def get_backend_instance(
+    backend_type: BaseModel,
```
Leobouloc, line 95:
```suggestion
    backend_types: Enum(BaseModel),
```
Two suggestions: "types" and `Enum`

src/ralph/api/routers/health.py
```diff
@@ -32,7 +35,7 @@ async def heartbeat():
 
     Returns a 200 if all checks are successful.
     """
-    content = {"database": DATABASE_CLIENT.status().value}
+    content = {"database": await_if_coroutine(BACKEND_CLIENT.status()).value}
```
Leobouloc, line 38: shouldn't this be awaited ?

If yes shouldn't this be breaking a test ?

src/ralph/api/routers/statements.py
```diff
@@ -416,19 +423,21 @@ async def put(
     _enrich_statement_with_authority(statement_as_dict, current_user)
 
     try:
-        existing_statement = DATABASE_CLIENT.query_statements_by_ids([statementId])
+        existing_statements = desync(
```
Leobouloc, line 426: why is `desync` necessary (if it only applies to synchronous generators) ?

src/ralph/api/routers/statements.py
```diff
@@ -416,19 +423,21 @@ async def put(
...
 
-    if existing_statement:
+    if existing_statements:
         # The LRS specification calls for deep comparison of duplicate statement ids.
         # In the case that the current statement is not equivalent to one found
         # in the database we return a 409, otherwise the usual 204.
-        for existing in existing_statement:
-            if not statements_are_equivalent(statement_as_dict, existing["_source"]):
+        async for existing in existing_statements:
+            if not statements_are_equivalent(statement_as_dict, existing):
```
Leobouloc, line 440: Just to make sure, was `_source` was removed somewhere else ? I couldn't find the origin of this change...

src/ralph/api/routers/statements.py
```diff
@@ -495,7 +507,9 @@ async def post(
         _enrich_statement_with_authority(statement, current_user)
 
     try:
-        existing_statements = DATABASE_CLIENT.query_statements_by_ids(statements_ids)
+        existing_statements = desync(
```
Leobouloc, line 510: same question here for `desync`

Leobouloc (commented):
Here is a partial review (up to `statements.py`)

src/ralph/cli.py
```diff
@@ -547,8 +553,15 @@ def convert(from_, to_, ignore_errors, fail_on_unknown, **conversion_set_kwargs)
         click.echo(event)
 
 
+read_backend_types = [
```
quitterie-lcs, line 558: Why moving this here ? Same for `write` and `list` backends

src/ralph/api/routers/statements.py
```diff
@@ -435,8 +446,8 @@ async def put(
 
     # For valid requests, perform the bulk indexing of all incoming statements
     try:
-        success_count = BACKEND_CLIENT.write(
-            data=[statement_as_dict], ignore_errors=False
+        success_count = await await_if_coroutine(
```
quitterie-lcs, line 449: Maybe its name should be changed for more readability (avoid `await` duplication)

src/ralph/backends/conf.py
```diff
@@ -0,0 +1,91 @@
...
+from ralph.backends.data.es import ESDataBackendSettings
+from ralph.backends.data.fs import FSDataBackendSettings
+from ralph.backends.data.ldp import LDPDataBackendSettings
+from ralph.backends.data.mongo import MongoDataBackendSettings
+from ralph.backends.data.s3 import S3DataBackendSettings
+from ralph.backends.data.swift import SwiftDataBackendSettings
+from ralph.backends.http.async_lrs import LRSHTTPBackendSettings
+from ralph.backends.lrs.clickhouse import ClickHouseLRSBackendSettings
+from ralph.backends.lrs.fs import FSLRSBackendSettings
+from ralph.backends.stream.ws import WSStreamBackendSettings
```
SergioSim, line 15: Aren't some backend dependencies optional?
Thus, if a backend dependency is missing, does importing it still work?
Could we load them dynamically?
For example, we could get a list of data backend module names with:
```python
map(
    lambda x: f"ralph.backends.data.{x[1]}",
    pkgutil.iter_modules(importlib.util.find_spec("ralph.backends.data").submodule_search_locations)
)
```

.env.dist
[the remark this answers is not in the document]
quitterie-lcs: In a future PR

.env.dist
```diff
@@ -13,63 +13,90 @@ RALPH_APP_DIR=/app/.ralph
...
+# RALPH_BACKENDS__DATA__S3__ENDPOINT_URL=
+# RALPH_BACKENDS__DATA__S3__DEFAULT_REGION=
+# RALPH_BACKENDS__DATA__S3__DEFAULT_BUCKET_NAME=
+# RALPH_BACKENDS__DATA__S3__DEFAULT_CHUNK_SIZE=
+# RALPH_BACKENDS__DATA__S3__LOCALE_ENCODING=
+
+# ES data backend
+
+RALPH_BACKENDS__DATA__ES__HOSTS=http://elasticsearch:9200
+RALPH_BACKENDS__DATA__ES__DEFAULT_INDEX=statements
```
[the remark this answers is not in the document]
quitterie-lcs: In .ENV, not all the variables are destined to be defaults. For some of them, it's configuration. Variables named with defaults are variables that could be set or reset by the user, so a default value can be/is necessary. Whereas variables without default in the name could not be changed as there are useful for launching services.

src/ralph/utils.py
```diff
@@ -37,23 +56,66 @@ def import_string(dotted_path):
...
-def get_backend_type(backends: BaseModel, backend_name: str):
-    """Returns the backend type from a backend name."""
+def get_backend_type(backend_types: List[BaseModel], backend_name: str):
+    """Return the backend type from a backend name."""
     backend_name = backend_name.upper()
-    for _, backend_type in backends:
+    for backend_type in backend_types:
         if hasattr(backend_type, backend_name):
             return backend_type
     return None
```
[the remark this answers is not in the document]
wilbrdt: `backend_name` would be `"es"` but `backend_type` would be a class like `BackendSettings`.
For now, we still have to differentiate in the CLI between `DATA`, `HTTP`, `LRS` or `STREAM` backends. It should not be the case after backends total unification.
An example of usage in the CLI:
```
backend_type = get_backend_type(
        backend_types=[
            backends_settings.BACKENDS.DATA,
            backends_settings.BACKENDS.HTTP,
        ],
        backend_name="es",
    )
print(f"{backend_type = }")
> backend_type = <class DataBackendSettings>
```

src/ralph/utils.py
```diff
@@ -37,23 +56,66 @@ def import_string(dotted_path):
         ) from err
 
 
-def get_backend_type(backends: BaseModel, backend_name: str):
-    """Returns the backend type from a backend name."""
+def get_backend_type(backend_types: List[BaseModel], backend_name: str):
```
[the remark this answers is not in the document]
quitterie-lcs: `BaseModel` is the common class for all backend models, so we can not be more precise. Having a list of the union of all backend models would be too verbose

src/ralph/utils.py
wilbrdt: We're only giving it one `backend_type` so not sure about the `types` remark.
Not sure about the Enum also, as we're giving it the reference to the settings class for a type, e.g. `LRSBackendSettings`

src/ralph/api/routers/health.py
wilbrdt: Indeed it should be awaited, good catch!

src/ralph/backends/conf.py
```diff
@@ -0,0 +1,91 @@
...
+class HTTPBackendSettings(BaseModel):
+    """Pydantic model for HTTP backend configuration settings."""
+
+    LRS: LRSHTTPBackendSettings = LRSHTTPBackendSettings()
+
+
+# Active LRS backend Settings.
+
+
+class LRSBackendSettings(BaseModel):
```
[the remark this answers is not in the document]
quitterie-lcs: I would add the "compatible" notion in the docstrings

src/ralph/api/routers/statements.py
wilbrdt: We don't wrap statements in a document with `{"_id": <id>, "_source": <statement>}` anymore with the new unified data backends

src/ralph/api/routers/statements.py
wilbrdt: It is a wrap to be able to iterator over either a synchronous or asynchronous generator ([https://stackoverflow.com/a/67931234/10512887](https://stackoverflow.com/a/67931234/10512887))

src/ralph/cli.py
wilbrdt: Moving it there so we stay DRY between what we give to `backends_options` for the CLI and what we use to retrieve `backend_type`. I agree that defining it as globals is not ideal!

src/ralph/api/routers/statements.py
wilbrdt: The duplication is really here! We await the wrapper -> we await if it's a coroutine else we return

src/ralph/api/routers/statements.py
```diff
@@ -509,8 +522,9 @@ async def post(
     # were skipped for being duplicates.
     # See: https://github.com/openfun/ralph/issues/345
     if existing_statements:
+        print(f"{existing_statements =}")
```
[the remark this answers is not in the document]
wilbrdt: :grin:

src/ralph/backends/conf.py
wilbrdt: Good catch!

src/ralph/api/routers/statements.py
quitterie-lcs: okay !

src/ralph/backends/lrs/base.py
```diff
@@ -69,6 +69,7 @@ class StatementParameters(BaseModel):
 class BaseLRSBackend(BaseDataBackend):
     """Base LRS backend interface."""
 
+    type = "lrs"
```
lebaudantoine, line 72: nit: if the class's name is `...LRS...`, should the type be `LRS`? Do you have any convention?

src/ralph/cli.py
```diff
@@ -570,7 +585,32 @@ def convert(from_, to_, ignore_errors, fail_on_unknown, **conversion_set_kwargs)
...
 )
-def read(backend, archive, chunk_size, target, query, **options):
+@click.option(
+    "-r",
+    "--raw-output",
+    is_flag=True,
+    show_default=True,
+    default=True,
+    help="Yield bytes instead of dictionaries.",
+)
```
SergioSim, line 595: In the cli context, is there a difference between `raw_output` set to `True` or `False`?

src/ralph/api/routers/statements.py
```diff
@@ -423,26 +423,64 @@ async def put(
...
+            # in the database we return a 409, otherwise the usual 204.
+            async for existing in existing_statements:
+                if not statements_are_equivalent(statement_as_dict, existing):
+                    raise HTTPException(
+                        status_code=status.HTTP_409_CONFLICT,
+                        detail="A different statement already exists with the same ID",
+                    )
+            return
+        else:
+            pass
```
wilbrdt, line 456:
```suggestion
        try:
            # Check if existing_statements is not empty
            first_statement = await anext(existing_statements)
        except StopIteration as exc:
            logger.debug("No statements with given ID exist: %s", exc)
            first_statement = None
            pass

        if not first_statement:
            pass
            
        existing_statements = chain((first_statement,), existing_statements)
        # The LRS specification calls for deep comparison of duplicate statement ids.
        # In the case that the current statement is not equivalent to one found
        # in the database we return a 409, otherwise the usual 204.
        async for existing in existing_statements:
            if not statements_are_equivalent(statement_as_dict, existing):
                raise HTTPException(
                    status_code=status.HTTP_409_CONFLICT,
                    detail="A different statement already exists with the same ID",
                )
        return
```

src/ralph/api/routers/statements.py
```diff
@@ -423,26 +423,64 @@ async def put(
...
+            # in the database we return a 409, otherwise the usual 204.
+            for existing in existing_statements:
+                if not statements_are_equivalent(statement_as_dict, existing):
+                    raise HTTPException(
+                        status_code=status.HTTP_409_CONFLICT,
+                        detail="A different statement already exists with the same ID",
+                    )
+            return
+        else:
+            pass
```
wilbrdt, line 483: ditto

src/ralph/backends/data/base.py
```diff
@@ -74,7 +74,7 @@ def wrapper(*args, **kwargs):
         query = kwargs.pop("query", None)
         self_ = args[0]
 
-        yield from method(*args, query=self_.validate_query(query), **kwargs)
+        return method(*args, query=self_.validate_query(query), **kwargs)
```
wilbrdt, line 77: Not sure why you changed from `yield from` to `return`. Wouldn't this lose the traceback context? cf [here](https://discuss.python.org/t/difference-between-return-generator-vs-yield-from-generator/2997/5)

wilbrdt (commented):
Alt text

src/ralph/backends/lrs/base.py
quitterie-lcs: We have chosen to keep the type in lowercase even if we use to write "LRS", as names and other backend types are often written in lowercase. We wanted to be homogeneous with what have been already implemented

Leobouloc (commented):
This is neat ! Looking forward to this being merged.

tests/test_logger.py
```diff
@@ -36,18 +36,19 @@ def test_logger_exists(fs, monkeypatch):
     }
 
     fs.create_dir("/dev")
+    fs.create_dir("foo")
```
wilbrdt, line 39: If it's really a pain to debug and make it works, you could try using `tmp_path` instead (cf. https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html )

wilbrdt (commented):
Alt text

SergioSim (commented):
Looks amazing! This is huge) Haven't reached the tests yet, will finish the review later today
Can't wait to see this merged) have only minor questions/suggestions)
image

src/ralph/backends/lrs/base.py
```diff
@@ -41,38 +40,24 @@ class AgentParameters(BaseModel):
...
-    attachments: Optional[bool]
-    ascending: Optional[bool]
+class RalphStatementsQuery(LRSStatementsQuery):
+    """Represents a dictionary of possible LRS query parameters."""
+
+    agent: Optional[AgentParameters] = AgentParameters.construct()
     search_after: Optional[str]
     pit_id: Optional[str]
-    authority: Optional[AgentParameters]
+    authority: Optional[AgentParameters] = AgentParameters.construct()
```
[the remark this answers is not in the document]
quitterie-lcs: See [here](https://github.com/openfun/ralph/commit/51554adc4abf36026675464a60dd88763f694388)

src/ralph/backends/lrs/async_es.py
```diff
@@ -21,7 +21,7 @@ class AsyncESLRSBackend(BaseAsyncLRSBackend, AsyncESDataBackend):
     settings_class = AsyncESDataBackend.settings_class
 
     async def query_statements(
-        self, params: StatementParameters
+        self, params: RalphStatementsQuery
```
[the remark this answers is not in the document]
quitterie-lcs: For more efficiency on the merging of this PR, this neat picking could be adressed in the future !

src/ralph/api/routers/health.py
```diff
@@ -32,7 +35,7 @@ async def heartbeat():
 
     Returns a 200 if all checks are successful.
     """
-    content = {"database": DATABASE_CLIENT.status().value}
+    content = {"database": await await_if_coroutine(BACKEND_CLIENT.status().value)}
```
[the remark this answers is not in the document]
quitterie-lcs: Only tested with sync backends. But tests fail with the following implementation

```
content = {"database": (await await_if_coroutine(BACKEND_CLIENT.status())).value}
```

src/ralph/api/routers/statements.py
```diff
@@ -498,7 +551,7 @@ async def post(
         )
 
     try:
-        existing_statements = DATABASE_CLIENT.query_statements_by_ids(
+        existing_statements = BACKEND_CLIENT.query_statements_by_ids(
```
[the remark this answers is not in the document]
quitterie-lcs: What are the benefits to keep it as a list ?

src/ralph/backends/data/base.py
```diff
@@ -82,6 +82,7 @@ def wrapper(*args, **kwargs):
 class BaseDataBackend(ABC):
     """Base data backend interface."""
 
+    type = "data"
```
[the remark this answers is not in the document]
quitterie-lcs: `type` variable is used in the cli to get the backend instance depending on the backend name given as input. Where would you want to use this checking ?

src/ralph/backends/http/async_lrs.py
```diff
@@ -65,41 +97,31 @@ class LRSStatementsQuery(BaseQuery):
...
+                If `settings` is `None`, a default settings instance is used instead.
         """
-        self.base_url = parse_obj_as(AnyHttpUrl, base_url)
-        self.auth = (username, password)
-        self.headers = headers
-        self.status_endpoint = status_endpoint
-        self.statements_endpoint = statements_endpoint
+        self.settings = settings if settings else self.settings_class()
+
+        self.base_url = parse_obj_as(AnyHttpUrl, self.settings.BASE_URL)
```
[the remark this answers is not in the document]
quitterie-lcs: Nope, it has been noticed when implementing the LRS backend

src/ralph/backends/lrs/async_es.py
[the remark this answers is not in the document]
SergioSim: Agreed)

src/ralph/api/routers/statements.py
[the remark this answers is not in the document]
SergioSim: 1, We catch the `BackendException` (generators are lazily evaluated - usually, they do not raise exceptions until they are iterated over)
2. Less changes in the rest of the method (as it previously worked with a list)
3. Avoid duplication (the contents of `if isasyncgen(existing_statements):` and `if isgenerator(existing_statements):` seem to be quite similar.

src/ralph/api/routers/health.py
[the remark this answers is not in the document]
SergioSim: Which tests fail with the change?

src/ralph/backends/http/async_lrs.py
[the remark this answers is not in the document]
SergioSim: Indeed, you are right; it's possible to skip the parsing part in pydantic, for example when using `setattr`.
However, can't we rely on the caller to provide a valid, well-parsed `pydantic` model to the constructor?
If not, shouldn't we potentially (re)validate/(re)parse all fields in `settings`?

src/ralph/backends/lrs/base.py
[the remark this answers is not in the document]
SergioSim: Thanks)

src/ralph/backends/http/async_lrs.py
```diff
@@ -41,41 +73,31 @@ class LRSQuery(BaseQuery):
...
+        self.auth = (self.settings.USERNAME, self.settings.PASSWORD)
```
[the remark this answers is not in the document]
Leobouloc: https://github.com/openfun/ralph/issues/454

src/ralph/cli.py
wilbrdt: Indeed, no different in the cli context, we can safely remove the option

wilbrdt: And set `raw_output` to `False` when calling the `read` backend method

src/ralph/utils.py
```diff
@@ -37,23 +55,60 @@ def import_string(dotted_path):
...
 
-def get_backend_instance(backend_type: BaseModel, backend_name: str, options: dict):
-    """Return the instantiated backend instance given backend-name-prefixed options."""
+def get_backend_class(backend_type: BaseModel, backend_name: str):
+    """Return the backend class given the backend type and backend name."""
+    # Get type name from backend_type class name
+    backend_type_name = backend_type.__class__.__name__[
+        : -len("BackendSettings")
+    ].lower()
+    backend_name = backend_name.lower()
```
Leobouloc, line 73: I believe this function and/or doc should be clarified in the interest of future developments. I find it pretty hard to follow.

In particular, it is confusing to have: `backend_type`, `backend_name`, and `backend_type_name` in addition to `backend_class`.

src/ralph/api/routers/health.py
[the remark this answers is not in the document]
quitterie-lcs: Forget my comment. The test failed as I forgot to add an other of your suggestion: 
```
BACKEND_CLIENT: Union[BaseLRSBackend, BaseAsyncLRSBackend]
```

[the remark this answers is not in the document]
wilbrdt: Opened issue here to add some tests: https://github.com/openfun/ralph/issues/455

src/ralph/api/routers/statements.py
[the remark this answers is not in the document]
wilbrdt: This is a good remark but coming a little late before integration of unified backends. I think we should keep the return type as sync/async generators for now, but I opened an issue to discuss it later https://github.com/openfun/ralph/issues/456

src/ralph/utils.py
wilbrdt: Opened an issue https://github.com/openfun/ralph/issues/457 so we can improve clarity in the future

src/ralph/api/routers/statements.py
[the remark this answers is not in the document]
SergioSim: I didn't mean to change the return type of the `query_statements_by_ids` method.
I think keeping the return type as `sync/async generators `is a good idea; however, we could convert the output of the generator to a list at this line using something like:
```
        if isinstance(BACKEND_CLIENT, BaseLRSBackend):
            existing_statements = list(BACKEND_CLIENT.query_statements_by_ids(...))
        else:
            existing_statements = [x async for x in BACKEND_CLIENT.query_statements_by_ids(...)]
```
Thus keeping `existing_statements` as a list. WDYT?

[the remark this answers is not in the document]
quitterie-lcs: Good idea! I change it

src/ralph/backends/conf.py
wilbrdt: Opened issue https://github.com/openfun/ralph/issues/458 as it will require a bit of work!

quitterie-lcs (approved):

PR #419, opened 2023-08-24
🔧(arnold) use var to allow override pvc name

claudusd:
## Purpose

Add the variable ralph_pvc_history_name in tray to allow to override the persistent volume name.


wilbrdt (approved):
Looks good to me, thanks for this @claudusd !

Issue #418, opened 2023-08-23
Establish an absolute max limit on the fetched statements while using the async backend


lebaudantoine:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

Using the `AsyncLRSHTTP` backend, I need to retrieve a defined number of statements.

Currently, `limit` and `chunk_size` only set the number of statements per batch. However, they lack the ability to establish an absolute upper limit on the total statements fetched.

This enhancement is particularly critical in production environments. In cases where an activity involves numerous students, resulting in the creation of millions of statements tied to a singular resource, the absence of an option to set a strict upper limit becomes apparent.

It's imperative for the client to have the capacity to define a stringent cap, thereby preventing the need to retrieve the entire dataset in one go. This not only streamlines the retrieval process but also safeguards network resources and system responsiveness.

**Describe the solution you'd like**
Support a new limit parameter in the `read` method.

**Describe alternatives you've considered**
Support a new parameter in the LRS Query.

lebaudantoine:
Presently, the terms limit and chunk_size have different naming, yet appear to share a common purpose. Is there room for enhancement?

PR #422, opened 2023-08-28
(backend) Add `max_statements` to read in LRSHTTP

Leobouloc:
## Purpose

As suggested in https://github.com/openfun/ralph/issues/418 , it would be nice to introduce a limit to the number of statements that can be returned by `AsyncLRSHTTP.read()`. This PR implements this feature (with option named `max_statements`)

## Proposal

This is implemented by simply adding a break condition when looping on the statements generator. 

- [x] implement `max_statements` for `AsyncLRSHTTP`
- [x] test `max_statements`

### DIFF
--- a/src/ralph/backends/http/async_lrs.py
+++ b/src/ralph/backends/http/async_lrs.py
@@ -10,6 +10,7 @@ from urllib.parse import ParseResult, parse_qs, urljoin, urlparse
 from httpx import AsyncClient, HTTPError, HTTPStatusError, RequestError
 from more_itertools import chunked
 from pydantic import AnyHttpUrl, BaseModel, parse_obj_as
+from pydantic.types import PositiveInt
 
 from ralph.conf import LRSHeaders, settings
 from ralph.exceptions import BackendException, BackendParameterException
@@ -105,10 +106,11 @@ class AsyncLRSHTTP(BaseHTTP):
         self,
         query: Union[str, LRSQuery] = None,
         target: str = None,
-        chunk_size: Union[None, int] = 500,
+        chunk_size: Optional[PositiveInt] = 500,
         raw_output: bool = False,
         ignore_errors: bool = False,
         greedy: bool = True,
+        max_statements: Optional[PositiveInt] = None,
     ) -> Iterator[Union[bytes, dict]]:
         """Get statements from LRS `target` endpoint.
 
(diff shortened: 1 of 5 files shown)


src/ralph/backends/http/async_lrs.py
```diff
@@ -157,15 +160,20 @@ async def read(  # pylint: disable=too-many-arguments
...
+
+            counter = 0
+            async for statement in statements_async_generator:
+                if max_statements and (counter >= max_statements):
+                    break
+                yield statement
+                counter += 1
 
         except HTTPError as error:
             msg = "Failed to fetch statements."
```
jmaupetit, line 179: We should try to reduce as much as possible the size of a try/except block to make explicit which part of the code is under surveillance.

Here is an alternative (more concise) implementation:

```python
        statements_generator = self._fetch_statements
        if greedy:
            statements_generator = self._greedy_fetch_statements

        try:
            async for counter, statement in enumerate(statements_generator(target, raw_output, query_params)):
                if max_statements and (counter >= max_statements):
                    break
                yield statement
        except HTTPError as error:
            msg = "Failed to fetch statements."
            logger.error("%s. %s", msg, error)
            raise BackendException(msg, *error.args) from error
```

Leobouloc: oups. Fixed this !

Leobouloc: I would have preferred that as well but `enumerate` is apparently not compatible with async generators. (https://groups.google.com/g/comp.lang.python/c/stdbLogcUZo?pli=1)

src/ralph/backends/http/base.py
```diff
@@ -114,19 +115,23 @@ async def read(  # pylint: disable=too-many-arguments
...
+        target: Optional[str] = None,
+        chunk_size: Optional[PositiveInt] = 500,
         ignore_errors: bool = False,
-        operation_type: Union[None, OperationType] = None,
+        operation_type: Optional[OperationType] = None,
+        simultaneous: bool = False,
+        max_num_simultaneous: Optional[int] = None,
     ) -> int:
-        """Writes statements into the HTTP server given an input endpoint."""
+        """Writes statements into the HTTP server given an input endpoint."""
```
jmaupetit: A carriage return seems to be missing here!

tests/backends/http/test_async_lrs.py
```diff
@@ -195,12 +195,80 @@ async def test_backends_http_lrs_list(caplog):
...
+                path=urlparse(more_target).path,
+                query=urlencode(params),
+                params="",
+                fragment="",
+            ).geturl(),
+            method="GET",
+            json=more_statements,
+        )
+
+    #
```
jmaupetit, line 249: Have something to say? :wink:

Leobouloc: I retract my comment

lebaudantoine (approved):

PR #426, opened 2023-08-30
Add LRS statement query parameter Pydantic model

quitterie-lcs:
## Purpose

Until now, LRS statement query parameter used in LRS backend was typed with a dictionary. However, the LRS specification defines a precise set of parameters. 

## Proposal

To query the statements resource with the LRS backend, a Pydantic model validates the input parameter according to the official LRS spec and the query parameters defined for `ralph` LRS API makes use of this model and extra parameters but without validation, as the usage is only for `ralph` and the data is trusted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to
 
 ### Changed
 
+- Refactor LRS Statements resource query parameters defined for `ralph` API
 - Helm chart: improve chart modularity
 - User credentials must now include an "agent" field which can be created
 using the cli
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- Implement Pydantic model for LRS Statements resource query parameters
 - Implement xAPI LMS Profile statements validation
 - `EdX` to `xAPI` converters for enrollment events
 
--- a/src/ralph/backends/database/base.py
+++ b/src/ralph/backends/database/base.py
@@ -4,13 +4,12 @@ import functools
 import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from datetime import datetime
 from enum import Enum, unique
-from typing import BinaryIO, List, Literal, Optional, TextIO, Union
-from uuid import UUID
+from typing import BinaryIO, List, Optional, TextIO, Union
 
 from pydantic import BaseModel
 
+from ralph.backends.http.async_lrs import LRSStatementsQuery
 from ralph.exceptions import BackendParameterException
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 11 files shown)


wilbrdt:
Not 100% sure it is feasible, but it would be nice to have a common `LRSStatements` model for both the LRS backends and the API (query parameters for `get` method). WDYT?

CHANGELOG.md
```diff
@@ -10,10 +10,12 @@ and this project adheres to
 
 ### Added
 
+- Implement Pydantic model for LRS Statements resource query parameters
 - Implement xAPI LMS Profile statements validation
 
 ### Changed
 
+- Refacto LRS Statements resource query parameters defined for `ralph` API
```
jmaupetit, line 18:
```suggestion
- Refactor LRS Statements resource query parameters defined for `ralph` API
```

src/ralph/backends/http/async_lrs.py
```diff
@@ -35,10 +40,31 @@ class StatementResponse(BaseModel):
...
+    format: Optional[Literal["ids", "exact", "canonical"]] = "exact"
+    attachments: Optional[bool] = False
+    ascending: Optional[bool] = False
+
+
 class LRSQuery(BaseQuery):
     """LRS body query model."""
 
-    query: Optional[dict]
+    query: Optional[LRSStatementsQuery]
```
jmaupetit, line 67: Do we need this level of indirection? Why not using the `LRSStatementQuery` and remove the `LRSQuery` model?

src/ralph/backends/database/base.py
```diff
@@ -57,38 +55,20 @@ class AgentParameters:
     account__home_page: Optional[str] = None
```
wilbrdt, line 55: We could take advantage of this change to create an `account` pydantic model. It could be done in another PR.
WDYT?

quitterie-lcs: Yup, in another PR indeed!

src/ralph/backends/http/async_lrs.py
```diff
@@ -141,12 +161,8 @@ async def read(  # pylint: disable=too-many-arguments
         if not target:
             target = self.statements_endpoint
 
-        query_params = {}
-        if query.query:
-            query_params.update(query.query)
-
-        # Set chunk_size to limit parameter if not exist in the query
-        query_params.setdefault("limit", chunk_size)
```
wilbrdt, line 149: We should still set the `limit` query parameter to the `chunk_size`!

We could also log a warning if the `limit` has already been set, as the user probably expect the `limit` parameter to limit the maximum number of statements when it limits the number of statements returned by each request.

src/ralph/backends/http/async_lrs.py
```diff
@@ -161,11 +177,13 @@ async def read(  # pylint: disable=too-many-arguments
         # Select the appropriate fetch function
         if greedy:
             statements_async_generator = self._greedy_fetch_statements(
-                target, raw_output, query_params
+                target, raw_output, query.dict(exclude_none=True)
```
wilbrdt, line 180:
```suggestion
                target, raw_output, query.dict(exclude_none=True, exclude_unset=True)
```

src/ralph/backends/http/async_lrs.py
```diff
@@ -161,11 +177,13 @@ async def read(  # pylint: disable=too-many-arguments
...
             )
         else:
             statements_async_generator = self._fetch_statements(
-                target=target, raw_output=raw_output, query_params=query_params
+                target=target,
+                raw_output=raw_output,
+                query_params=query.dict(exclude_none=True),
```
wilbrdt, line 186:
```suggestion
                query_params=query.dict(exclude_none=True, exclude_unset=True),
```

src/ralph/backends/http/async_lrs.py
quitterie-lcs: By default, the limit is set to 0 according to the spec. In this case, the limit parameter will always exist in the query and the case with having chunk_size ase default does not longer make sense.

wilbrdt: To the LRS specification yes! But in this client here, we are diverting the behavior of the `limit` query parameter to let the user choose the `chunk_size`.
If we don't set the `limit` to the `chunk_size`, we are not chunk sizing anymore, right?

src/ralph/backends/http/async_lrs.py
```diff
@@ -164,6 +168,12 @@ async def read(  # pylint: disable=too-many-arguments
         if query is None:
             query = LRSStatementsQuery()
 
+        query.limit = chunk_size
+        logger.warning(
+            "The limit query parameter value is overwritten by the chunk_size "
+            "parameter value."
+        )
```
wilbrdt, line 175: Nice, thanks! Would it be possible to log that warning only when `limit` inside the `query` parameter? WYDT about it?

wilbrdt (approved):
Alt text
Otherwise looks good to me!

src/ralph/backends/http/async_lrs.py
quitterie-lcs: Yup, that's better :smile:

PR #436, opened 2023-09-14
🎨(api) convert camelCase variable to snake_case

lebaudantoine:
## Purpose

Ensures PEP8 compliance without disabling PyLint.

## Proposal

Implemented `camelCase` aliases on attributes or routes' parameters.
Let me know if I am being too pedantic, and this breaking change is not worth it.


src/ralph/api/routers/statements.py
```diff
@@ -138,11 +138,13 @@ async def get(
     ###
     # Query string parameters defined by the LRS specification
     ###
-    # pylint: disable=invalid-name
-    statementId: Optional[str] = Query(None, description="Id of Statement to fetch"),
-    # pylint: disable=invalid-name, unused-argument
```
lebaudantoine, line 143: this param wasn't unused actually 🤗 did I miss smth?

wilbrdt: Probably an oversight!

quitterie-lcs: Yes it is! You have several occurences of `statementId` variable in this method. What made you notice this ?

quitterie-lcs: It's an oversight from the first implementation of the API, indeed. Then @Leobouloc added a support to return a 400 bad request if voidedStatementId is given as a parameter

wilbrdt (approved):
LGTM! :tada:

src/ralph/api/routers/statements.py
```diff
@@ -382,10 +384,9 @@ async def get(
 # pylint: disable=unused-argument
 async def put(
     current_user: Annotated[AuthenticatedUser, Depends(get_authenticated_user)],
-    # pylint: disable=invalid-name
-    statementId: UUID,
     statement: LaxStatement,
     background_tasks: BackgroundTasks,
+    statement_id: UUID = Query(..., alias="statementId"),
```
quitterie-lcs, line 389: Is the ellipsis mandatory ?

src/ralph/backends/http/async_lrs.py
```diff
@@ -49,8 +49,8 @@ class LRSStatementsQuery(BaseQuery):
 
     # pylint: disable=too-many-instance-attributes
 
-    statementId: Optional[str]  # pylint: disable=invalid-name
-    voidedStatementId: Optional[str]  # pylint: disable=invalid-name
+    statement_id: Optional[str] = Field(None, alias="statementId")
```
quitterie-lcs, line 52: Same here, is None necessary? I already used `Field()` in pydantic for xAPI models and only alias was enough

quitterie-lcs:
@lebaudantoine some camelCase remain in Pydantic model in [models](https://github.com/openfun/ralph/tree/master/src/ralph/models). To align with this PR, we could standardise all our Pydantic models, but that could be done in another PR

src/ralph/backends/http/async_lrs.py
lebaudantoine: I don't think so. https://stackoverflow.com/a/70993458

src/ralph/api/routers/statements.py
lebaudantoine: Not since this [release](https://github.com/tiangolo/fastapi/releases/tag/0.78.0), I fix it.

src/ralph/backends/http/async_lrs.py
lebaudantoine: I was inspired by the Optional Query parameters, inside the API router

lebaudantoine:
Issue opened @quitterie-lcs https://github.com/openfun/ralph/issues/439

src/ralph/backends/http/async_lrs.py
Leobouloc: Actually, migrating to `pydantic v2` requires to specify default `None` even for `Optional`. They removed it for clarity (`Optional` only indicates that a value MAY be `None`). I suggest we use this syntax even with `pydantic V1`, for clarity.

quitterie-lcs (approved):
Let's go !

PR #442, opened 2023-09-20
⚡️(api) optimized POST /statements endpoint to avoid quadratic time complexity

lebaudantoine:
## Purpose

Avoid quadratic time complexity

## Proposal

Replaced list-based `in` checks with set-based ones for improved efficiency (O(1) vs. O(n)).

- [X] Use Set instead of List
- [X] Early return when conditions are matched
- [X] Avoid unnecessary checks on statement's IDs

I removed `statements_ids` variable to maintain synchronization with `statements_dict`. Thus, I ensure the returned IDs reflect the order the new statements were stored.

Fixed a typo in a field checked by `statements_are_equivalent` function. We should consider using constants, enums or a Pydantic models instead of hard-coded literals to prevent potential errors. WDYT?

--------

Note(thx @Leobouloc): While these performance improvements optimize the code, the primary bottlenecks in processing time are the database calls, which have a more significant impact.


src/ralph/utils.py
```diff
@@ -131,13 +131,13 @@ def statements_are_equivalent(statement_1: dict, statement_2: dict):
     they may still be equivalent.
     """
     # Check that unmutable fields have the same values
-    fields = ["actor", "verb", "object", "id", "result", "context", "attachements"]
+    fields = {"actor", "verb", "object", "id", "result", "context", "attachments"}
```
lebaudantoine, line 134: Should I reflect this change in the Changelog? WDYT?

quitterie-lcs: Yes, in the `fixed` section. @wilbrdt ?

Leobouloc (approved):
Looks good to me.

src/ralph/utils.py
Leobouloc: Then you should also fix the `attachements` parameter in  `test_helpers_assert_statement_get_responses_are_equivalent`...

Leobouloc: https://github.com/openfun/ralph/blob/a42dbae8c4fd856aca6ac1cd555d381712ed80f6/tests/test_helpers.py#L50

src/ralph/utils.py
```diff
@@ -131,13 +131,13 @@ def statements_are_equivalent(statement_1: dict, statement_2: dict):
...
 
     # Check that some fields enriched by the LRS are equal when in both statements
     # The LRS specification excludes the fields below from equivalency. It was
     # decided to include them anyway as their value is inherent to the statements.
     other_fields = {"timestamp", "version"}  # "authority" and "stored" remain ignored.
-    fields.extend(other_fields & statement_1.keys() & statement_2.keys())
+    fields |= other_fields & statement_1.keys() & statement_2.keys()
```
lebaudantoine, line 140: `|=` was introduced in python 3.9, rollback this change (thx @quitterie-lcs)

src/ralph/utils.py
lebaudantoine: Fixed @Leobouloc. I wonder, do you know why these tests are under `test_helpers` and not `test_utils`?

Leobouloc: `tests/helpers.py` was introdued for functions shared across testing while `src/ralph/utils.py` is for functions shared in the Ralph library. 

The function I linked to tests `assert_statement_get_responses_are_equivalent` which is in `tests/helpers.py`, although it does indirectly test `statements_are_equivalent`.

src/ralph/api/routers/statements.py
```diff
@@ -504,7 +502,9 @@ async def post(
         _enrich_statement_with_authority(statement, current_user)
 
     try:
-        existing_statements = DATABASE_CLIENT.query_statements_by_ids(statements_ids)
+        existing_statements = DATABASE_CLIENT.query_statements_by_ids(
+            list(statements_dict.keys())
```
jmaupetit, line 506: Even if it's less explicit, the following should do the trick:

```suggestion
            list(statements_dict)
```

src/ralph/api/routers/statements.py
```diff
@@ -559,4 +557,5 @@ async def post(
 
     logger.info("Indexed %d statements with success", success_count)
 
-    return statements_ids
+    # Returns the list of IDs in the same order they were stored
+    return list(statements_dict.keys())
```
jmaupetit, line 561: ditto

src/ralph/api/routers/statements.py
```diff
@@ -481,19 +481,17 @@ async def post(
     statements_dict = {}
     for statement in map(lambda x: x.dict(exclude_unset=True), statements):
         statement_id = _enrich_statement_with_id(statement)
```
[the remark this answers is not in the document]
jmaupetit: Ping @Leobouloc :point_up_2:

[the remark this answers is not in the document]
Leobouloc: I agree. It would make things a bit clearer as well.

jmaupetit (approved):

Leobouloc (commented):
Looks correct !

PR #449, opened 2023-10-05
(api) Allow multiple authentication methods (oidc & basic) [after #463]

Leobouloc:
# WIP / Merge after `feature/enforce-scopes` (after #463 )

## Purpose

Currently, an instance of Ralph LRS has a fixed auth method (basic auth and OIDC). The suggestion is to allow multiple auth methods. The use case could be: LMS write to Ralph LRS via basic Auth; while data analysts fetch from Ralph using OIDC.

## Proposal

It is proposed to change `RALPH_RUNSERVER_AUTH_BACKEND` to `RALPH_RUNSERVER_AUTH_BACKENDS` and to make it accept comma separated lists as values `RALPH_RUNSERVER_AUTH_BACKEND=basic,oidc`.

The auth method for incoming requests will be chosen based on header information (`basic`, `oidc`)

- [ ] implement suggestion
- [ ] test that auth methods work individually and simultaneously


Leobouloc:
Replaced by https://github.com/openfun/ralph/pull/473 for convenience.

PR #464, opened 2023-10-13
✨(backends) more granular backends

SergioSim:
## Purpose

We want to have a base backend class that all other backends could inherit from.
This might be useful for backend discovery in the future plugin architecture.
Also, we want to know in advance what methods of a backend are functional before backend instantiation.

## Proposal

- [x] add BaseBackend
- [ ]  add more granular data backend interfaces or rely upon a class variable to get supported/unsupported methods (to be decided)


SergioSim:
This PR has been split into smaller PRs: #501 #510 #511 #512 #513 #515 #517 #518

PR #466, opened 2023-10-16
🐛(backends) fix default RALPH_BACKENDS__HTTP__LRS__HEADERS variable

SergioSim:
## Purpose

It seems that pydantic ignores field aliases when building the Settings objects from environment variables.

## Proposal

- [x] use the field names for the RALPH_BACKENDS__HTTP__LRS__HEADERS variable.


wilbrdt (approved):
LGTM! :tada:

quitterie-lcs (approved):

PR #470, opened 2023-10-17
✨(backends) es granular backend

SergioSim:
## Purpose

An extract from the `more granular backends` PR #464 concerning only the ESDataBackend

## Proposal

- [x] Adapt the ESDataBackend to use the template method pattern


src/ralph/backends/data/base.py
```diff
@@ -228,6 +296,47 @@ def write(  # pylint: disable=too-many-arguments
...
+
+        data = iter(data)
+        try:
+            first_record = next(data)
+        except StopIteration:
+            self.logger.info("Data Iterator is empty; skipping write to target.")
+            return 0
+        data = chain((first_record,), data)
+
+        chunk_size = chunk_size if chunk_size else self.settings.DEFAULT_CHUNK_SIZE
```
Leobouloc, line 314: not linked to this PR in particular but perhaps we could have `DEFAULT_READ_CHUNK_SIZE` & `DEFAULT_WRITE_CHUNK_SIZE` .

src/ralph/backends/data/base.py
```diff
@@ -79,24 +116,20 @@ def wrapper(*args, **kwargs):
...
-
-        Args:
-            settings (BaseDataBackendSettings or None): The data backend settings.
-                If `settings` is `None`, a default settings instance is used instead.
-        """
-
+    # This method is replaced by validate_backend_query
     def validate_query(
         self, query: Union[str, dict, BaseQuery, None] = None
     ) -> BaseQuery:
```
Leobouloc, line 135: So the intention is to remove this method ?

Leobouloc (commented):
Read up to utils.py. Looks cool, this will make things neater !

src/ralph/backends/base.py
```diff
@@ -0,0 +1,40 @@
...
+        """Pydantic Configuration."""
+
+        env_prefix = "RALPH_BACKENDS__"
+
+
+class BaseBackend(ABC):
+    """Base ralph backend class."""
+
+    settings_class = BaseBackendSettings
+    settings: settings_class
```
quitterie-lcs, line 31: I don't understand why you have to do this ? Why not defining one attribute ?

src/ralph/backends/data/base.py
quitterie-lcs: I think it is a good idea

SergioSim: Great Idea) I think we could do this change easily once all backends use this pattern)

src/ralph/backends/data/base.py
SergioSim: Yes) partially) we move it out of the class, add two new arguments to it (query_class and logger) and rename it)

src/ralph/backends/base.py
SergioSim: Well spotted) We do not have to do it) However, static-type checkers (e.g. pylance) seem to struggle to infer the type of the settings attribute(
With the generic version, this issue is not present anymore.

SergioSim:
This PR was the first attempt to shorten PR #464 by splitting it by backend.
However, for more simplicity, we opted to split the PR #464 by functionality in PRs: #501 #510 #511 #512 #513 #515 #517 #518 #519

PR #472, opened 2023-10-18
⚗️(backends) use generic base backend classes

SergioSim:
## Purpose

We experiment with the base backend class, making it generic.

## Proposal

- [x] use generic base backend classes.


SergioSim:
This PR is continued in  #501

PR #483, opened 2023-10-26
Setting up docs versioning

wilbrdt:
## Purpose

Documentation under https://openfun.github.io/ralph/ lacks a version and is always following `master`, which can be confusing for the Ralph python package users.

## Proposal

- [x] 🔧(docs) add mike tools for docs versioning
- [x] 💚(docs) add docs versioning in CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,14 @@ COMPOSE              = DOCKER_USER=$(DOCKER_USER) docker compose
 COMPOSE_RUN          = $(COMPOSE) run --rm
 COMPOSE_TEST_RUN     = $(COMPOSE_RUN)
 COMPOSE_TEST_RUN_APP = $(COMPOSE_TEST_RUN) app
-MKDOCS               = $(COMPOSE_RUN) --no-deps --publish "8000:8000" app mkdocs
+COMPOSE_RUN_DOCS     = $(COMPOSE_RUN) --no-deps --publish "8000:8000" app
+
+
+# -- Documentation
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -11,6 +11,7 @@ COMPOSE_RUN          = $(COMPOSE) run --rm
 COMPOSE_TEST_RUN     = $(COMPOSE_RUN)
 COMPOSE_TEST_RUN_APP = $(COMPOSE_TEST_RUN) app
 MKDOCS               = $(COMPOSE_RUN) --no-deps --publish "8000:8000" app mkdocs
+MIKE 				 = $(COMPOSE_RUN) --no-deps --publish "8000:8000" app mike
```
quitterie-lcs, line 14:
```suggestion
MKDOCS               = $(COMPOSE_RUN) --no-deps --publish "8000:8000" app mkdocs
MIKE                      = $(COMPOSE_RUN) --no-deps --publish "8000:8000" app mike
```

mkdocs.yml
```diff
@@ -43,3 +43,7 @@ nav:
 plugins:
     - search
     - mkdocstrings
+
+extra:
+    version:
+        provider: mike
```
quitterie-lcs, line 49: A carriage return is missing here
```suggestion
        provider: mike

```

Makefile
wilbrdt: Good catch!

quitterie-lcs (commented):
Maybe docs for version 2.X are no longer necessary ?

PR #486, opened 2023-10-27
📝(project) unify infinitive use in docstrings

quitterie-lcs:
## Purpose

In ralph, it has been established that all docstrings shall begin with an
infinitive verb. 

## Proposal

Some typos remaining are now unified.


wilbrdt (approved):
Alt text
LGTM!

PR #490, opened 2023-10-31
⚡️(cli) lazy backend options

SergioSim:
## Purpose

The `backends_options` function eagerly imports all backends to construct their options for each command that uses backend options, regardless of which CLI command is invoked (even when the cli module is just imported).
This is not very efficient.

Also, currently, CLI logging is configured after backend discovery, which does not allow controlling the logging levels used during backend discovery.

## Proposal

- [x] load backend options lazily and only compute them when they are requested
- [x] configure logging before backend discovery


wilbrdt (approved):
LGTM! :tada:

PR #491, opened 2023-10-31
✨(cli) replace archive argument with query

SergioSim:
## Purpose

The `archive` argument in the `read` cli command is not used anymore.
Instead, currently, a `--query` option is used, which only accepts JSON strings.
This complicates a bit the usage of backends that do not require a query in the form of a dictionary (e.g.: fs, s3, ldp backends).

Example:
```
ralph read -b fs -q '{"query_string": "LICENSE.md"}'
```

We want to accept a plain string as the query argument to simplify CLI usage with backends that don't need a dictionary query.

Example:
```
ralph read -b fs LICENSE.md
```

Also, this would simplify the use of the ES data backend with a Lucene query string - e.g.: `ralph read -b es id:1`

## Proposal

- [x] Replace the archive CLI argument with a query argument that accepts a JSON string or a plain string
- [x] Remove the previously defined query option, which was JSON only.


src/ralph/cli.py
```diff
@@ -638,21 +631,19 @@ def convert(from_, to_, ignore_errors, fail_on_unknown, **conversion_set_kwargs)
...
     backend,
-    archive,
     chunk_size,
     target,
     query,
     ignore_errors,
     **options,
 ):  # pylint: disable=too-many-arguments
-    """Read an archive or records from a configured backend."""
+    """Read records matching the QUERY (json or string) from a configured backend."""
```
wilbrdt, line 640: It would be the only documentation on the `read` command, which I find a bit light.
Do you know any way to document how to use the `read` methods? We could probably provide examples?

src/ralph/backends/http/async_lrs.py
```diff
@@ -157,14 +157,14 @@ async def read(  # pylint: disable=too-many-arguments
         if not target:
             target = self.settings.STATEMENTS_ENDPOINT
 
-        if query and query.limit:
+        if query.limit:
```
wilbrdt, line 160: What if `query` is not provided (at `None`)?

wilbrdt (approved):
Alt text
Just a few questions, but otherwise LGTM!

src/ralph/backends/http/async_lrs.py
SergioSim: I think the magic `@enforce_query_checks` decorator should handle this case. It calls `validate_query` before passing the `query` to this method) We intend to change that with the template pattern in the future)

wilbrdt: Oh true, missed it again!

src/ralph/cli.py
SergioSim: That's a great idea! Thanks) Indeed, having dedicated help pages for each backend would be helpful)

I think we could use sub-commands with click to achieve something like:
```
ralph read es --help # getting read docs for es backend
ralph read es --option1 --option2 QUERY
```
Will give it a try in a separate PR)

src/ralph/backends/data/es.py
```diff
@@ -111,6 +111,18 @@ class ESQuery(BaseQuery):
...
+class ESQuery(BaseESQuery):
+    """Elasticsearch query model.
+
+    Attributes:
+        query_string (str): Same as in `BaseESQuery`, however also accepts a JSON
+            instance of `BaseESQuery` which, if provided, is used instead.
+    """
+
+    # pylint: disable=unsubscriptable-object
+    query_string: Union[Json[BaseESQuery], str, None]
```
SergioSim, line 123: `Json[BaseESQuery], str` causes to silently fall back using a Lucene query if the query is not in JSON format.
We probably want to add a log message when this happens.

src/ralph/backends/data/es.py
```diff
@@ -197,20 +192,20 @@ def list(
...
         self,
         *,
-        query: Optional[Union[str, ESQuery]] = None,
+        query: Optional[ESQuery] = None,
         target: Optional[str] = None,
         chunk_size: Optional[int] = None,
         raw_output: bool = False,
         ignore_errors: bool = False,
     ) -> Iterator[Union[bytes, dict]]:
+        # pylint: disable=too-many-arguments,too-many-branches
```
wilbrdt, line 204:
```suggestion
```

src/ralph/backends/data/es.py
```diff
@@ -197,20 +192,20 @@ def list(
...
         """Read documents matching the query in the target index and yield them.
 
         Args:
-            query (str or ESQuery): A query in the Lucene query string syntax or a
+            query (ESQuery): A query in the Lucene query string syntax or a
```
wilbrdt, line 208: So from now on, we can't pass a string in the Lucene query syntax, right?
Fine with it but we should change the docstrings

src/ralph/backends/data/clickhouse.py
```diff
@@ -91,8 +89,17 @@ class Config(BaseSettingsConfig):
...
+class ClickHouseQuery(BaseModel):
+    """Base ClickHouse query model.
+
+    Attributes:
+        select:
+        where:
+        parameters:
+        limit:
+        sort:
+        column_oriented:
```
wilbrdt, line 101:
```suggestion
    Attributes:
        select (str or list): Name of the table(s) to query.
        where (str or list): Expression for filtering the data.
        parameters (dict): Dictionary of substitution values.
        limit (int): Maximum number of rows returned.
        sort (str): Expression determining the sorting direction.
        column_oriented (bool): Whether to return the results as a sequence of columns rather than a sequence of rows.
```

tests/backends/data/test_async_es.py
```diff
@@ -436,37 +436,26 @@ async def test_backends_data_async_es_read_with_query(es, async_es_backend, capl
...
+
     # Find every odd item.
-    query = {"query": {"term": {"modulo": 1}}}
+    query = ESQuery(query={"term": {"modulo": 1}})
     results = [statement async for statement in backend.read(query=query)]
     assert len(results) == 2
     assert results[0]["_source"]["id"] == 1
     assert results[1]["_source"]["id"] == 3
 
-    # Find documents with ID equal to one or five.
```
wilbrdt, line 446: I don't mind dropping the ability to use the Lucene syntax

SergioSim:
This PR has been continued in #527

Issue #482, opened 2023-10-25
Remove JSON column from ClickHouse backend


bmtcril:
## Feature Request

When I implemented the ClickHouse backend I used the experimental JSON column type based on feedback from ClickHouse that it would be promoted out of experimental status "soon". We discovered that it had some serious missing functionality and added in the String column version of the JSON data next to it as a workaround. I recently reconnected with ClickHouse on the topic and it seems like they are not prioritizing getting the JSON field fully functional any time soon, so I would like to remove it!

**Describe the solution you'd like**

I'd like to remove the JSON column and change the existing handful of places in the code that use it to use the String column with JSON string function calls. They are faster, fully functional, and it will reduce the storage for the xAPI statements significantly. We entirely use the String column in Aspects.

**Describe alternatives you've considered**

Waiting for CH to finish their feature and dropping the String column instead. 😄 Last I heard from them "no one is working on that".

**Discovery, Documentation, Adoption, Migration Strategy**

This would be a breaking change for any downstream consumers of the Ralph database directly, but since I think everyone using this would already have all of their data in the String column I think the only necessary migration internal to Ralph would be to drop the column. Since Ralph doesn't even manage the schema I hope that would be a small issue.

**Do you want to work on it through a Pull Request?**

I'd be happy to do this work, I just want to make sure it would be a welcome change and answer any questions you may have before doing the work. I also know there was work happening on the abstracting the backends that may need to land first.

wilbrdt:
Hi @bmtcril ,
As you mentioned, I don't think of any barrier for this change and it would be welcomed! Happy to assist you with this task, and we can aim for a release in version 4.0.

PR #493, opened 2023-11-02
♻️(backends) remove ClickHouse JSON column (#482)

bmtcril:
## Purpose

Initial implementation of the ClickHouse backend used the experimental JSON column type, with the string of the JSON as a backup. The JSON type has never graduated out of experimental status and has some serious issues that make it undesirable to support.

This commit removes all references to the column in favor of using the JSON ClickHouse functions, which are generally faster and better supported.


docs/backends.md
```diff
@@ -229,4 +228,4 @@ LRS parameters required to connect are:
 
 Optional parameters can be configured if necessary: 
 
-- `headers`: a comma-separated key=value list of LRS server headers
\ No newline at end of file
+- `headers`: a comma-separated key=value list of LRS server headers
```
wilbrdt, line 231: Not sure what the change is here :grin:

src/ralph/backends/data/clickhouse.py
```diff
@@ -235,6 +233,7 @@ def read(
             chunk_size = self.default_chunk_size
 
         query = (
+            # pylint: disable=too-many-function-args
```
wilbrdt, line 236:
```suggestion
```
Not sure to see the purpose of disabling this warning here (I can't reproduce it)

wilbrdt (approved):
Some nitpicking here and there but otherwise looks good to me! :tada: 
Thank you for the contribution! :raised_hands:

docs/backends.md
bmtcril: It's just a newline at the end that my editor puts in automatically since it's the last line in the file and didn't have one.

src/ralph/backends/data/clickhouse.py
bmtcril: Interesting, I couldn't get it to pass pylint locally without it. I'll remove it and see what happens in CI.

bmtcril: Looks like it works without it! Maybe because I was running Python 3.8 locally and CI is at 3.11. 🤷

bmtcril:
Thanks for the feedback @wilbrdt , it should all be addressed!

wilbrdt:
Thanks again @bmtcril !

PR #497, opened 2023-11-08
🐛(Makefile) add missing backslash

Leobouloc:
Missing backslash in Makefile leads to `-w` (write to disk) being ignored
when creating `auth.json`. This PR fixes this.

- [x] add missing `\`
- [x] update CHANGELOG

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ bin/init-cluster:
 		-u $(RALPH_LRS_AUTH_USER_NAME) \
 		-p $(RALPH_LRS_AUTH_USER_PWD) \
 		-s $(RALPH_LRS_AUTH_USER_SCOPE) \
-		-M $(RALPH_LRS_AUTH_USER_AGENT_MBOX)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ have an authority field matching that of the user
 
 - API: Fix a typo ('attachements' -> 'attachments') to ensure compliance with
     the LRS specification and prevent potential silent bugs.
+- Makefile: Fix bug in creating `.ralph/auth.json`
```
jmaupetit, line 61: There is no need to add a CHANGELOG entry for this as it's relevant from a developer-perspective, not users.

jmaupetit (commented):
Thank you for this! Two remarks:

1. your commit title scope should be `project` instead of `Makefile`
2. the commit message should describe the issue: the missing backslash leads to ignore the `-w` option that persists the authentication file to disk!

wilbrdt (approved):
Alt text

PR #500, opened 2023-11-09
🚚(backends) rename query_model to query_class

SergioSim:
## Purpose

Data backends have two class attributes referring to pydantic model types that follow different naming conventions:
`settings_class` and `query_model`. 

## Proposal

For consistency, we propose to follow one naming convention.

- [x] rename `query_model` to `query_class`


wilbrdt (approved):
Good call! :tada:

quitterie-lcs (approved):

Leobouloc (approved):
nice :)

PR #501, opened 2023-11-09
✨(backends) generic data backends

SergioSim:
## Purpose

When data backends inherit from `BaseDataBackend`, they alter (specialize) the `__init__` and `read` method signatures with their proper `settings_class` and `query_class` types.
This is intended; however, this leaves a lot of freedom for data backend implementations - (for example, setting the type of `settings_class` to an `int`)

## Proposal

We propose to improve data backend typing using generic types.

- [x] use generic base backend classes.

### DIFF
--- a/src/ralph/backends/data/async_mongo.py
+++ b/src/ralph/backends/data/async_mongo.py
@@ -4,7 +4,7 @@ import json
 import logging
 from io import IOBase
 from itertools import chain
-from typing import Any, Dict, Iterable, Iterator, Optional, Union
+from typing import Any, Dict, Iterable, Iterator, Optional, TypeVar, Union
 
 from bson.errors import BSONError
 from motor.motor_asyncio import AsyncIOMotorClient
@@ -29,22 +29,25 @@ from ..data.base import (
 )
 
 logger = logging.getLogger(__name__)
+Settings = TypeVar("Settings", bound=MongoDataBackendSettings)
 
 
-class AsyncMongoDataBackend(BaseAsyncDataBackend, AsyncWritable, AsyncListable):
+class AsyncMongoDataBackend(
+    BaseAsyncDataBackend[Settings, MongoQuery],
+    AsyncWritable,
+    AsyncListable,
+):
     """Async MongoDB data backend."""
 
     name = "async_mongo"
-    query_class = MongoQuery
-    settings_class = MongoDataBackendSettings
 
-    def __init__(self, settings: Optional[MongoDataBackendSettings] = None):
+    def __init__(self, settings: Optional[Settings] = None):
         """Instantiate the asynchronous MongoDB client.
 
         Args:
             settings (MongoDataBackendSettings or None): The data backend settings.
         """
-        self.settings = settings if settings else self.settings_class()
+        super().__init__(settings)
         self.client = AsyncIOMotorClient(
             self.settings.CONNECTION_URI, **self.settings.CLIENT_OPTIONS.dict()
         )
--- a/src/ralph/backends/data/base.py
+++ b/src/ralph/backends/data/base.py
@@ -5,7 +5,7 @@ import logging
 from abc import ABC, abstractmethod
 from enum import Enum, unique
 from io import IOBase
-from typing import Iterable, Iterator, Optional, Union
+from typing import Any, Generic, Iterable, Iterator, Optional, Type, TypeVar, Union
 
 from pydantic import BaseModel, BaseSettings, ValidationError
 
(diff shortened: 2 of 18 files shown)


src/ralph/backends/data/base.py
```diff
@@ -144,25 +144,65 @@ def list(
...
+        return None
+
+    if isinstance(argument, TypeVar):
+        return argument.__bound__
+
+    return argument
+
+
+def set_backend_settings_class(backend_class: Type):
+    """Set `settings_class` attribute with `Config.env_prefix` for `backend_class`."""
```
SergioSim, line 167: Currently, we don't alter the `backend_class`. However, we could automatize setting the `Config.env_prefix` here with something similar to:
```python
settings_class.Config.env_prefix = f"RALPH_BACKENDS__DATA__{backend_class.name.upper()}__"
```
or (to avoid altering the class):
```python
class Config(BaseSettingsConfig):
    """Pydantic Configuration."""

    env_prefix = f"RALPH_BACKENDS__DATA__{backend_class.name.upper()}__"
    env_file = ".env"
    env_file_encoding = core_settings.LOCALE_ENCODING

backend_class.settings_class = type(
    settings_class.__name__, (settings_class,), {"Config": Config}
)
```
However, then, we will have separate environment variables for configuring sync and async data backends. Ex:
```
RALPH_BACKENDS__DATA__ES__DEFAULT_INDEX=statements
RALPH_BACKENDS__DATA__ASYNC_ES__DEFAULT_INDEX=statements
```
WDYT?

src/ralph/backends/data/async_mongo.py
```diff
@@ -29,22 +29,25 @@
 )
 
 logger = logging.getLogger(__name__)
+Settings = TypeVar("Settings", bound=MongoDataBackendSettings)
 
 
-class AsyncMongoDataBackend(BaseAsyncDataBackend, AsyncWritable, AsyncListable):
+class AsyncMongoDataBackend(
+    BaseAsyncDataBackend[Settings, MongoQuery],
```
SergioSim, line 36: It might be easier to inherit directly from
```python
BaseAsyncDataBackend[MongoDataBackendSettings, MongoQuery]
```
However, then there is a type issue with `AsyncMongoLRSBackend`  that extends `AsyncMongoDataBackend` and, at the same time, changes its settings to `MongoLRSBackendSettings`.
Thus, using a `Settings` TypeVar here allows future specialization of this class.
Should we apply this pattern to other backends even when we don't specialize them further? And what about allowing the same thing for the `Query` type?
WDYT?

wilbrdt: I'm not a fan of having to define a `Settings` TypeVar for each backend.
If `lrs` backends would not have proper settings (e.g. they would only inherit from their `data` backend's settings), we wouldn't have to allow future specialization, right?
Only the `fs` backend has a proper `LRS` setting. Is it still relevant for `lrs` backends to have their proper settings class?

wilbrdt (commented):
Alt text
Don't have any remarks but we should discuss about postponing the specialization of settings!

src/ralph/backends/data/async_mongo.py
SergioSim: You are right) If `lrs` backends don't redefine their `settings` we could be less flexible and re-use their `data`  settings equivalent. However, currently, all `lrs` backends redefine their settings to update the `env_prefix` and besides the `fs` backend, the `clickhouse` backend too has one additional lrs-specific setting: `IDS_CHUNK_SIZE`.

src/ralph/backends/data/base.py
```diff
@@ -144,25 +144,65 @@ def list(
         """
 
 
-class BaseDataBackend(ABC):
+def get_backend_generic_argument(backend_class: Type, position: int) -> Optional[Type]:
+    """Return the generic argument of `backend_class` at specified `position`."""
+    if not hasattr(backend_class, "__orig_bases__"):
+        return None
+
+    bases = backend_class.__orig_bases__[0]
```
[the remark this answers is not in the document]
SergioSim: Good question) Getting `__orig_bases__[0]` seems to work with the latest Python 3.7 version; however, I'm not sure whether this might work with older versions in the 3.7 line( I think it's not documented(

Also, not all backends seem compatible with python3.7 - ex.: ES & LDP backends import `Litteral`, Swift backend uses `@cached_property`, etc.

I would vote for dropping python3.7 support for backends.
Currently, we don't test the backends with Python 3.7; thus, we cannot guarantee that they still work, right?

(Also, then we could replace `__orig_bases__[0]` with the safer `typing.get_args(tp)`)
WDYT?

Leobouloc (approved):
This looks good to me. An extra review might be nice as this PR has a fair amount of abstraction !

Issue #401, opened 2023-07-27
⚡️ Improve FastAPI performance by +265% 🚀🚀


lebaudantoine:
These performance improvement tips are from Marcelo Trylesinski, the maintainer of FastAPI. These insights are shared in a LinkedIn post, which you can find [here](https://www.linkedin.com/posts/hendorf_fastapi-uvicorn-starlette-ugcPost-7088104851280146432-qdfA/?utm_source=share&utm_medium=member_ios)

10 tips for performances improvement : 

> 1. use [uvloop](https://github.com/MagicStack/uvloop) +10%🚀, is a fast, drop-in replacement of the built-in asyncio event loop
> 2. use [httptools](https://github.com/MagicStack/httptools) - faster parsing +10%🚀, Python binding for nodejs HTTP parser, you just need to have it installed in your environment
> 3. use bigger threadpool +5%🚀
> 4. use simple async application, +15%🚀, no overhead of threads
> 5. remove duplicate validation, have only FastAPI do the validation +25%🚀
> 6. use ORJSON + 5% 🚀(if not using the latest Pydanctic version already)
> 7. consider running without validation is an option, 150%, without logging +25%. Both not a general recommendation!
> 8. use the most recent version of FastAPI, + 265% 🚀- this is related to pydantic going rust
> 9. use ASGI Middleware: see https://lnkd.in/eUW5-bwG
> 10. compress responses

Stay tuned for the recording of the conference, which will be released on the [EuroPython channel](https://youtube.com/@EuroPythonConference) in a few weeks, allowing us to dive deeper into these performance-boosting techniques. Happy optimizing!

lebaudantoine:
According to the [uvicorn](https://www.uvicorn.org/) documentation, the framework installs "Cython-based" dependencies like `uvloop` and `httptools` **"if possible"**.

Using `uvicorn`, should automatically install and utilize the `uvloop` event loop, same for the HTTP protocol.

While it's safe to assume that using `uvicorn` would generally address points 1 and 2, let's explore the actual implementation to understand the specific conditions that determine whether these optimizations can be applied, and if there are **indeed**  used by Ralph.

PR #504, opened 2023-11-10
Migrate to pydantic v2

Leobouloc:
## Purpose

As mentionned in https://github.com/openfun/ralph/issues/401 , upgrading to pydantic 2 should greatly improve performances. Also it would allow projects depending on `ralph` (such as `warren`) to upgrade more easily. 

### ERROR: Extra not permitted

In `conf.py`, errors such as: 
```
    super().__init__(
E   pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
E   LOCALE_ENCODING
E     Extra inputs are not permitted [type=extra_forbidden, input_value='utf-8', input_type=str]
E       For further information visit https://errors.pydantic.dev/2.1/v/extra_forbidden
```

may be due to a change in default behavior in models `extra=ignore` to `extra=failure`. We may set this parameter manually to fix this issue.

### ERROR: Unable to generate pydantic-core schema 
```
E   pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'ralph.models.xapi.base.common.IRI'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
E
E   If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
E
```
To fix this we change the way we define classes.

**Previously in Ralph**
```
class CommaSeparatedTuple(BaseModel):
    @model_validator(mode='before')
    @classmethod
    def validate(cls, value: Union[str, Tuple[str]]) -> Tuple[str]:
        """Checks whether the value is a comma separated string or a tuple."""

        if isinstance(value, tuple):
            return value

        if isinstance(value, str):
            return tuple(value.split(","))

        raise TypeError("Invalid comma separated list")
```
**New in Ralph** 
```
def validate_comma_separated_tuple(value: Union[str, Tuple[str, ...]]) -> Tuple[str]:
    """Checks whether the value is a comma separated string or a tuple."""

    if isinstance(value, tuple):
        return value

    if isinstance(value, str):
        return tuple(value.split(","))

    raise TypeError("Invalid comma separated list")

CommaSeparatedTuple = Annotated[Union[str, Tuple[str, ...]], AfterValidator(validate_comma_separated_tuple)]
```

### ERROR: Missing values

A recurrent error is the one below, which should be linked to the change of behavior when using `Optional[...]` in Pydantic V2. Previously, it would assign a default value of `None`, which is no longer the case. This should have been dealt with by `bump-pydantic` but was not always the case. It would be nice to know why before confirming this solution.
```
src/ralph/models/xapi/video/statements.py:85: in VideoPlayed
    verb: PlayedVerb = PlayedVerb()
E   pydantic_core._pydantic_core.ValidationError: 1 validation error for PlayedVerb
E   display
E     Field required [type=missing, input_value={}, input_type=dict]
E       For further information visit https://errors.pydantic.dev/2.1/v/missing
```

**Previously in Ralph**
```
class VideoActivity(BaseXapiActivity):
    name: Optional[Dict[Literal[LANG_EN_US_DISPLAY], str]]
```
**New in Ralph** 
```
class VideoActivity(BaseXapiActivity):
    name: Optional[Dict[Literal[LANG_EN_US_DISPLAY], str]] = None
```


src/ralph/models/edx/problem_interaction/fields/events.py
```diff
@@ -41,13 +42,13 @@ class CorrectMap(BaseModelWithConfig):
         queuestate (json): see QueueStateField.
     """
 
-    answervariable: Union[Literal[None], None, str]
+    answervariable: Union[Literal[None], None, str] = None
```
Leobouloc, line 45: "I wonder if setting a default value for this field makes it not required" Originally commented by @quitterie-lcs

wilbrdt:
Closing as it is done in #550

PR #507, opened 2023-11-10
🔧(project) add flake8-fixme set of rules for Ruff

wilbrdt:
## Purpose

Previously, pylint was checking if any `FIXME` or `TODO` was present in the
codebase.

## Proposal

Adding `FIX` (`flake8-fixme`) set of rules to Ruff to check these.


SergioSim (approved):
:+1: 

image

PR #510, opened 2023-11-13
♻️(backends) use common utilities among backends

SergioSim:
## Purpose

During data backend unification work, some shared utilities were developed to factorize duplication:
`parse_dict_to_bytes`, `read_raw` and `iter_by_batch`.
However, some backends still used their own implementation of these utilities, leading to some minor behavioral differences among backends.

## Proposal

- [x] update backends to use common utilities.
- [x] make `parse_bytes_to_dict` more generic (`parse_to_dict`) to improve re-usability (see usage in ClickHouseDataBackend).

~**Note:** This PR depends on #501~

### DIFF
--- a/src/ralph/backends/data/async_es.py
+++ b/src/ralph/backends/data/async_es.py
@@ -18,7 +18,7 @@ from ralph.backends.data.base import (
 )
 from ralph.backends.data.es import ESDataBackend, ESDataBackendSettings, ESQuery
 from ralph.exceptions import BackendException, BackendParameterException
-from ralph.utils import parse_bytes_to_dict, read_raw
+from ralph.utils import async_parse_dict_to_bytes, parse_iterable_to_dict
 
 logger = logging.getLogger(__name__)
 Settings = TypeVar("Settings", bound=ESDataBackendSettings)
@@ -131,7 +131,8 @@ class AsyncESDataBackend(
             chunk_size (int or None): The chunk size when reading documents by batches.
                 If chunk_size is `None` it defaults to `DEFAULT_CHUNK_SIZE`.
             raw_output (bool): Controls whether to yield dictionaries or bytes.
-            ignore_errors (bool): Ignored.
+            ignore_errors (bool): No impact as encoding errors are not expected in
+                Elasticsearch results.
 
         Yield:
             bytes: The next raw document if `raw_output` is True.
@@ -140,10 +141,23 @@ class AsyncESDataBackend(
         Raise:
             BackendException: If a failure occurs during Elasticsearch connection.
         """
+        if raw_output:
+            documents = self.read(
+                query=query,
+                target=target,
+                chunk_size=chunk_size,
+                raw_output=False,
+                ignore_errors=ignore_errors,
+            )
+            async for document in async_parse_dict_to_bytes(
+                documents, self.settings.LOCALE_ENCODING, ignore_errors, logger
+            ):
+                yield document
+
+            return
+
         target = target if target else self.settings.DEFAULT_INDEX
         chunk_size = chunk_size if chunk_size else self.settings.DEFAULT_CHUNK_SIZE
-        if ignore_errors:
-            logger.warning("The `ignore_errors` argument is ignored")
 
         if not query.pit.keep_alive:
             query.pit.keep_alive = self.settings.POINT_IN_TIME_KEEP_ALIVE
@@ -183,10 +197,6 @@ class AsyncESDataBackend(
             if count:
                 query.search_after = [str(part) for part in documents[-1]["sort"]]
             kwargs["search_after"] = query.search_after
-            if raw_output:
-                documents = read_raw(
-                    documents, self.settings.LOCALE_ENCODING, ignore_errors, logger
-                )
             for document in documents:
                 yield document
 
@@ -206,9 +216,9 @@ class AsyncESDataBackend(
                 If target is `None`, the `DEFAULT_INDEX` is used instead.
             chunk_size (int or None): The number of documents to write in one batch.
                 If chunk_size is `None` it defaults to `DEFAULT_CHUNK_SIZE`.
-            ignore_errors (bool): If `True`, errors during the write operation
-                will be ignored and logged. If `False` (default), a `BackendException`
-                will be raised if an error occurs.
+            ignore_errors (bool): If `True`, errors during decoding, encoding and
+                sending batches of documents are ignored and logged.
+                If `False` (default), a `BackendException` is raised on any error.
             operation_type (BaseOperationType or None): The mode of the write operation.
                 If `operation_type` is `None`, the `default_operation_type` is used
                 instead. See `BaseOperationType`.
@@ -217,8 +227,8 @@ class AsyncESDataBackend(
             int: The number of documents written.
 
         Raise:
-            BackendException: If a failure occurs while writing to Elasticsearch or
-                during document decoding and `ignore_errors` is set to `False`.
+            BackendException: If any failure occurs during the write operation or
+                if an inescapable failure occurs and `ignore_errors` is set to `True`.
             BackendParameterException: If the `operation_type` is `APPEND` as it is not
                 supported.
         """
@@ -240,7 +250,7 @@ class AsyncESDataBackend(
 
         data = chain((first_record,), data)
         if isinstance(first_record, bytes):
-            data = parse_bytes_to_dict(data, ignore_errors, logger)
+            data = parse_iterable_to_dict(data, ignore_errors, logger)
 
         logger.debug(
             "Start writing to the %s index (chunk size: %d)", target, chunk_size
--- a/src/ralph/backends/data/async_mongo.py
+++ b/src/ralph/backends/data/async_mongo.py
@@ -1,10 +1,9 @@
 """Async MongoDB data backend for Ralph."""
 
-import json
 import logging
 from io import IOBase
 from itertools import chain
-from typing import Any, Dict, Iterable, Iterator, Optional, TypeVar, Union
+from typing import Iterable, Iterator, Optional, TypeVar, Union
 
 from bson.errors import BSONError
 from motor.motor_asyncio import AsyncIOMotorClient
@@ -18,7 +17,7 @@ from ralph.backends.data.mongo import (
     MongoQuery,
 )
 from ralph.exceptions import BackendException, BackendParameterException
-from ralph.utils import parse_bytes_to_dict
+from ralph.utils import async_parse_dict_to_bytes, iter_by_batch, parse_iterable_to_dict
 
 from ..data.base import (
     AsyncListable,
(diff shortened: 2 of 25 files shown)


src/ralph/backends/data/es.py
```diff
@@ -219,7 +219,9 @@ def read(  # noqa: PLR0912, PLR0913
             chunk_size (int or None): The chunk size when reading documents by batches.
                 If chunk_size is `None` it defaults to `DEFAULT_CHUNK_SIZE`.
             raw_output (bool): Controls whether to yield dictionaries or bytes.
-            ignore_errors (bool): Ignored.
+            ignore_errors (bool): If `True`, errors during the encoding operation
+                will be ignored and logged. If `False` (default), a `BackendException`
+                will be raised if an error occurs.
```
Leobouloc, line 224: unless I am mistaken, this behavior has not been changed and `ignore_errors`. Or did I miss smt ? 

at least, the warning message is still there ```"The `ignore_errors` argument is ignored"```

src/ralph/backends/data/fs.py
```diff
@@ -224,12 +208,34 @@ def read(  # noqa: PLR0913
                 }
             )
 
+    def _iter_files_matching_query(
+        self, target: Union[str, None], query: BaseQuery
```
Leobouloc, line 212: perhaps arguments could be explicited in the docstring as "target" doesn't feel self explanatory.

src/ralph/backends/data/fs.py
```diff
@@ -224,12 +208,34 @@ def read(  # noqa: PLR0913
...
+            query.query_string = self.default_query_string
+
+        path = Path(target) if target else self.default_directory
+        if not path.is_absolute() and path != self.default_directory:
+            path = self.default_directory / path
+
+        paths = list(filter(lambda x: x.is_file(), path.glob(query.query_string)))
+        if not paths:
+            msg = "No file found for query: %s"
+            logger.info(msg, path / Path(str(query.query_string)))
```
Leobouloc, line 225: should this be debug ?

src/ralph/backends/data/fs.py
```diff
@@ -184,30 +183,15 @@ def read(  # noqa: PLR0913
...
-
-        logger.debug("Reading matching files: %s", paths)
-
-        for path in paths:
-            with path.open("rb") as file:
-                reader = self._read_raw if raw_output else self._read_dict
-                for chunk in reader(file, chunk_size, ignore_errors):
+        for file, path in self._iter_files_matching_query(target, query):
+            if raw_output:
+                while chunk := file.read(chunk_size):
```
Leobouloc, line 191: If I understand correctly, the chunk size here is in bytes and not in number of documents as it is the case with other backends. Is this correct ? If yes, shouldn't we change this ?

Either case, I think the behavior of `chunk_size` (bytes or documents) should be explicited in the docstring.

Leobouloc (commented):
This cleanup is very welcome !

Reviewed up to mongo backend.

Two of my comments don't apply to the code you changed but could be within the scope of this PR.

src/ralph/backends/data/swift.py
```diff
@@ -330,10 +334,9 @@ def write(  # noqa: PLR0912, PLR0913
                 msg = "Failed to write to object %s: %s"
                 error = err.msg
                 logger.error(msg, target_object, error)
-                if not ignore_errors:
-                    raise BackendException(msg % (target_object, error)) from err
+                raise BackendException(msg % (target_object, error)) from err
 
-        count = sum(1 for _ in data)
+        count = counter["count"]
```
Leobouloc, line 339: My understanding is that `counter` will reflect the number times `data` generator is read in the above line (`self.connection.put_object(...)`), in the `try` block. But if there is an error during "PUT", the value of `count` will be off by 1 (since the data was already read).

More hypothetic, if reading is done by batch through some internal logic and there is an error, then the value of count could be off by an unknown value.

Opened this issue for testing: https://github.com/openfun/ralph/issues/522

tests/backends/data/test_fs.py
```diff
@@ -972,11 +984,15 @@ def test_backends_data_fs_write_without_target(fs_backend, monkeypatch):
     ]
 
 
-def test_backends_data_fs_close(fs_backend):
-    """Test that the `FSDataBackend.close` method raise an error."""
+def test_backends_data_fs_close(fs_backend, caplog):
+    """Test that the `FSDataBackend.close` method write an info log."""
```
Leobouloc, line 988:
```suggestion
    """Test that the `FSDataBackend.close` method produces an info level log."""
```

tests/backends/data/test_ldp.py
```diff
@@ -679,7 +679,7 @@ def mock_post(url):
 
 
 def test_backends_data_ldp_close(ldp_backend, caplog):
-    """Test that the `LDPDataBackend.close` method raise an error."""
+    """Test that the `LDPDataBackend.close` method write an info log."""
```
Leobouloc, line 682:
```suggestion
    """Test that the `LDPDataBackend.close` method writes an info log."""
```
(or produces)

tests/backends/data/test_s3.py
```diff
@@ -603,18 +613,14 @@ def test_backends_data_s3_write_with_create_index_operation(
...
         )
     ) == (
         [
-            (
-                "ralph.backends.data.s3",
-                logging.ERROR,
-                f"Failed to encode JSON: {error}, for document {data[0]}",
-            )
+            ("ralph.backends.data.s3", logging.ERROR, msg),
+            ("ralph.backends.data.s3", logging.WARNING, msg),
```
Leobouloc, line 622: Should we log errors while  `ignore_errors == True` as `WARNING` ?

I don't have a strong opinion, but opening this in case someone does :)

Leobouloc (commented):
This tidying is much appreciated !

src/ralph/backends/data/async_mongo.py
```diff
@@ -18,7 +17,7 @@
     MongoQuery,
 )
 from ralph.exceptions import BackendException, BackendParameterException
-from ralph.utils import parse_bytes_to_dict
+from ralph.utils import async_parse_dict_to_bytes, iter_by_batch, parse_to_dict
```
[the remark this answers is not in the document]
SergioSim: Good Idea) Thanks) renamed it)

src/ralph/backends/data/clickhouse.py
```diff
@@ -284,16 +295,7 @@ def read(  # noqa: PLR0912, PLR0913
...
-                try:
-                    yield reader(statement)
-                except (TypeError, ValueError) as error:
-                    msg = "Failed to encode document %s: %s"
-                    if ignore_errors:
-                        logger.warning(msg, statement, error)
-                        continue
-                    logger.error(msg, statement, error)
-                    raise BackendException(msg % (statement, error)) from error
+            yield from parse_to_dict(result, ignore_errors, logger, self._read_json)
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks) Renamed the method to `_parse_event_json` and updated the docstring)

src/ralph/backends/data/fs.py
SergioSim: Good point) Thanks) Added a mention about the "target" here) I keep it brief to avoid repeating the full explication from the `read` method.

src/ralph/backends/data/fs.py
SergioSim: Good question) As the read operation stops after this line, I think it might be useful for the caller to know why no results are yielded (without starting a debug session)

src/ralph/backends/data/swift.py
SergioSim: Nice catch) Thanks) Yes, you are right; we indeed should increment the counter after yielding the statement (if we start to count with 0). Good idea to open an issue for testing this further. (Updated the counter)

tests/backends/data/test_s3.py
SergioSim: Good point) I think we could indeed decrease it) Don't have a strong opinion on it too)

tests/backends/data/test_fs.py
SergioSim: Thanks)

tests/backends/data/test_ldp.py
SergioSim: Thanks) used produced)

src/ralph/backends/data/es.py
SergioSim: Nice catch) Thanks) Have removed the `warning` message) You are right; the behavior has not changed, however, we do handle the `ignore_errors` argument, (even if it has no impact on the result) - updated the docstring to:
```
No impact as encoding errors are not expected in Elasticsearch results.
```

src/ralph/backends/data/swift.py
```diff
@@ -221,13 +221,13 @@ def read(  # noqa: PLR0913
             msg = "Failed to read %s: %s"
             error = err.msg
             logger.error(msg, query.query_string, error)
-            if not ignore_errors:
-                raise BackendException(msg % (query.query_string, error)) from err
-
-        reader = self._read_raw if raw_output else self._read_dict
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks) Yes, indeed) updated the docstring to reflect the change) I propose raising an exception here regardless of the `ignore_errors` argument, as the read operation can not proceed after a `ClientException`.

src/ralph/backends/data/fs.py
SergioSim: Good Idea) Thanks) added more docstring explication in PR #519 (dedicated to the `chunk_size` argument)
```
chunk_size (int or None): The chunk size when reading files.
                If `chunk_size` is `None` it defaults to `READ_CHUNK_SIZE`.
                If `raw_output` is set to `False`, files are read line by line.
```

src/ralph/utils.py
```diff
@@ -171,39 +186,80 @@ def statements_are_equivalent(statement_1: dict, statement_2: dict) -> bool:
...
+def iter_by_batch(data: Iterable[Any], batch_size: int):
+    """Iterate over `data` Iterable and yield batches of size `batch_size`."""
+    batch = []
+    for document in data:
+        batch.append(document)
+        if len(batch) >= batch_size:
+            yield batch
+            batch = []
+    if batch:
+        yield batch
```
wilbrdt, line 262: Do we have to yield a list? Else we could use the `itertools` implementation that will be added in Python 3.12 (https://docs.python.org/3/library/itertools.html#itertools.batched)
```suggestion
def batched(iterable: Iterable, n: int):
    """
    Batch data from an iterable into tuples of length n. The last batch may be
    shorter than n.

    Taken from itertools.batched introduced in python 3.12
    """
    if n < 1:
        raise ValueError("n must be at least one")
    it = iter(iterable)
    while batch := tuple(islice(it, n)):
        yield batch
```
WDYT?

tests/backends/data/test_s3.py
wilbrdt: I don't have one either!

wilbrdt (approved):
Alt text
Have some nitpicking suggestions, otherwise looks good to me!

src/ralph/backends/data/async_es.py
```diff
@@ -140,10 +141,23 @@ async def read(  # noqa: PLR0912, PLR0913
...
+                raw_output=False,
+                ignore_errors=ignore_errors,
+            )
+            async for document in async_parse_dict_to_bytes(
+                documents, self.settings.LOCALE_ENCODING, ignore_errors, logger
+            ):
+                yield document
+
+            return
+
```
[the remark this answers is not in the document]
SergioSim: Good point) It's indeed not ideal to have this here, however, it's just an intermediary step that will be removed in PR #515 (changed [here](https://github.com/openfun/ralph/pull/515/files#diff-e07e16cfa7f7121d12e4a323c09c1d74567acca8333c0371a7f61f8afa76f2baL111-L159))
It's purpose is to simplify the review of the latter.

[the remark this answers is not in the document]
wilbrdt: Alright!

src/ralph/utils.py
SergioSim: Great idea) Thanks) Have updated the implementation as you suggested:

```python
def iter_by_batch(iterable: Iterable, n: int) -> Iterable[list]:
    """Batch `iterable` into lists of length `n`. The last batch may be less than `n`.

    Taken from itertools.batched introduced in python 3.12
    """
    if n < 1:
        raise ValueError("n must be at least one")

    it = iter(iterable)
    while batch := list(islice(it, n)):
        yield batch
```
There are two slight differences :
- Could we keep yielding lists (it seems the Mongo data backend relies on the output being a list)?
- Could we postpone the method name change in a dedicated PR (to avoid rebase conflicts with other PRs)?

src/ralph/utils.py
```diff
@@ -171,39 +187,81 @@ def statements_are_equivalent(statement_1: dict, statement_2: dict) -> bool:
...
+            logger_class.error(msg, error, document, i)
+            raise BackendException(msg % (error, document, i)) from error
+
+        i += 1
+
+
+def iter_by_batch(iterable: Iterable, n: int) -> Iterable[list]:
+    """Batch `iterable` into lists of length `n`. The last batch may be less than `n`.
+
+    Taken from itertools.batched introduced in python 3.12
```
wilbrdt, line 257:
```suggestion
    Inspired by itertools.batched introduced in python 3.12
```

src/ralph/utils.py
wilbrdt: Fine with this implementation! And we can stick with this name too!
I added a small remark on the docstring though

src/ralph/utils.py
SergioSim: Thanks)

src/ralph/utils.py
SergioSim: Thanks) Updated it)

PR #511, opened 2023-11-13
♻️(backends) add logger property to data/lrs backends

SergioSim:
## Purpose

Currently, backends use a module-level logger instance for logging to allow filtering logs by backend.
However, when a backend inherits another backend, logs produced by the parent backend refer to the parent module (and not the current backend module).

## Proposal

- [x] make the logger an instance property of the backend class to allow a child backend to overwrite the logger of its parent
- [x] move `HistoryMixin` to the data backed module to allow its logger to be overwritten too.

**Note**: This PR depends on #510

### DIFF
--- a/src/ralph/backends/data/base.py
+++ b/src/ralph/backends/data/base.py
@@ -1,9 +1,9 @@
 """Base data backend for Ralph."""
 
-import functools
 import logging
 from abc import ABC, abstractmethod
 from enum import Enum, unique
+from functools import cached_property, wraps
 from io import IOBase
 from typing import Any, Generic, Iterable, Iterator, Optional, Type, TypeVar, Union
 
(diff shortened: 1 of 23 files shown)


src/ralph/backends/data/base.py
```diff
@@ -79,6 +77,15 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+class Loggable:
+    """A class with a `logger` property."""
+
+    @functools.cached_property
```
wilbrdt, line 83: nit: We could import `cached_property` from `functools`
```suggestion
    @cached_property
```

wilbrdt (approved):
Alt text

src/ralph/backends/data/base.py
SergioSim: Well spotted) Thanks) I have updated [this](https://github.com/openfun/ralph/pull/521/files#diff-abc4154695f25bd1399562f7c5e9f27fe9fba44ce29e46d606b46f1871b932a6R89) in PR #521 - could we postpone the change here?

SergioSim: After thought, I think it would be better to make this change here) will rebase #521 )

wilbrdt: Alright, as you prefer! :raised_hands:

PR #512, opened 2023-11-13
♻️(backends) fix backend type hints and refactor common settings

SergioSim:
## Purpose

The `DEFAULT_CHUNK_SIZE` and `LOCALE_ENCODING` settings are shared among data backends, however, currently, they are redefined in each backend. 

The async data backend read method was wrongly typed to return an `Iterator`. As it's an async generator it should return an `AsyncIterator`.

## Proposal

- [x] move  `DEFAULT_CHUNK_SIZE` and `LOCALE_ENCODING` settings to `BaseDataBackendSettings`
- [x] update type hints for data backend read method
- [x] add type hint for data backend `client` property

**Note:** This PR depends on #511


src/ralph/backends/data/async_mongo.py
```diff
@@ -280,8 +268,19 @@ async def close(self) -> None:
...
+    def _get_target_collection(
+        self, target: Union[str, None]
+    ) -> AsyncIOMotorCollection:
+        """Return the target collection."""
+        try:
+            return self.database[target] if target else self.collection
+        except InvalidName as error:
+            msg = "The target=`%s` is not a valid collection name: %s"
+            self.logger.error(msg, target, error)
+            raise BackendParameterException(msg % (target, error)) from error
```
wilbrdt, line 280: It was already covered by the tests, right?

src/ralph/backends/data/clickhouse.py
```diff
@@ -197,6 +193,9 @@ def list(
             self.logger.error(msg, error)
             raise BackendException(msg % error) from error
 
+        if new:
+            self.logger.warning("The `new` argument is ignored")
+
```
wilbrdt, line 198: We should also probably update the docstring for this argument

wilbrdt (commented):
Alt text

src/ralph/backends/data/async_mongo.py
SergioSim: Well spotted) Thanks) There was indeed a missing test for the write method with an invalid `target`) added it)

src/ralph/backends/data/clickhouse.py
SergioSim: Nice catch) Thanks) Updated the docstring)

src/ralph/backends/data/base.py
```diff
@@ -473,3 +453,45 @@ async def close(self) -> None:
...
+) -> Optional[Type]:
+    """Return the generic argument of `backend_class` at specified `position`."""
+    if not hasattr(backend_class, "__orig_bases__"):
+        return None
+
+    bases = backend_class.__orig_bases__[0]
+    if not hasattr(bases, "__args__") or len(bases.__args__) < abs(position) + 1:
+        return None
+
+    argument = bases.__args__[position]
```
wilbrdt, line 469: Why not using `typing.get_args` here? (https://docs.python.org/3/library/typing.html#typing.get_args)

src/ralph/backends/data/base.py
```diff
@@ -473,3 +453,45 @@ async def close(self) -> None:
         Raise:
             BackendException: If a failure occurs during the close operation.
         """
+
+
+def get_backend_generic_argument(
+    backend_class: Type[Union[BaseDataBackend, BaseAsyncDataBackend]], position: int
```
wilbrdt, line 459: Would it be possible to not rely on position of arguments?

src/ralph/backends/data/base.py
SergioSim: Good point) This was [previously mentioned](https://github.com/openfun/ralph/pull/501#discussion_r1399475889) - we did this change in [PR 527](https://github.com/openfun/ralph/pull/527/files#diff-abc4154695f25bd1399562f7c5e9f27fe9fba44ce29e46d606b46f1871b932a6L450-R461) as we have to re-change this method to support a `query_class` of type `BaseQuery` or `str`)

wilbrdt: Nice, OK!

wilbrdt (approved):

src/ralph/backends/data/base.py
wilbrdt: Fine by me!

PR #514, opened 2023-11-13
✅(api) enhance limit query parameter validation

lebaudantoine:
## Purpose

Adjusting response handling to return a more appropriate 422 HTTP exception
instead of a generic 500 error when the limit query parameter is strictly negative.

## Proposal

Use the arguments of the Query class. Set the greater or equal (`ge`) parameter's value to 0.


wilbrdt (approved):
LGTM! :tada:

PR #515, opened 2023-11-13
🏗️(backends) use template method for read method in data backends

SergioSim:
## Purpose

Data backends implicitly rely on the `enforce_query_checks` decorator to validate the query argument for the read method.
Also, other common tasks like setting a default `chunk size` are repeated among data backends.

## Proposal

To make query argument validation more explicit and automatize common tasks, we propose using a template pattern.
We move core backend implementation to private `_read_dicts` and `_read_bytes` methods and make the public `read` method perform common tasks (e.g., query validation) and call the private methods.

- [x] add private abstract `_read_dicts` and `_read_bytes` methods for data backends
- [x] make the `BaseDataBackend.read` method non-abstract.


src/ralph/backends/data/base.py
```diff
@@ -167,6 +153,48 @@ def list(
...
+            errors = error.errors()
+            logger.error(msg, query_name, errors)
+            raise BackendParameterException(msg % (query_name, errors)) from error
+
+    if isinstance(query, query_class):
+        return query
+
+    msg = "The 'query' argument is expected to be a %s instance."
+    logger.error(msg, query_name)
+    raise BackendParameterException(msg % (query_name,))
```
wilbrdt, line 195:
```suggestion
    raise BackendParameterException(msg % query_name)
```

tests/backends/data/test_base.py
```diff
@@ -23,25 +22,26 @@
...
 
         def status(self):
             pass
 
         def close(self):
             pass
 
-    MockBaseDataBackend().read(query=value)
+    assert list(MockBaseDataBackend().read(query=value, raw_output=True)) == [True]
+    assert list(MockBaseDataBackend().read(query=value, raw_output=True)) == [True]
```
wilbrdt, line 44:
```suggestion
    assert list(MockBaseDataBackend().read(query=value, raw_output=False)) == [True]
```

wilbrdt (commented):
Alt text
Looks good to me, have only minor suggestions

tests/backends/data/test_base.py
SergioSim: Oups) Nice catch) Thanks)

wilbrdt (approved):
Looks good to me! :tada:

PR #518, opened 2023-11-13
✨(backends) add max_statements option to data backends

SergioSim:
## Purpose

To align `data` backends with `http` backends, we add the `max_statements` option to the `read` method, allowing users to specify the maximum number of yielded statements.

## Proposal

- [x] add `max_statements` option to data backend `read` method

**Note:** This PR depends on #517

### DIFF
--- a/src/ralph/backends/data/async_es.py
+++ b/src/ralph/backends/data/async_es.py
@@ -5,6 +5,7 @@ from typing import AsyncIterator, Iterable, Optional, TypeVar, Union
 
 from elasticsearch import ApiError, AsyncElasticsearch, TransportError
 from elasticsearch.helpers import BulkIndexError, async_streaming_bulk
+from pydantic import PositiveInt
 
 from ralph.backends.data.base import (
     AsyncListable,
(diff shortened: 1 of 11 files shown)


src/ralph/backends/data/base.py
```diff
@@ -294,6 +295,7 @@ def read(  # noqa: PLR0913
             ignore_errors (bool): If `True`, encoding errors during the read operation
                 will be ignored and logged.
                 If `False` (default), a `BackendException` is raised on any error.
+            max_statements (int): The maximum number of statements to yield.
```
Leobouloc, line 298:
```suggestion
            max_statements (int): The maximum number of statements to yield. If `None` (default), there is no maximum.
```

Idem for similar docstrings.

src/ralph/backends/data/async_es.py
```diff
@@ -114,6 +114,7 @@ async def read(  # noqa: PLR0913
         chunk_size: Optional[int] = None,
         raw_output: bool = False,
         ignore_errors: bool = False,
+        max_statements: Optional[int] = None,
```
Leobouloc, line 117: Perhaps we could use `PositiveInt` ?

Leobouloc (approved):
Good for me ! I have only two minor comments regarding the docstrings and using PositiveInt instead of int.

src/ralph/backends/data/base.py
SergioSim: Good Idea) Thanks) Added it)

src/ralph/backends/data/async_es.py
SergioSim: Good point) It is indeed more explicit to use `PositiveInt`) Updated it)

PR #519, opened 2023-11-14
✨(backends) split backend chunk size setting for read/write methods

SergioSim:
## Purpose

Currently, backends use a shared `DEFAULT_CHUNK_SIZE` setting in the `read` and `write` methods.
We want to allow having different default chunk size settings for both methods.

## Proposal

- [x] split the `DEFAULT_CHUNK_SIZE` setting into `READ_CHUNK_SIZE` and `WRITE_CHUNK_SIZE`.

**Note:** This PR depends on #518
**Note:** This feature was originally suggested by @Leobouloc


tests/backends/data/test_async_es.py
```diff
@@ -87,24 +89,26 @@ async def test_backends_data_async_es_instantiation_with_settings():
...
         ALLOW_YELLOW_STATUS=True,
         CLIENT_OPTIONS={"verify_certs": False, "ca_certs": None},
-        DEFAULT_CHUNK_SIZE=5000,
         DEFAULT_INDEX=ES_TEST_INDEX,
         HOSTS=["https://elasticsearch_hostname:9200"],
         LOCALE_ENCODING="utf-16",
         POINT_IN_TIME_KEEP_ALIVE="5m",
+        READ_CHUNK_SIZE=5000,
         REFRESH_AFTER_WRITE=True,
+        WRITE_CHUNK_SIZE=5000,
```
Leobouloc, line 98:
```suggestion
        WRITE_CHUNK_SIZE=4999,
```
(to have a different test value than `READ_CHUNK_SIZE`)

(same for all tests below)

tests/test_cli.py
```diff
@@ -564,6 +564,42 @@ def mock_read(*_, **__):
     assert '{"foo": "bar"}' in result.output
 
 
+def test_cli_read_command_with_chunk_size(fs, monkeypatch):
+    """Test ralph read command with a chunk_size option."""
```
Leobouloc, line 568:
```suggestion
    """Test ralph `read` command with a `chunk_size` option."""
```

tests/test_cli.py
```diff
@@ -564,6 +564,42 @@ def mock_read(*_, **__):
...
+
+    def get_mock_read_bytes(expected_chunk_size: int):
+        """Return a `_read_bytes` function checking the `chunk_size` option."""
+
+        def mock_read_bytes(self, query, target, chunk_size, *_):
+            """Check chunk_size and return an archive."""
+            assert chunk_size == expected_chunk_size
+            yield b'{"foo": "bar"}'
```
Leobouloc, line 576: Shouldn't this be `yield b'{"foo": "bar"}'[:chunk_size]` , or am I missing something ?

Leobouloc (commented):
Thank you for taking care of this ! Looks good to me.

My only suggestion is putting different default values for read and write in tests, to be sure they are not inverted.

tests/backends/data/test_async_es.py
SergioSim: Good Idea) Thanks) Updated the write chunk size values)

tests/test_cli.py
SergioSim: Nice catch) Thanks) updated this)

tests/test_cli.py
SergioSim: You are right) That would be more rigorous) Updated the test)

wilbrdt (approved):
Alt text

PR #520, opened 2023-11-14
✨(backends) add simultaneous option for AsyncWritable data backends

SergioSim:
## Purpose

We try to align the data backend interface with the http backends.
The `simultaneous` option allows the caller of an async data backend to write data concurrently.

## Proposal

- [x] add `simultaneous` and `max_num_simultaneous` options to async Writable data backends

**Note:** This PR depends on #519

Note: The current behavior of the `max_num_simultaneous` option differs slightly with the http backed implementation:
- We log a warning instead of raising an exception when  `max_num_simultaneous`  is set and `simultaneous` is `False`
- We do not accept unlimited concurrency when `max_num_simultaneous` is set to `None` - we default to the sequential behavior instead.

### DIFF
--- a/src/ralph/backends/data/async_es.py
+++ b/src/ralph/backends/data/async_es.py
@@ -217,6 +217,7 @@ class AsyncESDataBackend(
         chunk_size: Optional[int] = None,
         ignore_errors: bool = False,
         operation_type: Optional[BaseOperationType] = None,
+        concurrency: Optional[PositiveInt] = None,
     ) -> int:
         """Write data documents to the target index and return their count.
 
@@ -232,6 +233,8 @@ class AsyncESDataBackend(
             operation_type (BaseOperationType or None): The mode of the write operation.
                 If `operation_type` is `None`, the `default_operation_type` is used
                 instead. See `BaseOperationType`.
+            concurrency (int): The number of chunks to write concurrently.
+                If `None` it defaults to `1`.
 
         Return:
             int: The number of documents written.
@@ -243,7 +246,7 @@ class AsyncESDataBackend(
                 supported.
         """
         return await super().write(
-            data, target, chunk_size, ignore_errors, operation_type
+            data, target, chunk_size, ignore_errors, operation_type, concurrency
         )
 
     async def _write_bytes(  # noqa: PLR0913
--- a/src/ralph/backends/data/async_mongo.py
+++ b/src/ralph/backends/data/async_mongo.py
@@ -196,6 +196,7 @@ class AsyncMongoDataBackend(
         chunk_size: Optional[int] = None,
         ignore_errors: bool = False,
         operation_type: Optional[BaseOperationType] = None,
+        concurrency: Optional[PositiveInt] = None,
     ) -> int:
         """Write data documents to the target collection and return their count.
 
@@ -210,6 +211,8 @@ class AsyncMongoDataBackend(
             operation_type (BaseOperationType or None): The mode of the write operation.
                 If `operation_type` is `None`, the `default_operation_type` is used
                     instead. See `BaseOperationType`.
+            concurrency (int): The number of chunks to write concurrently.
+                If `None` it defaults to `1`.
 
         Return:
             int: The number of documents written.
@@ -221,7 +224,7 @@ class AsyncMongoDataBackend(
                 supported.
         """
         return await super().write(
-            data, target, chunk_size, ignore_errors, operation_type
+            data, target, chunk_size, ignore_errors, operation_type, concurrency
         )
 
     async def _write_bytes(  # noqa: PLR0913
--- a/src/ralph/backends/data/base.py
+++ b/src/ralph/backends/data/base.py
@@ -23,6 +23,7 @@ from pydantic import BaseModel, BaseSettings, PositiveInt, ValidationError
 
 from ralph.conf import BaseSettingsConfig, core_settings
 from ralph.exceptions import BackendParameterException
+from ralph.utils import gather_with_limited_concurrency, iter_by_batch
 
 
 class BaseDataBackendSettings(BaseSettings):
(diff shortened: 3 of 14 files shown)


src/ralph/backends/data/async_es.py
```diff
@@ -234,6 +236,10 @@ async def write(  # noqa: PLR0913
             operation_type (BaseOperationType or None): The mode of the write operation.
                 If `operation_type` is `None`, the `default_operation_type` is used
                 instead. See `BaseOperationType`.
+            simultaneous (bool): If `True`, chunks will be written concurrently.
+                If `False` (default), chunks will be written sequentially.
+            max_num_simultaneous (int or None): If simultaneous is `True`, the maximum
+                number of chunks to write concurrently. If `None` it defaults to 1.
```
Leobouloc, line 242: Maybe we could remove the `simultaneous` parameter ? and only have `max_num_simultaneous` (default: 1). If `None`, there would be no maximum ?

src/ralph/backends/data/async_mongo.py
```diff
@@ -203,6 +205,10 @@ async def write(  # noqa: PLR0913
             operation_type (BaseOperationType or None): The mode of the write operation.
                 If `operation_type` is `None`, the `default_operation_type` is used
                     instead. See `BaseOperationType`.
+            simultaneous (bool): If `True`, chunks will be written concurrently.
+                If `False` (default), chunks will be written sequentially.
+            max_num_simultaneous (int or None): If simultaneous is `True`, the maximum
+                number of chunks to write concurrently. If `None` it defaults to 1.
```
Leobouloc, line 211: same comment as above: maybe we can remove simultaneous ?

src/ralph/backends/data/base.py
```diff
@@ -132,15 +133,15 @@ def write(  # noqa: PLR0913
...
+            msg = f"{operation_type.value.capitalize()} operation_type is not allowed"
             self.logger.error(msg)
             raise BackendParameterException(msg)
 
         data = iter(data)
         try:
             first_record = next(data)
         except StopIteration:
-            self.logger.info("Data Iterator is empty; skipping write to target.")
+            self.logger.info("Data Iterator is empty; skipping write to target")
```
Leobouloc, line 144: Wondering if this should be `warning` ?

tests/backends/data/test_async_es.py
```diff
@@ -475,6 +475,37 @@ async def test_backends_data_async_es_read_with_query(es, async_es_backend, capl
...
+    backend = async_es_backend()
+    data = ({"value": str(idx)} for idx in range(5))
+
+    with caplog.at_level(logging.INFO):
+        assert (
+            await backend.write(
+                data, chunk_size=3, simultaneous=True, max_num_simultaneous=2
+            )
+            == 5
+        )
```
Leobouloc, line 494: Unless I am mistaken, this does not test the actual simultaneity. 

For `AsyncLRSHTTP`, there are timing-based tests that are flagged to be skipped but are intended to be kept for dev purpose (There is perhaps a better solution):

https://github.com/openfun/ralph/blob/0bfcd837f3501699ffd5d994af3656a8506a9bc9/tests/backends/http/test_async_lrs.py#L1009

src/ralph/backends/data/async_es.py
SergioSim: Great idea) How about renaming it to `concurrency=1`?
I'm hesitating about allowing unlimited concurrency - wouldn't it be problematic if the `data` is unbound?

tests/backends/data/test_async_es.py
SergioSim: You are right) Maybe we should limit this test to only check that the `similtaneous` parameter is passed to the parent `write` method? The main testing of simultaneity is done in `test_backends_data_base_async_write_with_simultaneous`, however, indeed, for now, we do not test the timing (or order) of the write calls - would counting the number of times `gather_with_limited_concurrency` is called suffice (to avoid timing-based tests)? WDYT?

src/ralph/backends/data/async_es.py
Leobouloc: Agreed, unlimited concurrency is probably not a good idea.

`concurency=1` is fine with me !

SergioSim: Updated with `concurrency`)

tests/backends/data/test_async_es.py
SergioSim: Updated the `test_backends_data_base_async_write_with_concurrency` with a call counter for `gather_with_limited_concurrency`)

src/ralph/backends/data/base.py
```diff
@@ -383,22 +387,40 @@ async def write(  # noqa: PLR0913
...
         chunk_size = chunk_size if chunk_size else self.settings.WRITE_CHUNK_SIZE
         is_bytes = isinstance(first_record, bytes)
         writer = self._write_bytes if is_bytes else self._write_dicts
-        return await writer(data, target, chunk_size, ignore_errors, operation_type)
+
+        concurrency = concurrency if concurrency else 1
+        if concurrency == 1:
+            return await writer(data, target, chunk_size, ignore_errors, operation_type)
+
+        if concurrency < 1:
```
quitterie-lcs, line 410: We can type `concurrency` with `PositiveInt` from `pydantic` in the method signature. WDYT?

SergioSim: Good Idea) Thanks) Updated it)

quitterie-lcs (approved):

src/ralph/backends/data/base.py
wilbrdt: IMO info level is sufficient, it's the caller fault and no action has been made

PR #526, opened 2023-11-21
🏗️(backends) migrate WSStreamBackend to AsyncWSDataBackend

SergioSim:
## Purpose

The interface of the `stream` backends is quite similar to the `data` backends.
We want to limit the number of backend interfaces.

## Proposal

- [x] migrate `WSStreamBackend` to `AsyncWSDataBackend`


src/ralph/backends/data/async_ws.py
```diff
@@ -0,0 +1,186 @@
...
+        chunk_size: int,  # noqa: ARG002
+        ignore_errors: bool,  # noqa: ARG002
+    ) -> AsyncIterator[bytes]:
+        """Method called by `self.read` yielding bytes. See `self.read`."""
+        uri = target if target else self.settings.URI
+        client = await self.client(uri)
+        count = 0
+        try:
+            while event := await client.recv():
+                yield bytes(f"{event}" + "\n", encoding=self.settings.LOCALE_ENCODING)
```
jmaupetit, line 140:
```suggestion
                yield bytes(f"{event}\n", encoding=self.settings.LOCALE_ENCODING)
```

src/ralph/backends/data/async_ws.py
```diff
@@ -0,0 +1,186 @@
...
+                await client.close()
+            except (websockets.WebSocketException, OSError, TimeoutError) as error:
+                msg = "Failed to close websocket connection for %s: %s"
+                self.logger.error(msg, uri, error)
+                errors[uri] = error
+
+        self._clients = {}
+        if errors:
+            msg = "Failed to close websocket connections: %s"
+            raise BackendException(msg % errors)
```
jmaupetit, line 186: I'd rather raise an exception earlier when the first connection closing fails.

src/ralph/backends/data/async_ws.py
SergioSim: Nice catch) Thanks)

src/ralph/backends/data/async_ws.py
SergioSim: Good point) Why should we keep the remaining connections open if one fails? Wouldn't it be problematic if the caller catches the `BackendException` and continues execution?

src/ralph/utils.py
```diff
@@ -210,6 +210,29 @@ def parse_iterable_to_dict(
             raise BackendException(msg % (error, raw_document, i)) from error
 
 
+async def async_parse_iterable_to_dict(
+    raw_documents: AsyncIterable[T],
+    ignore_errors: bool,
+    logger_class: logging.Logger,
```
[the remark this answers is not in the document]
SergioSim: Good point) Indeed, injecting the logger is not ideal. I think, the idea was to allow configuring logging in a by-backend-basis.
As these methods are quite backend-specific, maybe we could move them into a mixin and use inheritance to provide them with their logger? Opened a draft PR #530 for an example) WDYT?

src/ralph/utils.py
```diff
@@ -210,6 +210,29 @@ def parse_iterable_to_dict(
...
+    parser: Callable[[T], Dict[str, Any]] = json.loads,
+    exceptions: Tuple[Type[Exception], ...] = (TypeError, json.JSONDecodeError),
+) -> AsyncIterator[dict]:
+    """Read the `raw_documents` Iterable and yield dictionaries."""
+    i = 0
+    async for raw_document in raw_documents:
+        try:
+            yield parser(raw_document)
+        except exceptions as error:
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks) Removed the `exceptions` argument)

src/ralph/backends/data/async_ws.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __init__(self, settings: Optional[WSDataBackendSettings] = None):
+        """Instantiate the async websocket data backend.
+
+        Args:
+            settings (WSDataBackendSettings or None): The websocket backend settings.
+                If `settings` is `None`, a default settings instance is used instead.
+        """
+        super().__init__(settings)
+        self._clients: Dict[str, websockets.WebSocketClientProtocol] = {}
```
jmaupetit, line 47: Why choosing to support multiple simultaneous connections instead of one per instance as it was previously implemented?

src/ralph/backends/data/async_ws.py
```diff
@@ -0,0 +1,183 @@
...
+
+    async def client(self, uri: str) -> websockets.WebSocketClientProtocol:
+        """Create an websocket client for the provided `URI` if it doesn't exist."""
```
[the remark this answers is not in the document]
SergioSim: Well spotted) Thanks)

src/ralph/backends/data/async_ws.py
SergioSim: Dropping multi-ws client support simplified this)

src/ralph/backends/data/async_ws.py
SergioSim: Good point) The idea was to align with other data backends and use the `target` argument for the URI—the fallback to the `settings.URI ` was implemented in the body of the `_read_bytes` method. However, I agree with you; limiting one connection per instance is much easier and straightforward to handle, Updated it)

src/ralph/backends/data/async_ws.py
```diff
@@ -0,0 +1,176 @@
...
+
+    async def status(self) -> DataBackendStatus:
+        """Implement data backend checks (e.g. connection, cluster status).
+
+        Return:
+            DataBackendStatus: The status of the data backend.
+        """
+        try:
+            await self.client()
+        except BackendException:
```
jmaupetit, line 68: Do you think it will work if the backend has already been instantiated?

jmaupetit: Why not trying a [`ping`](https://websockets.readthedocs.io/en/stable/reference/asyncio/client.html#websockets.client.WebSocketClientProtocol.ping) instead?

jmaupetit (commented):
Almost ready to ship! Great!

src/ralph/backends/data/async_ws.py
SergioSim: Nice catch) Thanks) Indeed, if the connection breaks after client instantiation, this does not work anymore) Updated using ping)

src/ralph/backends/data/async_ws.py
```diff
@@ -0,0 +1,228 @@
...
+        """
+        try:
+            client = await self.client()
+            pong_waiter = await client.ping()
+            await pong_waiter
+        except BackendException:
+            return DataBackendStatus.ERROR
+        except (websockets.WebSocketException, RuntimeError) as error:
+            self.logger.error("Failed to Ping %s: %s", self.settings.URI, error)
+            return DataBackendStatus.ERROR
```
jmaupetit, line 121: Maybe `AWAY` would be more appropriate?

```suggestion
            return DataBackendStatus.AWAY
```

SergioSim: Good point) Thanks) Updated it)

jmaupetit (approved):

PR #527, opened 2023-11-22
✨(backends) replace archive cli argument with query (rebased)

SergioSim:
## Purpose

The `archive` argument in the `read` cli command is not used anymore.
Instead, currently, a `--query` option is used, which only accepts JSON strings.
This complicates the usage of backends that do not require a query in the form of a dictionary (e.g.: fs, s3, ldp backends).

Example:
```
ralph read -b fs -q '{"query_string": "LICENSE.md"}'
```

We want to accept a plain string as the query argument to simplify CLI usage with backends that don't need a dictionary query.

Example:
```
ralph read -b fs LICENSE.md
```

Also, this would simplify the use of the ES data backend with a Lucene query string - e.g.: `ralph read -b es id:1`

Furthermore, the current usage of the `BaseQuery` model is complicated - not all backends need to define a custom query model and backends that do need a custom query model have to handle the case when the query is a string.

## Proposal

- [x] Replace the archive CLI argument with a query argument that accepts a JSON string or a plain string
- [x] Remove the previously defined query option, which was JSON only
- [x] Allow Backend `Query` to be of type `string`.
- [x] Remove `BaseQuery.query_string` field and handle JSON string queries in a dedicated `from_string` method.

Note: This PR rebases PR #491 to point to PR #515 to avoid merge conflicts. We deviated slightly from the original implementation to preserve the ability to use a Lucene query string for the ESDataBackend.

### DIFF
--- a/src/ralph/backends/data/base.py
+++ b/src/ralph/backends/data/base.py
@@ -3,7 +3,8 @@
 import logging
 from abc import ABC, abstractmethod
 from asyncio import Queue, create_task
-from enum import Enum, unique
+from enum import Enum, IntEnum, unique
+from inspect import isclass
 from io import IOBase
 from itertools import chain
 from typing import (
@@ -17,9 +18,12 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    get_args,
+    get_origin,
 )
 
 from pydantic import BaseModel, BaseSettings, PositiveInt, ValidationError
+from typing_extensions import Self, get_original_bases
 
 from ralph.conf import BaseSettingsConfig, core_settings
 from ralph.exceptions import BackendParameterException
@@ -58,7 +62,16 @@ class BaseQuery(BaseModel):
 
         extra = "forbid"
 
-    query_string: Union[str, None] = None
+    @classmethod
+    def from_string(cls, query: str) -> Self:
+        """Return an instance of BaseQuery from a string."""
+        try:
+            return cls.parse_raw(query)
+        except ValidationError as error:
+            msg = "Invalid %s query string: %s"
+            errors = error.errors()
+            logger.error(msg, cls.__name__, errors)
+            raise BackendParameterException(msg % (cls.__name__, errors)) from error
 
 
 @unique
@@ -112,7 +125,7 @@ class Writable(Configurable, ABC):
         """Write `data` records to the `target` container and return their count.
 
         Args:
-            data: (Iterable or IOBase): The data to write.
+            data (Iterable or IOBase): The data to write.
             target (str or None): The target container name.
                 If `target` is `None`, a default value is used instead.
             chunk_size (int or None): The number of records or bytes to write in one
@@ -210,11 +223,11 @@ class Listable(ABC):
 
 
 Settings = TypeVar("Settings", bound=BaseDataBackendSettings)
-Query = TypeVar("Query", bound=BaseQuery)
+Query = TypeVar("Query", bound=Union[BaseQuery, str])
 
 
 def validate_backend_query(
-    query: Union[str, dict, Query, None],
+    query: Optional[Query],
     query_class: Type[Query],
 ) -> Query:
     """Validate and transform the backend query."""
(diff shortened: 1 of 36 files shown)


src/ralph/backends/data/base.py
```diff
@@ -169,28 +181,22 @@ def validate_backend_query(
...
-            return query_class(query_string=query)
-        except ValidationError as error:
-            msg = "Invalid %s query string: %s"
-            errors = error.errors()
-            logger.error(msg, query_name, errors)
-            raise BackendParameterException(msg % (query_name, errors)) from error
+    if isinstance(query, query_class):
+        return query
+
+    if issubclass(query_class, BaseQuery) and isinstance(query, str):
```
Leobouloc, line 187: Why do we need `issubclass(query_class, BaseQuery) ` ? Shouldnt this always be True ?

src/ralph/backends/data/mongo.py
```diff
@@ -70,19 +70,20 @@ class Config(BaseSettingsConfig):
...
-    sort: Union[List[Tuple], None]
-
-
-class MongoQuery(BaseMongoQuery):
-    """MongoDB query model."""
+    Attributes:
+        filter (dict): A filter query to select which documents to include.
+        limit (int): The maximum number of results to return.
+        projection (dict): Dictionary specifying the fields to include or exclude.
+        sort (list): a list of (key, direction) pairs specifying the sort order.
```
Leobouloc, line 80:
```suggestion
        sort (list): A list of (key, direction) pairs specifying the sort order.
```

src/ralph/backends/data/s3.py
```diff
@@ -192,63 +181,73 @@ def read(  # noqa: PLR0913
...
+                "id": target + "/" + query,
                 "size": response["ContentLength"],
                 "timestamp": now(),
             }
         )
 
-    def _get_object(self, bucket: str, key: str) -> dict:
-        """Return S3 object wrapping the exception."""
+    def _get_object(self, bucket: Optional[str], key: str) -> dict:
+        """Validate bucket and key and return S3 object wrapping the exception."""
```
Leobouloc, line 240:
```suggestion
        """Validate bucket (target) and key (query) and return the S3 object."""
```
I don't understand "wrapping the exception"

src/ralph/backends/data/swift.py
```diff
@@ -206,63 +195,69 @@ def read(  # noqa: PLR0913
...
     ) -> Iterator[dict]:
         """Method called by `self.read` yielding dictionaries. See `self.read`."""
         target = target if target else self.default_container
-        self.logger.info(
-            "Getting object from container: %s (query_string: %s)",
-            target,
-            query.query_string,
-        )
-        resp_headers, content = self._get_object(target, query.query_string, chunk_size)
+        self.logger.info("Getting object: %s from target bucket: %s", query, target)
```
Leobouloc, line 233:
```suggestion
        self.logger.info("Getting object: %s from target container: %s", query, target)
```
?

tests/backends/data/test_base.py
```diff
@@ -114,8 +138,26 @@ class DummyAnyBackend(BaseDataBackend[DummySettings, Any]):
     assert get_backend_generic_argument(DummyAnyBackend, 1) is None
 
     # Given a backend that does not follow type hints, the function should return None.
-    class DummyBadBackend(BaseDataBackend[Union[dict, int], Union[int, str]]):
+    class DummyBadBackend(BaseDataBackend[Dict[dict, int], Any]):  # type: ignore
```
Leobouloc, line 141: I don't understand the choice of `Dict[dict, int]`

Leobouloc (commented):
This looks good to me !

src/ralph/backends/data/base.py
```diff
@@ -45,7 +48,16 @@ class Config:
 
         extra = "forbid"
 
-    query_string: Union[str, None] = None
+    @classmethod
+    def from_string(cls, query: str, logger: logging.Logger) -> Self:
+        """Return an instance of itself from a string."""
```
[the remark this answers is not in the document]
SergioSim: Thanks)

src/ralph/backends/data/mongo.py
SergioSim: Well spotted) Thanks)

src/ralph/backends/data/base.py
SergioSim: Good question) This is the key change of this PR - the `query_class` type is extended to `Union[BaseQuery, str, dict]`, which adds the need for this check

src/ralph/backends/data/base.py
```diff
@@ -553,25 +555,34 @@ async def close(self) -> None:
...
+    position: DataBackendArgument,
 ) -> Optional[Type]:
     """Return the generic argument of `backend_class` at specified `position`."""
-    if not hasattr(backend_class, "__orig_bases__"):
-        return None
+    for base in get_original_bases(backend_class):
+        if not (
+            hasattr(base, "__origin__")
+            and isclass(base.__origin__)
+            and issubclass(base.__origin__, (BaseDataBackend, BaseAsyncDataBackend))
```
wilbrdt, line 566: Here, could we use [`get_origin`](https://docs.python.org/3.12/library/typing.html#typing.get_origin)?

wilbrdt (approved):
I like it, way more clean! :1st_place_medal:

src/ralph/backends/data/base.py
SergioSim: Well spotted) Thanks) Updated it using get_origin)

quitterie-lcs (approved):

src/ralph/backends/data/base.py
SergioSim: Update: This change was removed as we chose to drop support for converting string queries to `BaseQuery` if the `query_class` is `BaseQuery`.

PR #529, opened 2023-11-24
💥(project) drop support for python3.7 and unpin dependencies

SergioSim:
## Purpose

As a significant part of `ralph` no longer supports python3.7, it might be time to drop support for python3.7 officially.

## Proposal

- [x] drop support for `python3.7`
- [x] unpin python dependencies (pinned to support python3.7)


Leobouloc (approved):
Adios 3.7

PR #531, opened 2023-11-30
🔧(project) add aliases for grouping extra dependencies

wilbrdt:
## Purpose

Ralph package has a lot of extra dependencies, and it can be harsh having to list them when wanting to install a full package.

## Proposal

Adding three aliases:
- `backends` that groups all backends extra dependencies;
- `full` that groups all functional extra dependencies (`backends`, `cli` and `lrs`);
- `all` that groups all the extra dependencies (`full`, `ci` and `dev`).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -73,11 +73,18 @@ pip install ralph-malph[backend-es,lrs]
 pip install ralph-malph[backend-es,lrs,cli]
 ```
 
+- If you want to play around with backends with Ralph as a library, you can install: 
+
+```bash
+pip install ralph-malph[backends]
+```
+
 - If you have various uses for Ralph's features or would like to discover all the existing functionnalities, it is recommended to install the **full package**: 
 
 ```bash
-pip install ralph-malph[backend-clickhouse,backend-es,backend-ldp,backend-lrs,backend-mongo,backend-s3,backend-swift,backend-ws,cli,lrs]
+pip install ralph-malph[full]
 ```
+
 
 ### Install from DockerHub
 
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -120,6 +123,12 @@ lrs = [
     "python-jose==3.3.0",
     "uvicorn[standard]==0.24.0.post1",
 ]
+full = [
+    "ralph-malph[backends,cli,lrs]",
+]
+all = [
+    "ralph-malph[ci,dev,full]",
```
quitterie-lcs, line 130: We do not need `ci` in `all` (cf.what you have replaced in the `circleci/config.yam`l)

wilbrdt: `ci` extra is used only in one place, in the CI job that pushes `ralph-malph` package to PyPi. It would lengthen CI time, so I would keep this extra

jmaupetit, line 130: I would remove this alias.

pyproject.toml
```diff
@@ -120,6 +123,9 @@ lrs = [
     "python-jose==3.3.0",
     "uvicorn[standard]==0.24.0.post1",
 ]
+full = [
```
SergioSim, line 126: Should we follow alphabetical ordering here? (`full` before `lrs`?)

SergioSim (approved):
That's a nice feature) Thanks for adding it) Looks good :+1: 

image

quitterie-lcs (approved):

docs/index.md
quitterie-lcs, line 1: You can add a line for `backends` alias. It can be useful for users who want to play with backends only !

wilbrdt: Added!

PR #533, opened 2023-11-30
🏗️(backends) remove Loggable and logger injection

SergioSim:
## Purpose

We do not want to support logging on a `by-backend-basis` as it requires customizing logging and adds complexity.

## Proposal

- [x] remove the custom `Loggable` class
- [x] remove logger dependency injections in backend utilities


wilbrdt (approved):
Alt text
Looks good to me! :tada:

PR #534, opened 2023-11-30
🏗️(backends) add default implementations for abstract _write and _read methods

SergioSim:
## Purpose

It is common for backends to implement only one of _read_bytes/_read_dicts methods and reuse its implementation for the other one. (same for _write_dicts and _write_bytes) which causes some repetition among backends.

## Proposal

- [x] add default implementations for abstract write and read methods
- [x] make one of the _read/_write methods non-abstract (_read_bytes and _write_bytes)


wilbrdt (approved):
I think having circular references between default implementations can be quite confusing at first for developers wanting to add a new backend, but I get the idea!

PR #545, opened 2023-12-21
Prepare branch renaming to `main`

wilbrdt:
## Purpose

`master` naming for repos is deprecated in the tech industry.
`main` is now the popular choice, and it would be easier for devs that openfun default branches be named `main`.

## Proposal

- 🧑‍💻(project) prepare branch renaming from master to main
- 📝(project) rename `main` doc version to `dev`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,11 +46,11 @@ jobs:
       - run:
           name: enforce absence of print statements in code
           command: |
-            ! git diff origin/master..HEAD -- . ':(exclude).circleci' | grep "print("
+            ! git diff origin/main..HEAD -- . ':(exclude).circleci' | grep "print("
       - run:
           name: enforce absence of FIXME in code
           command: |
-            ! git diff origin/master..HEAD -- . ':(exclude).circleci' | grep "FIXME"
+            ! git diff origin/main..HEAD -- . ':(exclude).circleci' | grep "FIXME"
       - run:
           name: Check absence of fixup commits
           command: |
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -488,9 +488,9 @@ jobs:
             # - DOCS_VERSION: 1.1 (for git tag v1.1.2)
             # - DOCS_ALIAS: latest
             # or
-            # - DOCS_VERSION: master (for master branch)
+            # - DOCS_VERSION: dev (for main branch)
             # - No DOCS_ALIAS
-            DOCS_VERSION=$([[ -z "$CIRCLE_TAG" ]] && echo $CIRCLE_BRANCH || echo ${CIRCLE_TAG} | sed 's/^v\([0-9]\.[0-9]*\)\..*/\1/')
+            DOCS_VERSION=$([[ -z "$CIRCLE_TAG" ]] && echo "dev" || echo ${CIRCLE_TAG} | sed 's/^v\([0-9]\.[0-9]*\)\..*/\1/')
```
quitterie-lcs, line 493: Why do you change to echo "dev"?

wilbrdt: I think a "main" version for the docs would be confusing (whereas no confusion with "master").
I personally find a "dev" version for the docs to be straight to the point. WDYT?

quitterie-lcs: You're right! That's better !

quitterie-lcs (approved):

PR #561, opened 2024-01-30
✨(project) add multitenancy for user-specific index/database

wilbrdt:
## Purpose

When having a federal LRS, we want to separate statements for each organization, so that they are on different indexes or databases. It should enhance the data isolation and scalability across the platform.

## Proposal

- [x] Add multi-tenancy support: we can now specify a target index/database at user creation, allowing for data storage separation. This is optional, if no target is given at user creation, default target (specified in .env) will be used.
- [x] Add tests
- [x] Edit CLI `auth` command to add a user with a `target`
- Add documentation (can be done in another PR)


src/ralph/backends/lrs/fs.py
```diff
@@ -53,7 +53,9 @@ def write(  # noqa: PLR0913
         target = target if target else self.settings.DEFAULT_LRS_FILE
```
SergioSim, line 53:
```suggestion
        if target:
            target = str(Path(target) / Path(self.settings.DEFAULT_LRS_FILE))
        else:
            target = self.settings.DEFAULT_LRS_FILE
```
If I understand it right, with the `target` argument, we now read the `default_lrs_file` in a target directory:
```bash
- fs_lrs.jsonl  # without target
- target_A
    - fs_lrs.jsonl
- target_B
    - fs_lrs.jsonl
```
However, when we write - the target is the (destination) file, right?

SergioSim (commented):
This looks nice :+1: have only one minor question regarding fslrs backend.

image

src/ralph/backends/lrs/fs.py
wilbrdt: That's true, good point! Fixed it with your suggestion

src/ralph/backends/lrs/clickhouse.py
```diff
@@ -99,7 +101,7 @@ def query_statements(self, params: RalphStatementsQuery) -> StatementQueryResult
             clickhouse_response = list(
                 self.read(
                     query=query,
-                    target=self.event_table_name,
```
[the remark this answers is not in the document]
wilbrdt: To be consistent with other backends, we should remove the attribute indeed!

tests/api/test_statements_post.py
```diff
@@ -102,6 +102,71 @@ async def test_api_statements_post_single_statement_directly(  # noqa: PLR0913
...
 
+@pytest.mark.anyio
+@pytest.mark.parametrize(
+    "backend",
+    [
+        get_async_es_test_backend,
+        get_async_mongo_test_backend,
+        get_es_test_backend,
+        get_clickhouse_test_backend,
+        get_mongo_test_backend,
```
SergioSim, line 113: Missing `get_fs_test_backend`?

src/ralph/backends/lrs/fs.py
```diff
@@ -50,10 +51,15 @@ def write(  # noqa: PLR0913
 
         See `FSDataBackend.write`.
         """
-        target = target if target else self.settings.DEFAULT_LRS_FILE
+        if target:
+            target = str(Path(target) / Path(self.settings.DEFAULT_LRS_FILE))
```
SergioSim, line 55: Should we create the `target` directory if it doesn't exist?

jmaupetit (approved):

SergioSim (approved):
Looks great) :+1: 

Have only minor questions)
image

tests/api/test_statements_post.py
wilbrdt: Yes, it seems the `fs` backend is not tested for a POST. Will need to add it in another PR

src/ralph/backends/lrs/fs.py
wilbrdt: I'm not sure Ralph should be responsible for creating the directory, to stay consistent with the other backends. For the other backends, user needs to create the index, the collection or the table in advance

PR #562, opened 2024-02-05
🐛(backends) disallow namespace package discovery

SergioSim:
## Purpose

The automatic backend discovery by package name implemented in the `get_backends` method allowed the use of namespace package plugins, which is undesired. We want to support backend discovery using package metadata (entry points) exclusively in the future.

## Proposal

- [x] update the `get_backends` method to load backends from a predefined list of dotted backend paths.
- [x] add entry point backend discovery.
- [ ] update documentation. (in a separate PR?)

P.S.: I haven't noticed a change in performance with the addition of entry point backend discovery:

Tried: `perf stat -r 10 ./bin/ralph read` (which loads all backends to display the help page):
 - main (2.6699 +- 0.0402 seconds time elapsed)
 - this PR (2.7070 +- 0.0624 seconds time elapsed)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ keywords = ["LRS", "Analytics", "xAPI", "Open edX"]
 dependencies = [
     # By default, we only consider core dependencies required to use Ralph as a
     # library (mostly models).
+    "importlib-metadata>=7.0.1, <8.0",
     "langcodes>=3.2.0",
     "pydantic[dotenv,email]>=1.10.0, <2.0",
     "rfc3987>=1.3.0",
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -132,6 +133,12 @@ full = [
 [project.scripts]
 ralph = "ralph.__main__:cli.cli"
 
+[project.entry-points."ralph.backends.data"]
+ralph_data_backends = "ralph.backends.loader:DATA_BACKENDS"
```
SergioSim, line 137: Should an entry point contain a list of dotted backend paths or just a single backend? WDYT?

jmaupetit: I prefer listing them explicitly. Your comment's proposal looks great to me.

src/ralph/backends/loader.py
```diff
@@ -17,69 +19,76 @@
...
+    "ralph.backends.data.swift.SwiftDataBackend",
+]
+LRS_BACKENDS: List[str] = [
+    "ralph.backends.lrs.async_es.AsyncESLRSBackend",
+    "ralph.backends.lrs.async_mongo.AsyncMongoLRSBackend",
+    "ralph.backends.lrs.clickhouse.ClickHouseLRSBackend",
+    "ralph.backends.lrs.es.ESLRSBackend",
+    "ralph.backends.lrs.fs.FSLRSBackend",
+    "ralph.backends.lrs.mongo.MongoLRSBackend",
+]
```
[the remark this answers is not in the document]
SergioSim: Good point) Updated this part with entry points that point to a single backend)

wilbrdt (approved):
Looks good to me!
We'll need to add some documentation, but it can indeed be done in another PR!

PR #584, opened 2024-04-03
🐛(api) fix activity querying for statements with object missing a type

wilbrdt:
## Purpose

In xAPI version 1.0.3, the presence of `objectType` is optional, defaulting to `Activity` when not specified.

## Proposal

When querying activities, the LRS should exclusively search for `object.id` rather than first filtering on objects with type `Activity`.
This concerns all LRS backends.


quitterie-lcs (approved):
Sounds good!

PR #595, opened 2024-05-02
♻️(models) change default value "Agent" object type

quitterie-lcs:
## Purpose

Until now, "Agent" object type was set to None as default value. When using
ralph xapi models to generate statements instances, objectType Agent is
generated with None value which is not the required behavior. 

## Proposal

We keep this field as optional in Pydantic model for validation and force
"Agent" value as a default value.


wilbrdt (approved):
LGTM! :tada:

PR #630, opened 2026-06-15
🐛(models) fix groups' optional member field defaulting to null

piptouque:
## Purpose

[The `member` field is optional in an Identified Group ](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#details-3)
Without this fix, the mbox, mbox_sha1sum etc. kinds of groups still expect a member.

## Proposal

The `member` field of `BaseXxapiIdentifiedGroup` did not have a default value even though it is `Optional`.
This commit sets it to `None`, which is in line with other optional fields.

- [x] Added above fix
- [x] Added tests
- [x] Updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 ### Fixed
 
 - Fix type of `statement.result.score.scaled` from `int` to `Decimal`
+- Fix optional `member` field in Identified Groups not actually being optional
 
 ## [5.0.1] - 2024-07-11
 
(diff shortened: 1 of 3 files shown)


MYilFun00 (approved):
🐛(models) fix groups' optional member field defaulting to null (#630)

Add `= None` default on `BaseXapiIdentifiedGroup.member` so Pydantic v2
treats the field as truly optional per the xAPI spec. Without it, valid
Identified Groups without `member` were rejected with a 422.

PR #632, opened 2026-06-15
🐛(validation) replace `TypeError`s with `ValueError`s

piptouque:
## Purpose

[`TypeError`s are not supported as validation errors by Pydantic](https://pydantic.dev/docs/validation/latest/concepts/validators/ \#raising-validation-errors).
So Pydantic won't encapsulate them in a `ValidationError`.

## Proposal

We replace them with `ValueError`s, which work as expected.

- [x] Replaced `TypeError`s with `ValueError`s in Pydantic models
- [x] Updated tests
- [x] Updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 ### Fixed
 
 - Fix type of `statement.result.score.scaled` from `int` to `Decimal`
+- Fix misregistered `TypeError` in Pydantic models validation
 
 ## [5.0.1] - 2024-07-11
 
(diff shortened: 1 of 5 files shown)


MYilFun00 (approved):
Approving. `TypeError`s raised inside validators are not caught by Pydantic
as validation errors, so they were never wrapped in a `ValidationError` —
replacing them with `ValueError`s is the correct fix and makes the models
raise as expected.

Reviewed the change scope: `backends/lrs/base.py`, `models/xapi/base/common.py`
and the two updated test files. Validated locally on top of the latest `main`
(branch rebased/merged, CHANGELOG conflict resolved): the xapi validation
suite passes (124 passed) and the full CI is green.

Thanks @piptouque!

PR #633, opened 2026-06-15
✨(oidc) add support for generic OIDC IdPs and OIDC clients as users 

piptouque:
## Purpose

This PR contains multiple changes regarding OIDC

### Fixes

- The server should ignore any OAuth2 scope that it does not know, instead of returning an error.
- [The 'aud' claim may be a list](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3)
- [The 'exp' and 'iat' claims are `NumericDate` and may be floats](https://www.rfc-editor.org/rfc/rfc7519#section-2)

#### Support for generic OIDC IdPs

More specifically, support for IdPs that *return a token that is not a JWT*.

The ID token and access token are different and have different purpose.
The ID token is always a JWT and contains user claims, but the access token may not be.
With this change, we get the user claims using the access token,
using the `/userinfo` OIDC endpoint, allowing us to support providers that return opaque access tokens.

#### Support OIDC clients as 'users'
 
Client applications, as authenticated with the 'client_credentials' flow, do not have a dedicated user.
As such, we can't get ID tokens for them, always opaque access tokens.
Instead, we authenticate them using their `client_id`.

But first, we need to check whether the access token we got was from a real Oauth2 user or a OIDC client application.
To do that, we query the `/introspection` endpoint of our IdP.
This requires that Ralph be registered as another client app to our IdP, and to have configured its client ID and secret.

## Proposal

- [x] Fixed type of `IDToken` not following specs on `aud` (may be a list), `exp` and `iat` (may be floats)
- [x] Added query to `/introspection` endpoint when receiving an OIDC access token to determine if it comes from a ODIC client application or a user
- [x] Added query to `/userinfo` endpoint when receiving an OIDC access token if that token represents a user
- [x] Updated technical documentation
- [x] Updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to
 - Fix type of OIDC ID tokens
 - Fix error with OIDC scopes unrelated to Ralph
 
+### Changed
+
+- OIDC: Add query to `/userinfo` endpoint when receiving a token to support more OIDC IdPs
+
 ## [5.0.1] - 2024-07-11
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to
 
 - Fix type of `statement.result.score.scaled` from `int` to `Decimal`
 - Fix type of OIDC ID tokens
+- Fix error with OIDC scopes unrelated to Ralph
 
 ## [5.0.1] - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 ### Fixed
 
 - Fix type of `statement.result.score.scaled` from `int` to `Decimal`
+- Fix type of OIDC ID tokens
 
 ## [5.0.1] - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to
 ### Changed
 
 - OIDC: Add query to `/userinfo` endpoint when receiving a token to support more OIDC IdPs
+- OIDC: Add token introspection to support querying from OIDC clients (Client Credentials flow)
 
 ## [5.0.1] - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to
 ### Changed
 
 - Refactor statements' ExtensionMap
+- Fix type of OIDC ID tokens
 
 ## [5.0.1] - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ and this project adheres to
 - Fix type of OIDC ID tokens
 - Fix error with OIDC scopes unrelated to Ralph
 
+### Changed
+
+- OIDC: Add query to `/userinfo` endpoint when receiving a token to support more OIDC IdPs
+
 ## [5.0.1] - 2024-07-11
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,16 +29,13 @@ and this project adheres to
 - Bump the Elasticsearch test service to `8.8.1` so its bundled JDK can read
   cgroup v2 hosts, fixing the `test-python` CI jobs
 - Fix XAPI definitions extensions not accepting empty strings as values.
-
-### Changed
-
-- Refactor statements' ExtensionMap
 - Fix type of OIDC ID tokens
 - Fix error with OIDC scopes unrelated to Ralph
 
 ### Changed
 
-- Auth: changed default TTL of cache to 60 seconds 
+- Refactor statements' ExtensionMap
+- Auth: changed default TTL of cache to 60 seconds
 - OIDC: Add query to `/userinfo` endpoint when receiving a
   token to support more OIDC IdPs
 - OIDC: Add token introspection to support querying from OIDC clients
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,8 +38,11 @@ and this project adheres to
 
 ### Changed
 
-- OIDC: Add query to `/userinfo` endpoint when receiving a token to support more OIDC IdPs
-- OIDC: Add token introspection to support querying from OIDC clients (Client Credentials flow)
+- Auth: changed default TTL of cache to 60 seconds 
+- OIDC: Add query to `/userinfo` endpoint when receiving a
+  token to support more OIDC IdPs
+- OIDC: Add token introspection to support querying from OIDC clients
+  (Client Credentials flow)
 
 ## [5.0.1] - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and this project adheres to
 - Fix XAPI definitions extensions not accepting empty strings as values.
 - Fix type of OIDC ID tokens
 - Fix error with OIDC scopes unrelated to Ralph
+- Fix oidc test `test_api_auth_oidc_get_whoami_invalid_backend`
+  being misconfigured
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to
 
 - Refactor statements' ExtensionMap
 - Fix type of OIDC ID tokens
+- Fix error with OIDC scopes unrelated to Ralph
 
 ## [5.0.1] - 2024-07-11
 
--- a/docs/tutorials/lrs/authentication/oidc.md
+++ b/docs/tutorials/lrs/authentication/oidc.md
@@ -2,14 +2,20 @@
 
 Ralph LRS also supports OpenID Connect on top of OAuth 2.0 for authentication and authorization.
 
-To enable OpenID Connect authentication mode, we should change the `RALPH_RUNSERVER_AUTH_BACKENDS` environment variable to `oidc` and we should define the `RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI` environment variable with the identity provider's Issuer Identifier URI as follows:
+To enable OpenID Connect authentication mode, we should change the `RALPH_RUNSERVER_AUTH_BACKENDS` environment variable to `oidc` and we should define the environment variables as follows:
+
+- `RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI` the identity provider's Issuer Identifier URI
+  This address must be accessible to the LRS on startup as it will perform OpenID Connect Discovery to retrieve public keys and other information about the OpenID Connect environment.
+- `RALPH_RUNSERVER_AUTH_OIDC_CLIENT_ID` the OIDC client id issued by the identity provider for this instance
+- `RALPH_RUNSERVER_AUTH_OIDC_CLIENT_SECRET` the OIDC client secret issued by the identity provider for this instance
 
 ```bash
 RALPH_RUNSERVER_AUTH_BACKENDS=oidc
 RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI=http://{provider_host}:{provider_port}/auth/realms/{realm_name}
+RALPH_RUNSERVER_AUTH_OIDC_CLIENT_ID=some_client_id
+RALPH_RUNSERVER_AUTH_OIDC_CLIENT_SECRET=some_client_secret
 ```
 
-This address must be accessible to the LRS on startup as it will perform OpenID Connect Discovery to retrieve public keys and other information about the OpenID Connect environment.
 
 It is also strongly recommended to set the optional `RALPH_RUNSERVER_AUTH_OIDC_AUDIENCE` environment variable to the origin address of Ralph LRS itself (e.g. "http://localhost:8100") to enable verification that a given token was issued specifically for that Ralph LRS.
 
@@ -74,7 +80,7 @@ services:
 networks:
   ralph:
     external: true
-    
+
 ```
 
 Again, we need to create the `.ralph` directory:
@@ -102,7 +108,7 @@ Now that both Keycloak and Ralph LRS server are up and running, we should be abl
     ```
 
     ```bash
-    {"access_token":"<access token content>","expires_in":300,"refresh_expires_in":1800,"refresh_token":"<refresh token content>","token_type":"Bearer","not-before-policy":0,"session_state":"0889b3a5-d742-45fb-98b3-20e967960e74","scope":"email profile"} 
+    {"access_token":"<access token content>","expires_in":300,"refresh_expires_in":1800,"refresh_token":"<refresh token content>","token_type":"Bearer","not-before-policy":0,"session_state":"0889b3a5-d742-45fb-98b3-20e967960e74","scope":"email profile"}
     ```
 === "HTTPie"
 
@@ -134,12 +140,12 @@ Now that both Keycloak and Ralph LRS server are up and running, we should be abl
 With this access token, we can now make a request to the Ralph LRS server:
 
 === "curl"
-    
+
     ```bash
     curl -H 'Authorization: Bearer <access token content>' \
     http://localhost:8100/whoami
     ```
-    
+
     ```bash
     {"agent":{"openid":"http://localhost:8080/auth/realms/fun-mooc/b6e85bd0-ce6e-4b24-9f0e-6e18d8744e54"},"scopes":["email","profile"]}
     ```
--- a/src/helm/ralph/templates/cm_lrs.yaml
+++ b/src/helm/ralph/templates/cm_lrs.yaml
@@ -31,6 +31,7 @@ data:
   {{- if .Values.lrs.auth.oidc.enabled }}
   RALPH_RUNSERVER_AUTH_OIDC_AUDIENCE: {{ .Values.lrs.authOIDCAudience | quote }}
   RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI: {{ .Values.lrs.authOIDCIssuerURI | quote }}
+  RALPH_AUTH_OIDC_CACHE_TTL: {{ .Values.lrs.auth.oidc.cacheTTL | quote }}
    {{ end }}
 
   # Sentry
--- a/src/helm/ralph/templates/cm_lrs.yaml
+++ b/src/helm/ralph/templates/cm_lrs.yaml
@@ -9,7 +9,7 @@ data:
   RALPH_APP_DIR: {{ .Values.lrs.appDir | quote }}
   RALPH_RUNSERVER_BACKEND: {{ .Values.lrs.backend | quote }}
   RALPH_RUNSERVER_MAX_SEARCH_HITS_COUNT: {{ .Values.lrs.maxSearchHitsCount | quote }}
-  
+
   # CLI
   RALPH_CONVERTER_EDX_XAPI_UUID_NAMESPACE: {{ .Values.lrs.converterNamespace | quote }}
 
@@ -42,4 +42,3 @@ data:
   RALPH_SENTRY_LRS_TRACES_SAMPLE_RATE: {{ .Values.lrs.sentry.lrsSampleRate | quote }}
   RALPH_SENTRY_IGNORE_HEALTH_CHECKS: {{ .Values.lrs.sentry.ignoreHealthChecks | quote }}
   {{- end }}
-
--- a/src/helm/ralph/templates/cronjob.yaml
+++ b/src/helm/ralph/templates/cronjob.yaml
@@ -36,7 +36,7 @@ spec:
             {{- toYaml .Values.podSecurityContext | nindent 12 }}
           containers:
             - name: "{{ template "ralph.fullname" . }}-{{ $job.name }}"
-              securityContext: 
+              securityContext:
                 {{- toYaml .Values.securityContext | nindent 16 }}
               image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
               imagePullPolicy: {{ .Values.image.pullPolicy }}
--- a/src/helm/ralph/values.yaml
+++ b/src/helm/ralph/values.yaml
@@ -139,6 +139,7 @@ lrs:
       enabled: false
       audience: "http://localhost:8100"
       issuerURI: "http://learning-analytics-playground_keycloak_1:8080/auth/realms/fun-mooc"
+      cacheTTL: 60
   sentry:
     enabled: false
     dsn: "https://fake@key.ingest.sentry.io/1234567"
--- a/src/ralph/api/auth/oidc.py
+++ b/src/ralph/api/auth/oidc.py
@@ -2,7 +2,7 @@
 
 import logging
 from functools import lru_cache
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Union, get_args
 
 import requests
 from fastapi import Depends, HTTPException, status
@@ -12,7 +12,7 @@ from jose.exceptions import JWTClaimsError
 from pydantic import AnyUrl, BaseModel, ConfigDict
 from typing_extensions import Annotated
 
-from ralph.api.auth.user import AuthenticatedUser, UserScopes
+from ralph.api.auth.user import AuthenticatedUser, Scope, UserScopes
 from ralph.conf import settings
 
 OPENID_CONFIGURATION_PATH = "/.well-known/openid-configuration"
@@ -94,6 +94,19 @@ def get_public_keys(jwks_uri: AnyUrl) -> Dict:
         ) from exc
 
 
+def get_user_scopes(oidc_scopes: Optional[str]) -> UserScopes:
+    """Extract Ralph's custom OAuth2 scopes from the global scope list.
+
+    Ignore incompatible scopes.
+    """
+    compatible_scopes = (
+        [scope for scope in oidc_scopes.split(" ") if scope in get_args(Scope)]
+        if oidc_scopes
+        else []
+    )
+    return UserScopes(compatible_scopes)
+
+
 def get_oidc_user(
     auth_header: Annotated[Optional[HTTPBearer], Depends(oauth2_scheme)],
 ) -> AuthenticatedUser:
@@ -147,7 +160,7 @@ def get_oidc_user(
 
     user = AuthenticatedUser(
         agent={"openid": f"{id_token.iss}/{id_token.sub}"},
-        scopes=UserScopes(id_token.scope.split(" ") if id_token.scope else []),
+        scopes=get_user_scopes(id_token.scope),
         target=id_token.target,
     )
     return user
--- a/src/ralph/api/auth/oidc.py
+++ b/src/ralph/api/auth/oidc.py
@@ -2,7 +2,7 @@
 
 import logging
 from functools import lru_cache
-from typing import Dict, Optional
+from typing import Dict, Optional, Union
 
 import requests
 from fastapi import Depends, HTTPException, status
@@ -35,7 +35,7 @@ class IDToken(BaseModel):
     Attributes:
         iss (str): Issuer Identifier for the Issuer of the response.
         sub (str): Subject Identifier.
-        aud (str): Audience(s) that this ID Token is intended for.
+        aud (str or list of str): Audience(s) that this ID Token is intended for.
         exp (int): Expiration time on or after which the ID Token MUST NOT be
                    accepted for processing.
         iat (int): Time at which the JWT was issued.
@@ -45,9 +45,9 @@ class IDToken(BaseModel):
 
     iss: str
     sub: str
-    aud: Optional[str] = None
-    exp: int
-    iat: int
+    aud: Optional[Union[list[str], str]] = None
+    exp: float
+    iat: float
     scope: Optional[str] = None
     target: Optional[str] = None
 
--- a/src/ralph/api/auth/oidc.py
+++ b/src/ralph/api/auth/oidc.py
@@ -2,7 +2,7 @@
 
 import logging
 from functools import lru_cache
-from typing import Dict, Optional, Union, get_args
+from typing import Dict, Optional, Union, Literal, get_args
 
 import requests
 from fastapi import Depends, HTTPException, status
@@ -26,17 +26,19 @@ oauth2_scheme = OpenIdConnect(
 logger = logging.getLogger(__name__)
 
 
-class IDToken(BaseModel):
-    """Pydantic model representing the core of an OpenID Connect ID Token.
+class UserInfo(BaseModel):
+    """Pydantic model representing the core of an OpenID Connect UserInfo endpoint response.
 
-    ID Tokens are polymorphic and may have many attributes not defined in the
+    They are common to both the ID token and the UserInfo endpoint.
+
+    They are polymorphic and may have many attributes not defined in the
     specification. This model ignores all additional fields.
 
     Attributes:
         iss (str): Issuer Identifier for the Issuer of the response.
         sub (str): Subject Identifier.
-        aud (str or list of str): Audience(s) that this ID Token is intended for.
-        exp (int): Expiration time on or after which the ID Token MUST NOT be
+        aud (str or list of str): Audience(s) that this ID Token/UserInfo is intended for.
+        exp (int): Expiration time on or after which the ID Token/UserInfo MUST NOT be
                    accepted for processing.
         iat (int): Time at which the JWT was issued.
         scope (str): Scope(s) for resource authorization.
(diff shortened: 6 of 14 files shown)


MYilFun00 (changes requested):
Thanks for this OIDC refactor — the move to introspection + `/userinfo`, Client Credentials support, and the scopes fix are solid improvements. However I found several blocking issues during local review:

**Blocking**

**1. `src/ralph/api/auth/oidc.py` — `get_token_info` return type hint is wrong**

**Current implementation:**

```python
def get_token_info(...) -> UserInfo:
    ...
    return TokenInfo.model_validate(token_info)
```

The function returns `TokenInfo`, not `UserInfo`.

**Suggested fix:** change the annotation to `-> TokenInfo`.

**2. `tests/fixtures/auth.py` + `get_user_info_data` — JWT `/userinfo` handling**

**Current mock:**

```python
return (200, {"Content-Type": "application/jwt"}, json.dumps(encoded_user_info))
```

**Current production (`get_user_info_data`):**

```python
content_type = response.headers["Content-Type"]
return (response.json(), content_type.lower() == "application/jwt")
```

A real `/userinfo` endpoint with `Content-Type: application/jwt` returns the **raw JWT** in the response body. The current code calls `response.json()` unconditionally — this fails on a real IdP (`JSONDecodeError`). The mock passes **by accident** because `json.dumps(jwt)` makes `response.json()` return a Python string.

**Both fixes are required together** (tested locally):

**Suggested fix (mock):**

```python
return (200, {"Content-Type": "application/jwt"}, encoded_user_info)
```

**Suggested fix (production — `get_user_info_data`):**

```python
content_type = response.headers["Content-Type"]
media_type = content_type.split(";", 1)[0].strip().lower()
is_jwt = media_type == "application/jwt"
body = response.text if is_jwt else response.json()
return (body, is_jwt)
```

**Local test results (current vs fix):**

| Content-Type | Body | Current code | Fix (mock + prod) |
|---|---|---|---|
| `application/json` | JSON dict | ✅ dict | ✅ dict |
| `application/jwt` | raw JWT | ❌ `JSONDecodeError` | ✅ JWT string → `jwt.decode` OK |
| `application/jwt; charset=utf-8` | raw JWT | ❌ `JSONDecodeError` | ✅ JWT string |
| `application/jwt+json` | raw JWT | ❌ `JSONDecodeError` | ❌ (non-standard type, expected) |

⚠️ Applying the production fix **without** fixing the mock would still break tests: `response.text` on `json.dumps(jwt)` returns `"eyJ..."` (with JSON quotes) → `jwt.decode` fails. Both changes must land together.

**3. `src/ralph/api/auth/oidc.py` L202 — f-string syntax breaks Python < 3.12**

The multiline f-string in the `Authorization` header raises `SyntaxError` on Python 3.10/3.11. Project requires `>=3.9`, CI tests 3.9→3.12. Module import fails locally.

**Suggested fix:**

```python
basic = encode_client_secret_basic_token(client_id=client_id, client_secret=client_secret)
headers={"Authorization": f"Basic {basic}"}
```

**Non-blocking (nice to have / discuss before prod)**

**4. `@lru_cache()` keyed by access token — tested behaviour**

Verified locally: `@lru_cache()` defaults to **maxsize=128** (not unbounded). After 130 distinct tokens, `currsize=128` (LRU eviction). Same token twice → cache hit (useful for repeated requests on one connection).

**Remaining risks:**
- A **revoked** token can stay "valid" in cache until evicted (no TTL).
- Up to 128 token introspection/userinfo results kept in memory.

**Suggested fix (minimal):** remove `@lru_cache()` from token-specific functions; keep it only on stable data:

```python
@lru_cache(maxsize=1)   # discover_provider — OK (issuer config)
def discover_provider(...): ...

@lru_cache(maxsize=1)   # get_public_keys — OK (JWKS rotates rarely)
def get_public_keys(...): ...

# NO cache on get_token_info / get_user_info_data — tokens are short-lived
def get_token_info(...): ...
def get_user_info_data(...): ...
```

**Suggested fix (if caching is desired):** explicit bounded cache with TTL ≤ token lifetime:

```python
@lru_cache(maxsize=128)  # document the choice
def get_token_info(...): ...
# + add cache_clear() in tests (already done for basic auth in fixtures)
```

Or a TTL cache (e.g. 60s) so revoked tokens are not served stale for long.

---

**5. Double network call `/introspect` + `/userinfo` — flow analysis**

**Current flow** (`get_oidc_user`):
- **User token** (`token_info.sub` set): `POST /introspect` → `GET /userinfo` → compare `sub`
- **Client credentials** (`sub` absent): `POST /introspect` only ✅

So client-app auth is already optimized (1 call). User auth always costs 2 IdP round-trips vs 1 JWT decode on old `main`.

**Suggested optimizations (pick one):**
- **A (simplest):** if introspection already returns `scope` + `target` + `sub`, skip `/userinfo` when all Ralph claims are present (config flag `RUNSERVER_AUTH_OIDC_SKIP_USERINFO_IF_COMPLETE=true`).
- **B (safe default):** keep 2 calls but cache both per token with short TTL (see #4).
- **C (document only):** accept 2 calls as OIDC-correct, document IdP latency impact in `oidc.md`.

The `sub` cross-check between introspection and userinfo is valuable — do not remove without replacement.

---

**6. `Content-Type: application/jwt` detection — tested and verified**

Prefer parsing the media type over `startswith` (avoids false positive on `application/jwt+json`):

| Content-Type | `== "application/jwt"` | `startswith("application/jwt")` | `split(";")[0]` parse |
|---|---|---|---|
| `application/jwt` | ✅ | ✅ | ✅ |
| `application/jwt; charset=utf-8` | ❌ fails | ✅ (works) | ✅ (works) |
| `application/jwt+json` | ❌ | ⚠️ **false positive** | ✅ correctly rejects |

The recommended parse + `response.text` fix is **verified locally** against simulated IdP responses. See blocking issue **#2** above for the complete production code — it is the same fix, and must be paired with the mock correction.

Do **not** use `startswith("application/jwt")` alone — it would treat `application/jwt+json` as JWT.

---

**7. `conf.py` formatting**

**Current:**

```python
BASE_SETTINGS_CONFIG = SettingsConfigDict(
    case_sensitive=True, env_nested_delimiter="__", env_prefix="RALPH_", extra="ignore"
    , secrets_dir=os.environ.get("RALPH_SECRETS_DIR")
)
```

**Suggested fix:**

```python
BASE_SETTINGS_CONFIG = SettingsConfigDict(
    case_sensitive=True,
    env_nested_delimiter="__",
    env_prefix="RALPH_",
    extra="ignore",
    secrets_dir=os.environ.get("RALPH_SECRETS_DIR"),
)
```

---

**8. Long parametrize line in `test_oidc.py`**

Line 23 is a single 109-char entry (readable but dense). Split for lint/readability:

```python
@pytest.mark.parametrize(
    "runserver_auth_backends,userinfo_response_type",
    [
        ([AuthBackend.BASIC, AuthBackend.OIDC], "jwt"),
        ([AuthBackend.OIDC], "plain"),
        ([AuthBackend.OIDC], "jwt"),
    ],
)
```

Once the blocking issues above are addressed, please rebase on current `main` (includes CI fixes + #630/#632) before re-requesting review.

MYilFun00 (changes requested):
</p><hr><h1><strong>Request changes</strong></h1><p>Thanks for this OIDC refactor — the move to token introspection combined with <code>/userinfo</code>, the addition of Client Credentials support, and the scope handling improvements are all valuable changes.</p><p>However, during my local review, I identified several blocking issues, along with a few points worth discussing before production deployment.</p><h2>🔴 Blocking</h2><h3>1. <code>src/ralph/api/auth/oidc.py</code> — Incorrect return type annotation for <code>get_token_info</code></h3><p><strong>Current implementation:</strong></p><pre><code class="language-python">def get_token_info(...) -&gt; UserInfo:
    ...
    return TokenInfo.model_validate(token_info)
</code></pre><p>The function returns a <code>TokenInfo</code> instance, not a <code>UserInfo</code> instance.</p><p><strong>Suggested fix:</strong></p><p>Update the return type annotation to:</p><pre><code class="language-python">def get_token_info(...) -&gt; TokenInfo:
</code></pre><hr><h3>2. <code>tests/fixtures/auth.py</code> — JWT <code>/userinfo</code> mock does not reflect the real endpoint behavior</h3><p><strong>Current implementation:</strong></p><pre><code class="language-python">return (
    200,
    {"Content-Type": "application/jwt"},
    json.dumps(encoded_user_info),
)
</code></pre><p>A real <code>/userinfo</code> endpoint returning <code>Content-Type: application/jwt</code> sends the raw JWT in the response body, not a JSON-encoded string.</p><p>The current mock causes the tests to pass accidentally because <code>response.json()</code> parses <code>"eyJ..."</code> into a Python string.</p><p><strong>Suggested fix (test fixture):</strong></p><p>Return <code>encoded_user_info</code> directly, without <code>json.dumps()</code>.</p><p><strong>Suggested fix (production code – <code>get_user_info_data</code>):</strong></p><p>When the response content type is <code>application/jwt</code>, use <code>response.text</code> instead of <code>response.json()</code>.</p><hr><h3>3. <code>src/ralph/api/auth/oidc.py</code> (L202) — Multiline f-string is incompatible with Python &lt; 3.12</h3><p>The multiline f-string used to build the <code>Authorization</code> header raises a <code>SyntaxError</code> on Python 3.9–3.11.</p><p>Since the project supports <strong>Python &gt;= 3.9</strong> and the CI pipeline runs against Python 3.9 through 3.12, the module cannot be imported on supported versions.</p><p><strong>Suggested fix:</strong></p><pre><code class="language-python">basic = encode_client_secret_basic_token(
    client_id=client_id,
    client_secret=client_secret,
)

headers = {"Authorization": f"Basic {basic}"}
</code></pre><hr><h2>🟡 Non-blocking / Discussion points</h2><h3>4. <code>@lru_cache()</code> on access tokens — verified behavior</h3><p>Verified locally:</p><ul><li><p><code>@lru_cache()</code> defaults to <code>maxsize=128</code> (it is <strong>not</strong> unbounded).</p></li><li><p>After 130 distinct access tokens, <code>currsize=128</code> and LRU eviction occurs.</p></li><li><p>Reusing the same token correctly results in a cache hit.</p></li></ul><p>While the implementation works, a few concerns remain:</p><ul><li><p>A revoked token may continue to be considered valid until it is evicted (no TTL).</p></li><li><p>Up to 128 introspection/userinfo results may remain cached in memory.</p></li></ul><p><strong>Suggested minimal fix:</strong></p><p>Keep caching only for stable provider metadata:</p><pre><code class="language-python">@lru_cache(maxsize=1)
def discover_provider(...):
    ...


@lru_cache(maxsize=1)
def get_public_keys(...):
    ...


# No cache for token-specific data
def get_token_info(...):
    ...


def get_user_info_data(...):
    ...
</code></pre><p><strong>If caching is intentional:</strong></p><p>Use an explicit bounded cache and document the choice:</p><pre><code class="language-python">@lru_cache(maxsize=128)
def get_token_info(...):
    ...
</code></pre><p>or use a short TTL cache (e.g. 60 seconds) to reduce the risk of serving stale results for revoked tokens.</p><hr><h3>5. Double network call (<code>/introspect</code> + <code>/userinfo</code>) — flow analysis</h3><p>Current <code>get_oidc_user()</code> flow:</p><p><strong>User access token (<code>token_info.sub</code> present)</strong></p><pre><code>POST /introspect
GET /userinfo
Compare `sub`
</code></pre><p><strong>Client Credentials (<code>sub</code> absent)</strong></p><pre><code>POST /introspect only
</code></pre><p>The Client Credentials flow is already optimized (single network call).</p><p>However, user authentication always requires two IdP round trips, whereas the previous implementation only decoded the JWT locally.</p><p>The <code>sub</code> consistency check between introspection and <code>/userinfo</code> is valuable and should not be removed without an equivalent safeguard.</p><p>Possible improvements:</p><ul><li><p><strong>Option A (simplest):</strong> If introspection already provides all required Ralph claims (<code>scope</code>, <code>target</code>, <code>sub</code>, etc.), skip <code>/userinfo</code> (possibly behind a configuration flag such as <code>RUNSERVER_AUTH_OIDC_SKIP_USERINFO_IF_COMPLETE=true</code>).</p></li><li><p><strong>Option B:</strong> Keep both requests, but cache them per access token with a short TTL (see point #4).</p></li><li><p><strong>Option C:</strong> Keep the current behavior and document the additional IdP latency in <code>oidc.md</code>.</p></li></ul><hr><h3>6. <code>Content-Type: application/jwt</code> detection</h3><p>Verified locally:</p>
Content-Type | == "application/jwt" | startswith("application/jwt") | Media type parsing
-- | -- | -- | --
application/jwt | ✅ | ✅ | ✅
application/jwt; charset=utf-8 | ❌ | ✅ | ✅
application/jwt+json | ❌ | ❌ | ✅

<p>Using <code>startswith()</code> may incorrectly match unexpected media types.</p><p><strong>Suggested fix:</strong></p><p>Parse the media type before comparing:</p><pre><code class="language-python">media_type = content_type.split(";", 1)[0].strip().lower()

is_jwt = media_type == "application/jwt"

if is_jwt:
    body = response.text
else:
    body = response.json()

return (body, is_jwt)
</code></pre><p>Verified locally:</p><ul><li><p>A raw JWT body cannot be parsed with <code>json.loads()</code>.</p></li><li><p>The current mock using <code>json.dumps(jwt)</code> unintentionally hides this issue.</p></li></ul><hr><h3>7. <code>conf.py</code> formatting</h3><p><strong>Current:</strong></p><pre><code class="language-python">BASE_SETTINGS_CONFIG = SettingsConfigDict(
    case_sensitive=True, env_nested_delimiter="__", env_prefix="RALPH_", extra="ignore"
    , secrets_dir=os.environ.get("RALPH_SECRETS_DIR")
)
</code></pre><p><strong>Suggested fix:</strong></p><pre><code class="language-python">BASE_SETTINGS_CONFIG = SettingsConfigDict(
    case_sensitive=True,
    env_nested_delimiter="__",
    env_prefix="RALPH_",
    extra="ignore",
    secrets_dir=os.environ.get("RALPH_SECRETS_DIR"),
)
</code></pre><hr><h3>8. <code>test_oidc.py</code> — Long <code>parametrize</code> entry</h3><p>One of the parametrized entries exceeds the project's preferred line length.</p><p>For readability, consider formatting it as:</p><pre><code class="language-python">@pytest.mark.parametrize(
    "runserver_auth_backends,userinfo_response_type",
    [
        ([AuthBackend.BASIC, AuthBackend.OIDC], "jwt"),
        ([AuthBackend.OIDC], "plain"),
        ([AuthBackend.OIDC], "jwt"),
    ],
)
</code></pre><hr><p>Once the blocking issues above have been addressed, please rebase this branch onto the current <code>main</code> (which already includes the CI fixes as well as PRs <strong>#630</strong> and <strong>#632</strong>) before requesting another review.</p></body></html>

piptouque:
Thanks for the review.
Fixed most issues with the first commit, and added caching using TTL as suggested in the next commit.
Last commit is for setting the `AUTH_CACHING_TTL` to 60 (previously 3600) because I think this setting should be shared with Basic auth, and it seems that you find a shorter TTL preferable.

MYilFun00:
Thanks @piptouque. I re-tested the branch locally at `33bb61a`.

The three blocking points are resolved:

- **Point 1** — `get_token_info` is now annotated `-> TokenInfo` ✅
- **Point 2** — the mock returns the raw JWT (`encoded_user_info`, no `json.dumps`), and
  `get_user_info_data` parses the media type then picks `response.text` vs
  `response.json()` ✅
- **Point 3** — the multiline f-string is gone; `oidc.py`, `basic.py` and `conf.py` all
  parse under Python 3.10 ✅

And the caching rework is exactly what I was hoping for: `@lru_cache(maxsize=1)`
on `discover_provider` / `get_public_keys`, `TTLCache` with a `Lock` on the two
token-keyed functions. That closes both point 4 and point 5 (option B).

I verified point 2 and point 6 end to end against simulated IdP responses:

```
  OK       application/json                 is_jwt=False  body={'sub': 'u1'}
  OK       application/jwt                  is_jwt=True   body=eyJhbGciOiJIUzI1NiJ9...
  OK       application/jwt; charset=utf-8   is_jwt=True   body=eyJhbGciOiJIUzI1NiJ9...
```

The three valid cases behave correctly. The two remaining cases do not — see
below.

---

## 🔴 1. `HTTPException(error=...)` raises `TypeError` at runtime

`src/ralph/api/auth/oidc.py:150`

```python
raise HTTPException(
    status_code=status.HTTP_400_BAD_REQUEST,
    error="invalid_request",          # <- not a valid parameter
    detail=f"Invalid Media type in header: {media_type}, ...",
    headers={"WWW-Authenticate": "Bearer"},
)
```

FastAPI's signature is `HTTPException(status_code, detail=None, headers=None)`.
There is no `error` parameter, so this line raises:

```
TypeError: HTTPException.__init__() got an unexpected keyword argument 'error'
```

Continuing the run above with the media types that are supposed to hit this branch:

```
  TypeError  application/jwt+json    HTTPException.__init__() got an unexpected keyword argument 'error'
  TypeError  text/html               HTTPException.__init__() got an unexpected keyword argument 'error'
```

So an IdP answering with an unexpected `Content-Type` gets a 500 with a stack
trace instead of the intended 400. Nothing currently covers this path, which is
why the tests stay green.

**Suggested fix:**

```python
raise HTTPException(
    status_code=status.HTTP_400_BAD_REQUEST,
    detail=(
        f"Invalid media type in header: {media_type}, "
        "expected application/jwt or application/json"
    ),
    headers={"WWW-Authenticate": "Bearer"},
)
```

Splitting the string also clears the `E501` on line 153. A test with an
unexpected `Content-Type` would be worth adding — it is the only branch of this
function that is currently unexercised.

---

## 🔴 2. `ci/circleci: lint-git` — leftover debug `print`

`src/ralph/api/auth/oidc.py:145`

```python
content_type = response.headers["Content-Type"]
print(content_type)
```

The `lint-git` job explicitly rejects this:

```yaml
- run:
    name: enforce absence of print statements in code
    command: |
      ! git diff origin/main..HEAD -- . ':(exclude).circleci' | grep "print("
```

Reproduced locally — this single line is the only reason `lint-git` is red;
`gitlint` itself passes cleanly on all six commit messages. Removing the `print`
fixes the job. It also writes the content type to stdout on every authenticated
request, which you probably don't want in production logs.

---

## 🔴 3. `ci/circleci: lint` — 6 ruff errors + 1 black

```
src/ralph/api/auth/oidc.py:145:9   T201  `print` found                     (same line as above)
src/ralph/api/auth/oidc.py:153:89  E501  Line too long (115 > 88)          (fixed by the split above)
src/ralph/api/auth/oidc.py:222:34  F541  f-string without any placeholders
src/ralph/conf.py:3:1              I001  Import block is un-sorted
tests/conftest.py:3:1              I001  Import block is un-sorted
tests/helpers.py:3:1               I001  Import block is un-sorted
```

`F541` is on the new introspection header:

```python
"Authorization": f"Basic "
+ encode_client_secret_basic_token(client_id=client_id, client_secret=client_secret)
```

`f"Basic "` has no placeholder. Assigning first reads better and drops the
concatenation:

```python
basic_token = encode_client_secret_basic_token(
    client_id=client_id, client_secret=client_secret
)
...
headers={"Authorization": f"Basic {basic_token}"},
```

The three `I001` are auto-fixable with `ruff check --fix`.

`black --check` flags one file:

```diff
-from tests.fixtures.auth import AUDIENCE, ISSUER_URI,CLIENT_ID,CLIENT_SECRET
+from tests.fixtures.auth import AUDIENCE, ISSUER_URI, CLIENT_ID, CLIENT_SECRET
```

---

## 🔴 4. `ci/circleci: lint-changelog` — two lines over 80 characters

The job asserts that no `CHANGELOG.md` line exceeds 80 characters:

```
  90 chars | - OIDC: Add query to `/userinfo` endpoint when receiving a token to support more OIDC IdPs
  95 chars | - OIDC: Add token introspection to support querying from OIDC clients (Client Credentials flow)
```

There is also a trailing space on `- Auth: changed default TTL of cache to 60 seconds `.

**Suggested rewrap:**

```markdown
- Auth: changed default TTL of cache to 60 seconds
- OIDC: Add query to `/userinfo` endpoint when receiving a token,
  to support more OIDC IdPs
- OIDC: Add token introspection to support querying from OIDC
  clients (Client Credentials flow)
```

---

## 🟠 5. Worth discussing: `AUTH_CACHE_TTL` 3600 → 60 is shared with Basic auth

This is the one I'd like your opinion on before it lands.

You're right that the setting is shared — that's precisely the problem.
`src/ralph/api/auth/basic.py:102` uses the same `AUTH_CACHE_TTL` to cache
`get_basic_auth_user`, and what that cache is protecting is a bcrypt
verification. Measured locally:

```
bcrypt.checkpw average: 294.5 ms

TTL 3600s (current main) ->  1 bcrypt recomputation/hour/user =   295 ms/h
TTL   60s (this PR)      -> 60 bcrypt recomputations/hour/user = 17672 ms/h
```

So the change makes every Basic-auth user pay a ~300 ms penalty once a minute
instead of once an hour, a 60× increase in authentication CPU cost. bcrypt is
deliberately slow, so this is not a micro-optimisation.

The reason a short TTL is desirable for OIDC does not apply to Basic auth: OIDC
access tokens are revocable at the IdP, so serving a stale introspection result
matters. Basic credentials live in a local file and change only when an operator
edits it.

**Suggested fix — split the two:**

```python
# conf.py
AUTH_CACHE_TTL: int = 3600            # basic auth, unchanged
AUTH_OIDC_CACHE_TTL: int = 60         # IdP responses, short by design
```

```python
# oidc.py
@cached(
    cache=TTLCache(
        maxsize=settings.AUTH_CACHE_MAX_SIZE, ttl=settings.AUTH_OIDC_CACHE_TTL
    ),
    lock=Lock(),
)
```

That keeps the security property you want for tokens without regressing Basic
auth. If you'd rather keep a single setting, that's a defensible call, but it
should be called out in the CHANGELOG as a performance-affecting change to Basic
auth rather than described only as an auth caching tweak.

Related: `src/helm/ralph/values.yaml` still ships `cacheTTL: 3600`, so the chart
default and the code default now disagree. Whichever way you go, they should be
aligned.

---

## 🟡 6. Minor points

**Test cache isolation.** `tests/api/auth/test_oidc.py:156` clears
`discover_provider`, but nothing clears the two new `TTLCache`s. `cachetools`
exposes `cache_clear()` on both (verified locally), so adding
`get_token_info.cache_clear()` and `get_user_info_data.cache_clear()` next to
the existing call would prevent cross-test bleed as the suite grows.

**Commit message typo.** Two commits use `(oicd)` instead of `(oidc)`:

```
🐛(oicd) fix ignore unrelated scopes
🐛(oicd) fix types of id token
```

`gitlint` accepts any scope so this doesn't fail CI, but since you'll be rebasing
anyway it's a cheap fix.

---

## Rebase and conflict

The branch is still based on `5e1558d`. `CHANGELOG.md` conflicts with `main`:

```
CONFLICT (content): Merge conflict in CHANGELOG.md
```

`check-changelog` and the four `test-python` jobs should clear on their own after
the rebase: your diff touches nothing under `tests/backends/`, and #634 (already
on `main`) is what fixed the mongo tests and the `check-changelog` job.

To summarise what is actually left: remove the `print`, fix the `HTTPException`
call, run `ruff check --fix` plus `black`, rewrap the CHANGELOG lines, and rebase.
The TTL question in point 5 is the only one that needs a decision rather than a
mechanical fix.

piptouque:
Thanks for the review, I think that's all for the fixes.

MYilFun00 (changes requested):
Thanks @piptouque. I re-tested the current head after a local rebase onto `main` (`53cc58c3`). Lint is green and the 22 CI checks match that — but CI does not cover the upgrade path of an existing OIDC deployment, nor the `Content-Type` error branch.

The mechanical points from last round are done (`print`, ruff/black, CHANGELOG ≤ 80, cache_clear, rebase). The runtime bug from last round is **not** fixed, and the new flow introduces a regression on current `main`.

Please do **not** update-branch / merge until the points below are addressed. Suggested fixes below are the ones I validated locally: existing JWT deployments keep working, `/userinfo` + client-credentials still work, `AUTH_CACHE_TTL` stays 3600 for Basic auth.

---

## 🔴 1. `HTTPException(error=...)` — still raises `TypeError` (same as last review)

`src/ralph/api/auth/oidc.py` still passes `error="invalid_request"`. FastAPI's signature is `(status_code, detail=None, headers=None)`.

```
TypeError: HTTPException.__init__() got an unexpected keyword argument 'error'
```

An IdP returning `text/html` or `application/jwt+json` still gets a **500 + stacktrace** instead of the intended 400. Still no test on that branch, which is why CI stays green.

**Suggested fix:**

```python
        if not is_jwt and not is_json:
            raise HTTPException(
                status_code=status.HTTP_400_BAD_REQUEST,
                detail=(
                    f"Invalid media type in header: {media_type}, "
                    "expected application/jwt or application/json"
                ),
                headers={"WWW-Authenticate": "Bearer"},
            )
        body = response.text if is_jwt else response.json()
        return (body, is_jwt)
    except HTTPException:
        raise
    except requests.exceptions.RequestException as exc:
        ...
```

Drop `error=`, split `detail` (clears E501), and re-raise `HTTPException` so it is not swallowed.

**Test** (the only branch of this function currently unexercised):

```python
@pytest.mark.anyio
@responses.activate
async def test_api_auth_oidc_userinfo_invalid_content_type():
    get_user_info_data.cache_clear()
    responses.add(
        responses.GET,
        "https://idp.example/userinfo",
        body="not json",
        status=200,
        headers={"Content-Type": "text/html"},
    )
    with pytest.raises(HTTPException) as exc_info:
        get_user_info_data("https://idp.example/userinfo", "token")
    assert exc_info.value.status_code == 400
    assert "text/html" in exc_info.value.detail
```

---

## 🔴 2. Existing OIDC deployments 500 after upgrade

On current `main`, `get_oidc_user` decodes the JWT **locally** via JWKS. This PR always calls `/introspection` with:

```
RUNSERVER_AUTH_OIDC_CLIENT_ID     = None   # default
RUNSERVER_AUTH_OIDC_CLIENT_SECRET = None
```

```
AttributeError: 'NoneType' object has no attribute 'encode'
```

Every authenticated request dies. Tests always `monkeypatch` `CLIENT_ID`/`CLIENT_SECRET`, so they never see this.

`introspection_endpoint` is also **optional** in OIDC Discovery. A generic IdP without it raises `KeyError` → 500. That contradicts the PR title.

**Suggested behaviour (opt-in, no break):**

```python
def _can_introspect(provider_config: dict) -> bool:
    """True when Ralph has client credentials and the IdP exposes introspect."""
    return bool(
        settings.RUNSERVER_AUTH_OIDC_CLIENT_ID
        and settings.RUNSERVER_AUTH_OIDC_CLIENT_SECRET
        and provider_config.get("introspection_endpoint")
    )


def get_oidc_user(...):
    ...
    access_token = auth_header.split(" ")[-1]
    provider_config = discover_provider(settings.RUNSERVER_AUTH_OIDC_ISSUER_URI)

    if not _can_introspect(provider_config):
        # Existing deployments: decode the JWT locally, no client credentials.
        return _user_from_jwt(access_token, provider_config)

    token_info = get_token_info(
        provider_config["introspection_endpoint"],
        token=access_token,
        client_id=settings.RUNSERVER_AUTH_OIDC_CLIENT_ID,
        client_secret=settings.RUNSERVER_AUTH_OIDC_CLIENT_SECRET,
    )
    ...
```

- Client id **and** secret **and** `introspection_endpoint` set → new flow (introspect, then `/userinfo` for users / `client_id` for apps).
- Otherwise → keep today's local JWT validation.

That is how existing Keycloak-style deployments survive, and how opaque-token / client-credentials support stays available once operators register Ralph as a client.

**Test** for the upgrade path (no client id/secret, JWT bearer token, IdP without `introspection_endpoint`):

```python
async def test_api_auth_oidc_legacy_jwt_without_client_credentials(client, monkeypatch):
    configure_env_for_mock_oidc_auth(monkeypatch, with_oidc_client=False)
    oidc_token = mock_oidc_jwt_user(scopes=["all", "profile/read"])
    response = await client.get("/whoami", headers={"Authorization": f"Bearer {oidc_token}"})
    assert response.status_code == 200
    assert response.json()["agent"]["openid"] == "https://iss.example.com/123|oidc"
```

---

## 🔴 3. `{iss}/{sub}` → `{iss}/user/{sub}` breaks stored authority

On `main` the agent is `f"{iss}/{sub}"`. This PR writes `f"{iss}/user/{sub}"`.

That value is copied onto every statement as `authority`, and `mine=true` / `LRS_RESTRICT_BY_AUTHORITY` filter on it. After upgrade, previously stored statements **no longer match** the same user.

**Suggested fix** — keep `{iss}/{sub}` for end-users; `/application/` only for the new client-credentials path:

```python
    if token_info.sub:
        user_info = get_user_info(provider_config, access_token=access_token)
        if user_info.sub != token_info.sub:
            ...
            raise _unauthorized()
        return AuthenticatedUser(
            agent={"openid": f"{token_info.iss}/{user_info.sub}"},
            scopes=get_user_scopes(user_info.scope),
            target=user_info.target,
        )

    return AuthenticatedUser(
        agent={"openid": f"{token_info.iss}/application/{token_info.client_id}"},
        scopes=get_user_scopes(token_info.scope),
        target=token_info.target,
    )
```

Also update the `whoami` assertion (and `tests/api/test_statements_get.py`) from `{iss}/user/{sub}` back to `{iss}/{sub}`.

---

## 🔴 4. Helm: `RALPH_AUTH_OIDC_CACHE_TTL` is gated on **basic** auth

Current:

```yaml
{{- if .Values.lrs.auth.basic.enabled }}
  RALPH_AUTH_CACHE_MAX_SIZE: {{ .Values.lrs.auth.basic.cacheMaxSize | quote }}
  RALPH_AUTH_CACHE_TTL: {{ .Values.lrs.auth.basic.cacheTTL | quote }}
  RALPH_AUTH_OIDC_CACHE_TTL: {{ .Values.lrs.auth.oidc.cacheTTL | quote }}
{{- end }}
{{- if .Values.lrs.auth.oidc.enabled }}
  RALPH_RUNSERVER_AUTH_OIDC_AUDIENCE: ...
  RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI: ...
{{- end }}
```

Rendered: default (basic on, oidc off) sets an unused OIDC TTL; **oidc on, basic off** (this PR's target) **omits** the variable.

**Suggested fix** in `src/helm/ralph/templates/cm_lrs.yaml`:

```yaml
{{- if .Values.lrs.auth.basic.enabled }}
  RALPH_AUTH_CACHE_MAX_SIZE: {{ .Values.lrs.auth.basic.cacheMaxSize | quote }}
  RALPH_AUTH_CACHE_TTL: {{ .Values.lrs.auth.basic.cacheTTL | quote }}
{{- end }}
{{- if .Values.lrs.auth.oidc.enabled }}
  RALPH_AUTH_OIDC_CACHE_TTL: {{ .Values.lrs.auth.oidc.cacheTTL | quote }}
  RALPH_RUNSERVER_AUTH_OIDC_AUDIENCE: {{ .Values.lrs.authOIDCAudience | quote }}
  RALPH_RUNSERVER_AUTH_OIDC_ISSUER_URI: {{ .Values.lrs.authOIDCIssuerURI | quote }}
{{- end }}
```

---

## 🟠 5. `AUTH_CACHE_TTL` 3600 → **36000** (not 60, not 3600)

You did split `AUTH_OIDC_CACHE_TTL = 60` — thank you, that was the request.

But Basic auth now defaults to **10 hours** (`36000`), while `values.yaml` still has `cacheTTL: 3600`. Revoking a user in `auth.json` can take 10 h to take effect.

**Suggested fix** in `src/ralph/conf.py` — restore the Basic default, keep the OIDC split:

```python
    AUTH_CACHE_MAX_SIZE: int = 100
    AUTH_CACHE_TTL: int = 3600
    AUTH_OIDC_CACHE_TTL: int = 60
```

Drop the CHANGELOG line « changed default TTL of cache to 60 seconds » (wrong for Basic, trailing space).

---

## 🟡 6. The headline feature has no test

`tests/api/auth/test_oidc.py` still has the same 7 functions. The `agent={"openid": ".../application/{client_id}"}` path (token with no `sub`) is never exercised.

**Suggested test:**

```python
@pytest.mark.anyio
@responses.activate
async def test_api_auth_oidc_client_credentials_application_agent(client, monkeypatch):
    """A token without `sub` is treated as a client application."""
    configure_env_for_mock_oidc_auth(monkeypatch)
    oidc_token = mock_oidc_user(sub=None, scopes=["all", "profile/read"])
    response = await client.get(
        "/whoami",
        headers={"Authorization": f"Bearer {oidc_token}"},
    )
    assert response.status_code == 200
    assert response.json()["agent"] == {
        "openid": "https://iss.example.com/application/my-other-client-id",
        "objectType": "Agent",
    }
```

(`_mock_oidc_token_info` must omit `sub` when it is `None`.)

CHANGELOG: move the two `Fix …` lines into the existing `### Fixed` section (they currently sit under a second `### Changed`):

```markdown
### Fixed
- Fix type of OIDC ID tokens (`aud` may be a list, `exp`/`iat` may be floats)
- Ignore OIDC scopes unrelated to Ralph instead of returning an error

### Changed
- OIDC: query `/userinfo` for access tokens that are not JWTs
- OIDC: introspect tokens to support client-credentials apps as users
```

---

Happy to re-review once those are in.

piptouque:
Hi again, sorry I did not properly address your last review. Everything should be in order now.

- added tests for queries to the IdP's `/instrospection` and `/userinfo`
- Fixed the `TypeError` in `HTTPException`
- added back the previous authentication mechanism if the above endpoints can't be queried, as you suggested
- Fixed the various config errors
- Reverted the changed `openid` authority format for user (I did not initially plan to merge this work, and then forgot about it)

PR #639, opened 2026-09-06
🐛(tests) fix missing responses in oidc test

piptouque:
## Purpose

That test was missing `@responses.activate`, which meant that the reported error was not the right one.
It also made it leek out in tests that were run afterwards.

You can check that despite this test being configured for Basic authentication, it actually ran OpenID backend.
The corrected response comes from Basic backend.


piptouque:
Change was needed and integrated in #633 .

